PR #3, opened 2022-05-24
Setup "yarn create-package" command

pnavarrc:
Setup the [Plopfile](https://plopjs.com/documentation)  to scaffold a package with the basics (`README.ms`, `index.ts`, `package.json`). I also included the `LICENSE` file, but I'm not sure whether we need that on each package or not.

**Example**

```sh
$ yarn create-package
yarn run v1.22.17
$ yarn plop package
$ /Users/pablo/code/covidactnow/act-now-packages/node_modules/.bin/plop package
? name: datetime utils
? description: utility functions to manipulate and format times and dates
✔  ++ /packages/datetime-utils/README.md
✔  ++ /packages/datetime-utils/src/index.ts
✔  ++ /packages/datetime-utils/package.json
✔  ++ /packages/datetime-utils/LICENSE
✨  Done in 24.41s.
```

I will update the `README.md` file with better instructions too, but I want to do that on a separate PR.

### DIFF
--- a/packages/assert/package.json
+++ b/packages/assert/package.json
@@ -13,9 +13,6 @@
   "scripts": {
     "build": "tsc src/index.ts --outDir lib --declaration"
   },
-  "publishConfig": {
-    "registry": "https://registry.npmjs.org"
-  },
   "devDependencies": {
     "typescript": "^4.6.4"
   }
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
     "test": "jest",
     "test:watch": "jest --watch",
     "test:ci": "CI=true yarn test",
-    "lint": "eslint ."
+    "lint": "eslint .",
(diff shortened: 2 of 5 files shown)


packages/assert/package.json
```diff
@@ -13,9 +13,6 @@
   "scripts": {
     "build": "tsc src/index.ts --outDir lib --declaration"
   },
-  "publishConfig": {
-    "registry": "https://registry.npmjs.org"
-  },
```
mikelehen, line 18: Just double-checking this is an intentional change?

mikelehen (approved):

packages/assert/package.json
pnavarrc: Intentional, now that we use NPM we don't need to specify the registry (that's the default)

PR #4, opened 2022-05-25
Run yarn after generating a package

pnavarrc:
Adds a custom command to the Plopfile to run `yarn` in the package folder after creating a new package. It can be tested by running `yarn create-package` - the newly created package should have its own `node_modules` directory after creation.

I also simplified the templates a bit (removed the partial) to make the Plopfile easier to read (for people not familiar with Handlebars).

### DIFF
--- a/plopfile.mjs
+++ b/plopfile.mjs
@@ -1,16 +1,23 @@
+/**
+ * Plop is a tool that saves time and helps scaffolding code with consistency.
+ *
+ * See https://plopjs.com/documentation/ for more details.
+ */
+
 import _ from "lodash";
+import { exec } from "child_process";
 
 const templateReadme = prepareTemplate(`
-# {{> packageName}}
+# {{dashCase name}}
 
-{{> packageDescription}}
+{{sentenceCase description}}
 `);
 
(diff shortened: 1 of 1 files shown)


plopfile.mjs
```diff
@@ -28,35 +35,53 @@ const templatePackage = prepareTemplate(`
...
-  plop.setPartial("packageName", "{{dashCase name}}");
-  plop.setPartial("packageDescription", "{{sentenceCase description}}");
+  /**
+   * Sets a custom action type to run `yarn` on the new package folder. Note
+   * that it will only include typescript, any other dependencies will still
+   * need to be added after creating the package.
+   */
+  plop.setActionType("yarn", function (answers, config, plop) {
+    const dashCase = plop.getHelper("dashCase");
+    const command = `cd packages/${dashCase(answers.name)}; yarn; cd ../..;`;
```
mikelehen: FWIW- You can run it from the root of the repo, I think.  It doesn't have to be in the package directory.

mikelehen (approved):

plopfile.mjs
pnavarrc: amazing!

PR #5, opened 2022-05-25
Number formatting utils package

chasulin:
Adds package (and tests) with number formatting utils: `formatInteger`, `formatDecimal`, `formatPercent`, `formatMillions`, `formatBillions`


packages/number-format/src/index.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  describe("formatDecimal", () => {
+    test("Value to decimal string", () => {
+      expect(formatDecimal(1.2345)).toBe("1.23");
+      expect(formatDecimal(1.9)).toBe("1.9");
+      expect(formatDecimal(0)).toBe("0");
+    });
+    test("Value to decimal string with decimal places specified", () => {
+      expect(formatDecimal(1.2345, 3)).toBe("1.235");
+      expect(formatDecimal(0)).toBe("0");
+      expect(formatDecimal(1.9)).toBe("1.9");
```
mikelehen: These latter two tests aren't specifying decimal places.  Update or remove?

packages/number-format/src/index.ts
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @example
+ * formatDecimal(1.2345)    // "1.23"
+ * formatDecimal(1.2345, 3) // "1.235"
+ *
+ * @param value Number to format
+ * @param places Number of decimal places to round to (defaults to 2)
+ * @returns Formatted decimal (as string)
+ */
+export function formatDecimal(value: number, places = 2): string {
```
mikelehen: It seems inconsistent that formatMillions() and formatBillions() accept `Intl.NumberFormatOptions` but formatDecimal() and formatPercent() just accept `places` as a number.  Since we are officially making these into a reusable package, it's tempting to try to make them more consistent.

One option, which would preserve backwards compatibility, would be to change it to:

```
export function formatDecimal(value: number, placesOrOptions: number |  Intl.NumberFormatOptions = 2): string {
```

and use a typeof check to determine if `placesOrOptions` is a number or an object.  And then we'd do something similar for formatPercent, formatMillions, and formatBillions.

That said, this is a bit of scope creep, so maybe a TODO to make them more consistent would be fine for now. 😁   I'll leave it up to you based on free time / motivation / etc.

packages/number-format/src/index.ts
```diff
@@ -0,0 +1,99 @@
...
+ * formatInteger(1.1)   // "1"
+ * formatInteger(15999) // "15,999"
+ *
+ * @param value Number to format
+ * @returns Formatted number (as string)
+ */
+export function formatInteger(value: number): string {
+  return new Intl.NumberFormat(undefined, {
+    maximumFractionDigits: 0,
+  }).format(value);
```
mikelehen: FWIW- I think you can do `value.toLocaleString(undefined, { maximumFractionDigits: 0 })` which is a bit more concise.

packages/number-format/src/index.ts
```diff
@@ -0,0 +1,99 @@
+/**
+ * Format a number as an integer, rounding to the closest integer if necessary.
+ *
+ * @example
+ * formatInteger(1.9)   // "2"
+ * formatInteger(1.1)   // "1"
+ * formatInteger(15999) // "15,999"
+ *
+ * @param value Number to format
+ * @returns Formatted number (as string)
```
mikelehen, line 10: Not sure if these comments were preexisting or not, but they're very nice! 🎉

packages/number-format/src/index.ts
```diff
@@ -0,0 +1,99 @@
...
+ */
+export function formatDecimal(value: number, places = 2): string {
+  return new Intl.NumberFormat(undefined, {
+    maximumFractionDigits: places,
+  }).format(value);
+}
+
+/**
+ * Format a percentage representation of a number, rounding to the amount of decimal places specified
+ * by the places param, or rounded to the nearest whole integer if no places param is provided.
```
mikelehen: Perhaps add "Note that the input number is multiplied by 100 to create a percentage (e.g. 0.85 => 85%)." 

That's pretty clear from the examples, but might not be expected if you don't pay attention.

packages/number-format/src/index.ts
```diff
@@ -0,0 +1,99 @@
...
+ * @example
+ * formatPercent(0.85)      // 85%
+ * formatPercent(0.1234)    // 12%
+ * formatPercent(0.8567, 2) // 85.67%
+ *
+ * @param value Number to format
+ * @param places Number of decimal places to round to (defaults to 0)
+ * @returns Formatted percentage (as string)
+ */
+
```
mikelehen, line 59: extra newline

mikelehen (approved):
Looks great! 😍 

Left a couple small suggestions and one big (optional) suggestion for improved consistency, but feel free to punt on that.

packages/number-format/src/index.ts
mikelehen: By the way, if we went this route we could use TypeScript [function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) to make it a bit cleaner / explicit:

```
export function formatDecimal(value: number, places: number): string;
export function formatDecimal(value: number, options: Intl.NumberFormatOptions): string;
export function formatDecimal(value: number, placesOrOptions: number |  Intl.NumberFormatOptions = 2): string {
  ...
}
```

packages/number-format/src/index.test.ts
chasulin: oh whoops, this was my poor copy and pasting

packages/number-format/src/index.ts
chasulin: yes! i grabbed this format from gcat // @pnavarrc

packages/number-format/src/index.ts
chasulin: sweeet! updated

packages/number-format/src/index.ts
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * Format a number using fixed point notation.
+ *
+ * @example
+ * formatDecimal(1.2345)    // "1.23"
+ * formatDecimal(1.2345, 3) // "1.235"
+ *
+ * @param value Number to format
+ * @param places (optional) Number of decimal places to round to (defaults to 2)
+ * @param options (optional) Object of international number formatting options (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat)
```
mikelehen, line 25: Ah, nice.  Thanks for updating the param docs to match!

packages/number-format/src/index.ts
```diff
@@ -0,0 +1,139 @@
...
+): string;
+export function formatDecimal(
+  value: number,
+  placesOrOptions: number | Intl.NumberFormatOptions = 2
+): string {
+  return typeof placesOrOptions === "number"
+    ? new Intl.NumberFormat(undefined, {
+        maximumFractionDigits: placesOrOptions,
+      }).format(value)
+    : new Intl.NumberFormat(undefined, placesOrOptions).format(value);
```
mikelehen: Totally optional, but consider a helper like:

```
function getOptionsFromPlacesOrOptions(placesOrOptions?: number | Intl.NumberFormatOptions): Intl.NumberFormatOptions | undefined {
  return (typeof placesOrOptions === 'number') ?
    {
      maximumFractionDigits: placesOrOptions,
    } : placesOrOptions;
}
```

So then we can just do:
```
return new Intl.NumberFormat(undefined, getOptionsFromPLacesOrOptions(placesOrOptions)).format(value);
```

in each of these functions.   Less duplication / boilerplate!

packages/number-format/src/index.ts
mikelehen: Extra newline is still here (and in a couple of the other functions in here too), I think?

mikelehen (approved):
One last cleanup suggestion, but this looks great. I like the way it turned out. And thanks for the attention to detail in updating the param docs and tests, etc.!

PR #7, opened 2022-05-30
Add time utils

faihegberg:
- Add time utils as a NPM package (code mostly taken from CAN repo).
- Fix minor errors in README and plop file.


packages/time-utils/src/index.test.ts
```diff
@@ -0,0 +1,222 @@
...
+  formatUTCDateTime,
+  parseDateString,
+  parseDateUnix,
+  addTime,
+  subtractTime,
+  getStartOf,
+  getTimeDiff,
+} from "./index";
+
+// Date object constructor takes a zero-indexed month number.
```
mikelehen: This comment feels a little out of place.  I guess it's not tied to any specific code and it's just a general reminder for somebody trying to understand this code?  That's a good idea, but unless somebody is reading the entire file top to bottom (which is unusual), they're likely to miss the comment when it would be most helpful (e.g. debugging a particular test).

Perhaps consider creating a helper function that hides this oddity of `Date` instead, like:

```
/** 
 * Helper to create Date instances from a numeric year, month, and day.
 *
 * This mostly exists because the Date constructor expects 0-indexed months, 
 * which is pretty unnatural.
 * 
 * The resulting Date will be set to midnight local (Pacific) time.
 * 
 * @param year Numeric year (e.g. 2020)
 * @param month The 1-indexed month (e.g. 3 for March)
 * @param day 1-indexed day (e.g. 25 for the 25th of March).
 * @returns The created Date object.
 */
function createDate(year: number, month: number, day: number) {
  return new Date (year, month-1, day);
}
```

packages/time-utils/src/index.test.ts
```diff
@@ -0,0 +1,222 @@
...
+
+describe("time formatting", () => {
```
mikelehen: nit: These tests are all more about date formatting than time formatting.  So I'd say "date formatting" or "date/time formatting" if we might intend to add time-related tests in the future.

packages/time-utils/src/index.test.ts
```diff
@@ -0,0 +1,222 @@
...
+  const testDate = new Date(2020, 2, 1);
+  test("Date in ISO format", () => {
+    expect(formatDateTime(testDate, DateFormat.YYYY_MM_DD)).toBe("2020-03-01");
+  });
+  test("Dash separated date", () => {
+    expect(formatDateTime(testDate, DateFormat.MM_DD_YYYY)).toBe("03/01/2020");
```
mikelehen, line 53: This says "Dash separated" but the date is slash-separated?

packages/time-utils/src/index.test.ts
```diff
@@ -0,0 +1,222 @@
...
+} from "./index";
+
+// Date object constructor takes a zero-indexed month number.
+
+describe("time formatting", () => {
+  const testDate = new Date(2020, 2, 1);
+  test("Date in ISO format", () => {
+    expect(formatDateTime(testDate, DateFormat.YYYY_MM_DD)).toBe("2020-03-01");
+  });
+  test("Dash separated date", () => {
```
mikelehen: Can we put newlines between tests for readability?

packages/time-utils/src/index.test.ts
```diff
@@ -0,0 +1,222 @@
...
+  test("Past date", () => {
+    expect(formatDateTime(new Date(2015, 1, 1), DateFormat.YYYY_MM_DD)).toBe(
+      "2015-02-01"
+    );
+  });
+  test("Future date", () => {
+    expect(formatDateTime(new Date(2025, 1, 1), DateFormat.YYYY_MM_DD)).toBe(
+      "2025-02-01"
+    );
+  });
```
mikelehen: nit: Is there a particular reason we need specific tests for past dates and future dates?  In general, if a test hits the exact same codepath as other tests, then it's probably redundant and you might consider removing it. 😁 

While it doesn't hurt much to have extra tests, it can become annoying to find the tests you're looking for, or to have extra tests to update if we were to e.g. change the way `formatDateTime` works in the future such that all tests need to be updated.

packages/time-utils/src/index.test.ts
```diff
@@ -0,0 +1,222 @@
...
+});
+
+describe("utc time formatting", () => {
+  // Pacific timezone is set as the default in package.json for testing consistency.
+  const testDate = new Date(2020, 2, 1, 20); // March 1st 2020 8 pm. Pacific.
```
mikelehen: It took me a minute to figure out exactly what's going on with this set of tests.  They look very very similar to the "time formatting" tests above, but they are using `formatUTCDateTime()` instead of `formatDateTime()` and *also* they are using an 8PM (Pacific) time instead of a midnight (Pacific) time.  But why are we changing both the method we call and the Date we call it on?

I think it might be much clearer if we create a single date that is used by both sets of tests (see my comment above).  That way the only thing that changes between the two sets of tests is whether we call `formatDateTime()` vs `formatUTCDateTime()`.

E.g.:

```
/**
 * Test date to use in tests.  Note that our tests are configured in the 
 * package.json to always run in the Pacific timezone, for testing consistency.
 * So this Date represents 2020/03/02 3AM in UTC but 2020/03/01 7PM (or 8PM 
 * depending on daylight savings) in local (Pacific) timezone. So depending which
 * formatting function you use, you'll get back a different date.
 */
const MARCH_2_2020_3AM_UTC = new Date('2020-03-02T03:00:00Z')
```

WDYT?

packages/time-utils/src/index.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    expect(parseDateString("2020-03-01")).toEqual(new Date(2020, 2, 1));
+  });
+  test("Invalid date string", () => {
+    expect(() => parseDateString("Hello")).toThrow();
+  });
+});
+
+// Unix time are expressed in milliseconds.
+describe("unix time parsing", () => {
+  test("unix time at midnight", () => {
```
mikelehen: This is really "unix time at 8AM UTC", not midnight, right?

packages/time-utils/src/index.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    expect(subtractTime(new Date(2020, 2, 1), 3, TimeUnit.MONTHS)).toEqual(
+      new Date(2019, 11, 1)
+    );
+  });
+});
+
+describe("get start of time", () => {
+  test("Start of hour", () => {
+    expect(getStartOf(new Date(2020, 2, 1, 0, 30), TimeUnit.HOURS)).toEqual(
+      new Date(2020, 2, 1, 0, 0)
```
mikelehen: nit: This might not be the best test case since hours is `0` already?  So e.g. if we had a bug that treated `TimeUnit.HOURS` as `TimeUnit.DAYS` this test would still pass, I think?

packages/time-utils/src/index.test.ts
mikelehen: Oh, I guess we might need hours/minutes as well.

packages/time-utils/src/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  HOURS = "hour",
+  DAYS = "day",
+  WEEKS = "week",
+  MONTHS = "month",
+}
+
+/**
+ * Parse date string to JS date object.
+ * Date string can be in any ISO 8601 format: https://moment.github.io/luxon/docs/manual/parsing.html
+ * Example: parseDateString('2020-03-01') // Return JS date object for March 1st 2020.
```
mikelehen: Can you follow the example format we've started using in some of the other packages? https://github.com/covid-projections/act-now-packages/blob/4b707346f574c4fd713d97d9e73444f2489ae99f/packages/number-format/src/index.ts#L4

packages/time-utils/src/index.ts
```diff
@@ -0,0 +1,116 @@
...
+// This enum maps to singular to make it usable by other Luxon functions that accept singular strings (e.g. 'hour').
+export enum TimeUnit {
+  HOURS = "hour",
+  DAYS = "day",
+  WEEKS = "week",
+  MONTHS = "month",
+}
+
+/**
+ * Parse date string to JS date object.
```
mikelehen, line 28: nit: By convention there's usually a newline after the initial 1-sentence description of the function.

packages/time-utils/src/index.ts
```diff
@@ -0,0 +1,116 @@
...
+ * Date string can be in any ISO 8601 format: https://moment.github.io/luxon/docs/manual/parsing.html
+ * Example: parseDateString('2020-03-01') // Return JS date object for March 1st 2020.
+ */
```
mikelehen, line 34: Feel free to punt for now, but if we wanted to go all-in with tsdoc comments, we could add `@param` and `@returns` descriptions of the params / return value, similar to what we did in number-format and assert. https://github.com/covid-projections/act-now-packages/blob/4b707346f574c4fd713d97d9e73444f2489ae99f/packages/number-format/src/index.ts#L23

packages/time-utils/src/index.ts
```diff
@@ -0,0 +1,116 @@
...
+ * Example: addTime(new Date(2020, 2, 1), 5, TimeUnit.DAYS) // Return JS date object for March 6th 2020.
+ */
+export function addTime(date: Date, amount: number, unit: TimeUnit): Date {
+  return DateTime.fromJSDate(date)
+    .plus(getTimeUnitOption(amount, unit))
+    .toJSDate();
+}
+
+/**
+ * Subtract a specified amount of time (in units) to date object.
```
mikelehen: grammar nit: You generally say, "subtract ... from" not "subtract ... to" so this would be "... from date object."

packages/time-utils/src/index.ts
```diff
@@ -0,0 +1,116 @@
...
+ * Example: subtractTime(new Date(2020, 2, 10), 5, TimeUnit.DAYS) // Return JS date object for March 5th 2020.
+ */
+export function subtractTime(date: Date, amount: number, unit: TimeUnit): Date {
+  return DateTime.fromJSDate(date)
+    .minus(getTimeUnitOption(amount, unit))
+    .toJSDate();
+}
+
+// Some Luxon functions accept a DurationObject instead of a string as parameter.
```
mikelehen: I think this comment is explaining why this function exists instead of what it does.  Can you rewrite it as a full tsdoc comment and add a bit more explanation (e.g. "Converts our TimeUnit enum into a luxon Duration object.")?

plopfile.mjs
```diff
@@ -15,7 +15,7 @@ const templateReadme = prepareTemplate(`
 ## Installing
 
 \`\`\`sh
-yarn add @actnowcoalition/number-format
+yarn add @actnowcoalition/{{dashCase name}}
```
mikelehen, line 18: Hah, woops.  Good catch! 😆

mikelehen (changes requested):
Nice!  So this looks 100% fine but given we are turning this into a standalone package I took the opportunity to go over it with a fine-tooth comb and be pretty picky. 😁   I think they should all be quick fixes (mostly comments, etc.) but holler if anything feels onerous or whatever.

packages/time-utils/src/index.test.ts
faihegberg: Yep, this definitely makes sense! This would be much clearer 👍🏼

faihegberg: Actually I think I'll just use the same date and time as you suggested for both sets of tests, but I'll create it using the new `createDate` helper function.

packages/time-utils/src/index.test.ts
faihegberg: Yes!

packages/time-utils/src/index.test.ts
faihegberg: Changed test case to a non-zero hour instead.

packages/time-utils/src/index.ts
faihegberg: I'll do this in a follow up PR 🙂

packages/time-utils/src/index.test.ts
```diff
@@ -0,0 +1,298 @@
...
+  getTimeDiff,
+} from "./index";
+
+/**
+ * Helper to create Date instances from a numeric year, month, and day.
+ *
+ * This mostly exists because the Date constructor expects 0-indexed months,
+ * which is pretty unnatural.
+ *
+ * The resulting Date will be set to midnight local (Pacific) time.
```
mikelehen: Oh, this doesn't quite apply anymore since we accept hour / minute params.  So I'd remove or tweak it so it still applies.

packages/time-utils/src/index.test.ts
```diff
@@ -0,0 +1,298 @@
...
+ *
+ * @param year Numeric year (e.g. 2020)
+ * @param month The 1-indexed month (e.g. 3 for March)
+ * @param day 1-indexed day (e.g. 25 for the 25th of March).
```
mikelehen: Can you add `hour` and `minute`?

packages/time-utils/src/index.test.ts
```diff
@@ -0,0 +1,298 @@
...
+  month: number,
+  day: number,
+  hour?: number,
+  minute?: number
+) {
+  if (hour && minute) {
+    return new Date(year, month - 1, day, hour, minute);
+  } else if (hour) {
+    return new Date(year, month - 1, day, hour);
+  } else return new Date(year, month - 1, day);
```
mikelehen: Huh. I'm kinda' surprised `prettier` doesn't automatically "fix" this, but I think by convention we almost always wrap if/else blocks in brackets (`{ ... }`).

That said, I think you can maybe just get rid of the if/else logic and do

```
return new Date(year, month - 1, day, hour ?? 0, minute ?? 0);
```

?

packages/time-utils/src/index.test.ts
```diff
@@ -0,0 +1,298 @@
...
+}
+
+/**
+ * Test date to use in tests.  Note that our tests are configured in the
+ * package.json to always run in the Pacific timezone, for testing consistency.
+ * So this Date represents 2020/03/01 8AM in UTC but 2020/03/01 12AM (or 1AM
+ * depending on daylight savings) in local (Pacific) timezone. So depending which
```
mikelehen: Oh, I just realized I think I confused myself with the "depending on daylight savings" bit.

Since 3/1 is not during daylight savings time, it'll just always be 12AM I guess.  Sorry!

packages/time-utils/src/index.test.ts
```diff
@@ -0,0 +1,298 @@
...
+    return new Date(year, month - 1, day, hour, minute);
+  } else if (hour) {
+    return new Date(year, month - 1, day, hour);
+  } else return new Date(year, month - 1, day);
+}
+
+/**
+ * Test date to use in tests.  Note that our tests are configured in the
+ * package.json to always run in the Pacific timezone, for testing consistency.
+ * So this Date represents 2020/03/01 8AM in UTC but 2020/03/01 12AM (or 1AM
```
mikelehen: I think the date/time you have chosen defeats the purpose of the tests.  I think the original intention was to test that formatDateTime() interprets the Date as local timezone and formatUtcDateTime() interprets the Date as UTC timezone, but because it ends up being 2020/03/01 in both timezones, our tests don't actually verify that `formatDateTime()` and `formatUtcDateTime()` behave differently anymore.

packages/time-utils/src/index.test.ts
```diff
@@ -0,0 +1,298 @@
...
+ * depending on daylight savings) in local (Pacific) timezone. So depending which
+ * formatting function you use, you'll get back a different date.
+ */
+const MARCH_1_2020_8AM_UTC = createDate(2020, 3, 1, 0, 0);
```
mikelehen: nit but I think using `createDate()` but then defining the variable name in terms of UTC makes this code harder to read.  I have to do timezone math to understand it, where as ideally code should be "obviously correct" (not require much thought to verify its correctness). So I would stick with either local timezone or UTC timezone and do either:

```
const MARCH_1_2020_8AM_UTC = new Date('2020-03-01T08:00:00Z');
```

or

```
const MARCH_1_2020_12AM_PACIFIC = createDate(2020, 3, 1, 0, 0)
```

Does that make sense?

packages/time-utils/src/index.ts
```diff
@@ -36,36 +39,42 @@ export function parseDateString(dateString: string): Date {
 
 /**
  * Parse unix timestamp (in milliseconds) to JS date object.
- * Example: parseDateUnix(1616612130826) // Return JS date object for March 24th 2021 18:55:30 UTC.
+ *
+ * @example: parseDateUnix(1616612130826) // "2021-03-24T18:55:30.826Z"
```
mikelehen: No `:` after `@example`.  https://tsdoc.org/pages/tags/example/

packages/time-utils/src/index.ts
```diff
@@ -74,16 +83,24 @@ export function addTime(date: Date, amount: number, unit: TimeUnit): Date {
...
     .minus(getTimeUnitOption(amount, unit))
     .toJSDate();
 }
 
-// Some Luxon functions accept a DurationObject instead of a string as parameter.
+/**
+ * Convert a TimeUnit enum to a luxon Duration object.
+ *
+ * Some luxon functions accept a Duration object instead of a string as parameter.
+ * So we must convert a TimeUnit enum to a luxon Duration object first before using these functions.
```
mikelehen, line 106: Very helpful comment, thanks!

mikelehen (changes requested):
LGTM but I think we accidentally defeated the purpose of some of the `formatDateTime()` / `formatUtcDateTime()` tests.  Mind taking another look?

packages/time-utils/package.json
```diff
@@ -0,0 +1,23 @@
...
+  ],
+  "author": "Act Now Coalition",
+  "license": "MIT",
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
+  "scripts": {
+    "build": "tsc src/index.ts --outDir lib --declaration"
+  },
+  "devDependencies": {
+    "typescript": "^4.6.4"
```
pnavarrc, line 24: We need to add `@actnowcoalition/assert` as `dependency` here

package.json
```diff
@@ -34,6 +35,7 @@
     "jest": "^28.1.0",
     "lint-staged": "^12.4.1",
     "lodash": "^4.17.21",
+    "luxon": "^2.4.0",
```
pnavarrc: The `luxon` library should be a dependency of the `@actnowcoalition/time-utils` package, so it should be in `packages/time-utils/package.json`

packages/time-utils/src/index.ts
```diff
@@ -0,0 +1,135 @@
+import { DateTime, Duration } from "luxon";
+import { assert } from "@actnowcoalition/assert/src";
```
pnavarrc: This should be `@actnowcoalition/assert`

packages/time-utils/src/index.test.ts
faihegberg: True!

packages/time-utils/src/index.test.ts
faihegberg: I'll change it to 3.15 am. UTC on March 1st instead so we can more clearly see the difference between `formatDateTime()` and `formatUTCDateTime()`.

packages/time-utils/package.json
mikelehen: Oh, good catch!  Normally I don't look that closely at this kind of change because I assume if it compiles / runs, then we have all the necessary dependencies.  But I guess that's not necessarily the case here.  Good to know!

package.json
mikelehen: Same here.  

I do wonder if there is a way we could / should test for this sort of thing so that it gets caught by CI rather than needing to carefully review dependencies. 🤔

packages/time-utils/src/index.ts
mikelehen: Another good catch.  I wonder if there's a way to include the source with the package but not make it import-able directly?  Hrm.

packages/time-utils/src/index.test.ts
```diff
@@ -31,52 +31,47 @@ function createDate(
...
 }
 
 /**
  * Test date to use in tests.  Note that our tests are configured in the
  * package.json to always run in the Pacific timezone, for testing consistency.
- * So this Date represents 2020/03/01 8AM in UTC but 2020/03/01 12AM (or 1AM
- * depending on daylight savings) in local (Pacific) timezone. So depending which
- * formatting function you use, you'll get back a different date.
+ * So this Date represents 2020/03/01 3.15 AM in UTC but 2020/02/29 7.15 PM in local (Pacific) timezone.
+ * So depending which formatting function you use, you'll get back a different date.
```
mikelehen, line 41: Not a big deal, but why did you decide to include 15 minutes?  It's not really harmful, but I'm also not sure if it's helpful? 😁  I'd generally err towards keeping tests as simple as possible while testing the functionality that you need to test. 🤷

packages/time-utils/src/index.test.ts
```diff
@@ -31,52 +31,47 @@ function createDate(
...
+const MARCH_1_2020_3_15_AM_UTC = new Date("2020-03-01T03:15:00Z");
 
-describe("date/time formatting", () => {
-  const testDate = MARCH_1_2020_8AM_UTC;
+describe("date formatting", () => {
+  const testDate = MARCH_1_2020_3_15_AM_UTC;
 
   test("Date in ISO format", () => {
-    expect(formatDateTime(testDate, DateFormat.YYYY_MM_DD)).toBe("2020-03-01");
+    expect(formatDateTime(testDate, DateFormat.YYYY_MM_DD)).toBe("2020-02-29");
```
mikelehen: Okay, I may be nitpicking too much, but in order for somebody reading this test to understand why the result is 2020-02-29 they need to know:

1. That our test date is 2020/03/01 3:15 AM UTC and there is an 8 hour time difference between UTC and local (Pacific) time so the local time will represent the date prior to 2020/03/01.
2. That February normally has 28 days.
3. That 2020 was a leap year, and so February had 29 days.

If we picked e.g. `MARCH_2_2020_3_15_AM_UTC` as the test date (March 2 instead of March 1) then a reader of the code wouldn't need to know 2 or 3 and so the test would be more straightforward / easy to understand.

mikelehen (approved):
Couple more optional minor comments from me.  Otherwise LGTM after you fix Pablo's 3 points.

packages/time-utils/src/index.test.ts
faihegberg: I added it just in case we add functionalities dealing with minutes in the future, but for now, we don't _need_ it. I'll keep it at the hour.

PR #11, opened 2022-07-05
Initial setup for regions

pnavarrc:
Initial implementation of the regions package.

In this iteration, we are trying to have `Region` to be generic enough that it can replace `State`, `County`, `CongressionalDistrict`, etc. Ideally, the code using the regions won't need to know about the type of region.

We did add a `parent` property to be able to represent hierarchy between regions and to test if one region contains the other, which can be used to have lists of counties in a state, etc.

### Example

```tsx
// state page (/us/[stateId]/index.tsx)
import { Region, states } from "@actnowcoalition/regions";

const StatePage: React.FC<{regionJSON: regionJSON }> = ({ regionJSON }) => {
  const region = Region.fromJSON(regionJSON);
  return (
    <Fragment>
      <h1>{region.fullName}</h1>
      <p>{region.population}</p>
    </Fragment>
  );
}
export const getStaticPaths: GetStaticPaths = async () => {
  const paths = states.map(state => ({ 
    params: { stateId: state.urlFragment }
  }));

  return { paths, fallback: false };
}


```


packages/regions/tsconfig.json
```diff
@@ -0,0 +1,11 @@
...
+  "extends": "../../tsconfig",
+  "compilerOptions": {
+    "baseUrl": ".",
+    "declaration": true,
+    "declarationDir": "lib",
+    "outDir": "lib",
+    "rootDir": "src"
+  },
+  "include": ["src/**/*"]
+}
```
pnavarrc, line 13: Apparently the build command was using the default config (I noticed this because when I tried to build there was an error saying that `get` wasn't enabled with `es5` as a target) - we probably need to add a `tsconfig.json` file on every package

packages/regions/src/datasets/us/County.ts
```diff
@@ -0,0 +1,55 @@
...
+    return `${this.name}, ${this.state.stateCode}`;
+  }
+
+  get abbreviation(): string {
+    return getAbbreviatedCounty(this.name);
+  }
+
+  get relativeUrl(): string {
+    return `${this.state.relativeUrl}/county/${this.urlSegment}`;
+  }
```
pnavarrc: TODO: Use `urlJoin` here (there is some error about `url-join` having 'export'). Specifically, if I update the code to be

```ts
import urlJoin from 'url-join';

// ...

  get relativeUrl(): string {
    return urlJoin(`${this.state.relativeUrl}`, `/county/${this.urlSegment}`);
  }
```

then when running the tests, I get this error:

```
 FAIL  packages/regions/src/datasets/us/counties_db.test.ts
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /Users/pablonavarro/code/actnow/act-now-packages/packages/regions/node_modules/url-join/lib/url-join.js:60
    export default function urlJoin() {
    ^^^^^^

    SyntaxError: Unexpected token 'export'

      1 | import { Region } from "../../Region";
      2 | import { State } from "./State";
    > 3 | import urlJoin from "url-join";
        | ^
      4 |
      5 | export class County extends Region {
      6 |   constructor(

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1773:14)
      at Object.<anonymous> (packages/regions/src/datasets/us/County.ts:3:1)
```

I'm going to try to figure out how to update the Jest config to better handle this!

**Update**: I tried updating `transformIgnorePatterns` to include the path to `url-join` (to exclude it from the ignored paths), but it still doesn't work.

packages/regions/src/datasets/us/counties_db.ts
```diff
@@ -0,0 +1,18 @@
...
+  return new County(
+    county.name,
+    county.fipsCode,
+    county.population,
+    state as State
+  );
+});
+
+const countiesDB = new RegionDB(counties);
+export default countiesDB;
```
pnavarrc, line 22: Not sure if we should export an instance, that might prevent us from marking the package as side-effects free and therefore preventing us from enabling tree shaking.

packages/regions/src/Region.ts
```diff
@@ -0,0 +1,39 @@
...
+    public readonly name: string,
+    public readonly regionId: string,
+    public readonly population: number
+  ) {}
+
+  abstract get fullName(): string;
+  abstract get shortName(): string;
+  abstract get abbreviation(): string;
+  abstract get relativeUrl(): string;
+  abstract get urlSegment(): string;
```
pnavarrc: TODO: Document the intent for each name, maybe add an example?

packages/regions/src/datasets/us/County.ts
```diff
@@ -0,0 +1,59 @@
...
+    return `${Region.toSlug(this.abbreviation)}`;
+  }
+
+  contains() {
+    return false;
+  }
+
+  static isCounty(region: Region): region is County {
+    return region instanceof County;
+  }
```
pnavarrc: This is new, I think it might be useful (and maybe cleaner than checking region types)

packages/regions/src/datasets/us/County.ts
mikelehen: Not sure what's going on, but it's tempting to just write our own urlJoin helper. 🤷

packages/regions/src/datasets/us/County.ts
mikelehen: It might be a good exercise to figure out why we think we need this (or region types in general).

In a totally ideal world, all of the functionality we need for regions would be baked into the `Region` base class so all of our maps / charts / whatever components would only need to know about `Region` and don't need custom code for counties or cities or census blocks or whatever other regions we come up with in the future.

If we *do* need region types in some form, I think it may be better to have a `RegionType` enum and a `.type` property on `Region`.  That way, like I said, external code doesn't need to be aware of the `County` class at all.  But maybe this is wishful thinking.

packages/regions/src/datasets/us/County.ts
```diff
@@ -0,0 +1,59 @@
+import { Region } from "../../Region";
+import { State } from "./State";
+
+export class County extends Region {
+  constructor(
+    name: string,
+    fipsCode: string,
+    population: number,
+    public readonly state: State
```
mikelehen: Could we generalize this and have the `Region` class have a `parent` property?

In a totally ideal world, all of the functionality we need for regions would be baked into the `Region` base class so all of our maps / charts / whatever components would only need to know about `Region` and don't need custom code to deal with `County` and its `state` property, etc.

packages/regions/src/Region.ts
mikelehen: FWIW- I think it's possible that we could make these all be `public readonly` properties on `Region` instead of using getters, and then instead of having a `County` class, we'd just construct `Region` objects with all of these properties set correctly to represent the counties correctly.

I say that because in an ideal world, none of our code would need to explicitly reference `County` or `State`, etc. All of our maps / charts / components would be powered by `Region` directly and then we wouldn't actually need to have those subclasses if we don't want.  But maybe it's cleaner / more convenient to have the subclasses.  I'm not opposed.  But I do want to try to make the `Region` interface sufficient that the subclasses can just be an implementation detail that the rest of our code doesn't know about.

packages/regions/src/Region.ts
```diff
@@ -0,0 +1,39 @@
...
+   *
+   *   Region.toSlug("Río Grande")    // rio_grande
+   *   Region.toSlug("DeSoto County") // de_soto_county
+   */
+  static toSlug(name: string): string {
+    return lowerCase(deburr(name)).split(" ").join("_");
+  }
+
+  // toJSON
+  // fromJSON
```
mikelehen: I'd remove these or make them self-explanatory TODOs.

packages/regions/src/Region.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  abstract contains(subregion: Region): boolean;
+
+  toString() {
+    return `${this.name} (regionId=${this.regionId})`;
+  }
+
+  /**
+   * Generates a slug from the input string by replacing accented characters
+   * with their basic Latin equivalents, replacing whitespaces with lodash
```
mikelehen: "lodash" => "underscores" ?  lodash is the name of the javascript library.  I guess maybe some people also call underscores a "low dash" or something, but never "lodash" as far as I know. :-)

packages/regions/src/RegionDB.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export class RegionDB {
+  private regionsById: { [regionId: string]: Region };
+
+  constructor(public readonly regions: Region[]) {
+    this.regionsById = keyBy(regions, (region: Region) => region.regionId);
+  }
+
+  findByRegionId(regionId: string): Region | null {
+    return this.regionsById[regionId] || null;
```
mikelehen: nit: `?? null` is cleaner / more explicit.

packages/regions/src/RegionDB.ts
```diff
@@ -0,0 +1,27 @@
+import keyBy from "lodash/keyBy";
+import { assert } from "@actnowcoalition/assert";
+import { Region } from "./Region";
+
+export class RegionDB {
+  private regionsById: { [regionId: string]: Region };
+
+  constructor(public readonly regions: Region[]) {
```
mikelehen: We have `regions` as a public property but then we also have `all` as a public property as well.  I think we should only have one way to get the list of regions.  Probably `all`?

packages/regions/src/datasets/us/County.ts
```diff
@@ -0,0 +1,59 @@
...
+  get shortName(): string {
+    return `${this.name}, ${this.state.stateCode}`;
+  }
+
+  get abbreviation(): string {
+    return getAbbreviatedCounty(this.name);
+  }
+
+  get relativeUrl(): string {
+    return `${this.state.relativeUrl}/county/${this.urlSegment}`;
```
mikelehen: If we add a `parent` to `Region` then `relativeUrl()` can maybe be implemented inside `Region` by walking up the parent chain.

packages/regions/src/Region.ts
```diff
@@ -0,0 +1,39 @@
+import lowerCase from "lodash/lowerCase";
+import deburr from "lodash/deburr";
+
+export abstract class Region {
+  constructor(
+    public readonly name: string,
+    public readonly regionId: string,
+    public readonly population: number
```
mikelehen, line 16: FWIW- It is tempting to treat population as a metric instead of baking it into the region class.  I'm not sure though, so this is fine for now.

packages/regions/src/datasets/us/State.ts
```diff
@@ -0,0 +1,41 @@
+import toLower from "lodash/toLower";
+import { Region } from "../../Region";
+import { County } from "./County";
+
+export class State extends Region {
+  constructor(
+    name: string,
+    regionId: string,
+    population: number,
+    public readonly stateCode: string
```
mikelehen: Again, I'd question why we need this to be public.  Ideally everything public would be on the `Region` class so that our code doesn't need to know about `State` or other custom region types.

packages/regions/src/datasets/us/State.ts
```diff
@@ -0,0 +1,41 @@
...
+  get relativeUrl(): string {
+    return `/us/${this.urlSegment}`;
+  }
+
+  contains(subregion: Region): boolean {
+    if (subregion instanceof County) {
+      return subregion.state.regionId === this.regionId;
+    }
+    return false;
+  }
```
mikelehen: I expect we'll need a way to get all counties within a state, right?  E.g. for populating the county map or compare table on a state page.

To do that, I'd like to explore if we can do it in a generic way on the `Region` class, by e.g. having a `children: Region[]` list or something.

Note that we may end up with use cases down the road where a region contains multiple types of other regions (e.g. a state contains both cities and counties).  And we might need to deal with "overlap" in addition to "containment".  E.g. CBSAs overlap with states but aren't strictly contained by a state in some cases.  I'm not sure the best way to represent these sorts of relationships, so probably best to punt for now. But it's something to think about.

packages/regions/src/datasets/us/counties_db.test.ts
```diff
@@ -0,0 +1,29 @@
...
+import { State } from "./State";
+
+describe("counties_db", () => {
+  test("`all` includes all counties", () => {
+    expect(countiesDB.all).toHaveLength(3224);
+  });
+
+  test("each county is correctly initialized", () => {
+    countiesDB.all.filter(County.isCounty).forEach((county) => {
+      expect(county).toBeTruthy();
```
mikelehen, line 10: nit: This isn't necessary since the following line would fail if it wasn't truthy.  I'd delete it.

packages/regions/src/datasets/us/counties_db.test.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  test("findRegionById", () => {
+    expect(countiesDB.findByRegionId("53033")).toBeDefined();
+    expect(countiesDB.findByRegionId("NO_COUNTY")).toBeNull();
+  });
+
+  test("findRegionByIdStrict", () => {
+    expect(countiesDB.findByRegionIdStrict("53033")).toBeDefined();
+    expect(countiesDB.findByRegionIdStrict("53033")).not.toBeNull();
+    expect(() => countiesDB.findByRegionIdStrict("53033")).not.toThrow();
```
mikelehen: nit: This is redundant since the above two lines didn't throw.  I'd remove it.

packages/regions/src/datasets/us/counties_db.ts
mikelehen: I'm not sure I follow.  I don't think creating an instance of an object counts as a side-effect.  If we were making a web request, or writing to localstorage, or even modifying some other global state, then it'd be a side-effect.

packages/regions/src/datasets/us/index.ts
```diff
@@ -0,0 +1,6 @@
+import states from "./states_db";
+import counties from "./counties_db";
+
+export { states, counties };
+export * from "./County";
+export * from "./State";
```
mikelehen: Per my other comments, we possibly don't even need to export `County` and `State` if we can make `Region` general enough that the rest of the code doesn't explicitly need to know about `County` and `State`.

packages/regions/src/datasets/us/states_db.test.ts
```diff
@@ -0,0 +1,27 @@
...
+import statesDB from "./states_db";
+
+describe("states_db", () => {
+  test("`all` includes all states", () => {
+    expect(statesDB.all).toHaveLength(53);
+  });
+
+  test("each state is correctly initialized", () => {
+    statesDB.all.forEach((state) => {
+      expect(state).toBeTruthy();
```
mikelehen, line 10: nit: This is redundant since the below assert would fail if it weren't truthy, so I'd delete it.

packages/regions/src/datasets/us/states_db.test.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  test("findByRegionId", () => {
+    expect(statesDB.findByRegionId("53")).toBeDefined();
+    expect(statesDB.findByRegionId("NO_STATE")).toBeNull();
+  });
+
+  test("findByRegionIdStrict", () => {
+    expect(statesDB.findByRegionIdStrict("53")).toBeDefined();
+    expect(statesDB.findByRegionIdStrict("53")).not.toBeNull();
+    expect(() => statesDB.findByRegionIdStrict("53")).not.toThrow();
```
mikelehen: nit: This is redundant since the above two lines verify it doesn't throw.

packages/regions/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../tsconfig",
+  "compilerOptions": {
+    "baseUrl": ".",
+    "declaration": true,
+    "declarationDir": "lib",
+    "outDir": "lib",
+    "rootDir": "src"
```
mikelehen, line 9: Just to be sure, we have to define these compilerOptions here?  We can't put them in the root tsconfig?  Guessing there's an issue with relative directories or something?

packages/regions/README.md
```diff
@@ -0,0 +1,13 @@
+# @actnowcoalition/regions
+
+> Regions
```
mikelehen: What's our stance on documenting our open source packages?  I think at least eventually it'd be nice if they had some minimal docs.  E.g. a description of what the package is for and a basic usage example.  But we can probably punt on that until things start to stabilize.  But maybe throw a description in here at least?

mikelehen (changes requested):
This looks great!  My main high-level feedback is I'd like to see how much functionality we can generalize into the `Region` class so that the rest of our code doesn't need to know about the `State` or `County` classes.  Ideally when we build new sites, we can create new region types as needed (e.g. `City` or `HospitalFacility` or whatever) and it'll be completely transparent to the rest of the code because it will all just deal with the `Region` class directly.

As always, holler if any of the feedback doesn't make sense or you think we should punt on it for now or whatever.

packages/regions/src/Region.ts
pnavarrc: We can migrate it once Metric is more stable, that makes sense!

packages/regions/src/datasets/us/County.ts
pnavarrc: Let's give it a try! Agree that it would be much cleaner if we could have a single `Region` class so location pages and other components don't need to be aware of variants.

packages/regions/README.md
pnavarrc: will do!

packages/regions/src/Region.ts
```diff
@@ -0,0 +1,56 @@
+import lowerCase from "lodash/lowerCase";
+import deburr from "lodash/deburr";
+import { urlJoin } from "./utils";
+
+export class Region {
+  constructor(
+    public readonly regionId: string,
```
smcclure17, line 7: Just for my understanding, we shouldn't have issues with duplicated/non-unique `regionId`s because different region datasets will use their own `RegionDB` instance (and we expect all the `regionIds` inside a single dataset will be unique)?

E.g., Country ID codes and US Health Service Area codes are both 3-digit numerics, and while these values overlap, it isn't an issue because we expect them to be handled in separate `RegionDB`s?

packages/regions/src/Region.ts
```diff
@@ -0,0 +1,56 @@
...
+    public readonly fullName: string,
+    public readonly shortName: string,
+    public readonly abbreviation: string,
+    public readonly urlFragment: string,
+    public readonly parent: Region | null,
+    // TODO: transfrom population into a Metric
```
smcclure17: Again for my understanding, the philosophy here is to add any additional attributes that a region might have as metrics?

smcclure17 (approved):
Looks great to me as far as I can tell! Just a couple of high-level questions

I assume adding countries would work similarly to adding states/counties? E.g, creating a RegionDB reading from a `countries.json` file, and adding parents to the state regions?

packages/regions/src/Region.ts
pnavarrc: I just added a note to the docs, I hope it's more clear now! I imagine that different projects will use different datasets, and that there could be overlap between the IDs used in the U.S. and IDs used in other countries, so I think it's reasonable to define unique IDs by `RegionID` rather than across all datasets in the package.

PR #12, opened 2022-07-18
Add toJSON and fromJSON mathods to Region class

smcclure17:


packages/regions/src/Region.ts
```diff
@@ -55,4 +65,28 @@ export class Region {
...
+  }
+
+  toJSON(): RegionJSON {
+    return {
+      regionId: this.regionId,
+      fullName: this.fullName,
+      shortName: this.shortName,
+      abbreviation: this.abbreviation,
+      urlFragment: this.urlFragment,
+      parent: this.parent,
```
pnavarrc: Also here, we probably want 

```tsx
// ...
parent: this.parent ? this.parent.toJSON() : null
// ...
```

Consider adding unit tests for this method to confirm that the JSON serialization works as intended.

packages/regions/src/Region.ts
```diff
@@ -55,4 +65,28 @@ export class Region {
...
   }
+
+  static fromJSON(regionJSON: RegionJSON): Region {
+    return new Region(
+      regionJSON.regionId,
+      regionJSON.fullName,
+      regionJSON.shortName,
+      regionJSON.abbreviation,
+      regionJSON.urlFragment,
+      regionJSON.parent,
```
pnavarrc: This should probably be 

```tsx
return new Region(
  // ...
  regionJSON.parent ? Region.fromJSON(regionJSON.parent) : null
  // ...
);
```

packages/regions/src/Region.ts
```diff
@@ -2,6 +2,16 @@ import lowerCase from "lodash/lowerCase";
 import deburr from "lodash/deburr";
 import { urlJoin } from "./utils";
 
+interface RegionJSON {
+  regionId: string;
+  fullName: string;
+  shortName: string;
+  abbreviation: string;
+  urlFragment: string;
+  parent: Region | null;
```
pnavarrc: The `parent` property should have type `RegionJSON | null`

pnavarrc (changes requested):
Just a few notes!

packages/regions/src/Region.test.ts
```diff
@@ -1,8 +1,46 @@
...
   });
 });
+
+describe("Region serialization", () => {
+  const parentJSON: RegionJSON = {
+    regionId: "12",
+    fullName: "Florida",
+    shortName: "Florida",
+    abbreviation: "FL",
+    urlFragment: "us/fl-florida",
```
pnavarrc: I think it should be only `us/florida-fl` - we are using the convention of splitting words with `_` (`miami_dade`) but separating "parts" with `-` (`florida-fl`). Everything works here, but it would be good to have an example that follows the conventions that we use for CAN

packages/regions/src/Region.test.ts
```diff
@@ -1,8 +1,46 @@
...
+    parent: null,
+    population: 100_000,
+  };
+
+  const regionJSON: RegionJSON = {
+    regionId: "12089",
+    fullName: "Miami-Dade County, Florida",
+    shortName: "Miami-Dade, FL",
+    abbreviation: "Miami-Dade Co.",
+    urlFragment: "us/fl-miami-dade",
```
pnavarrc: The `urlFragment` should only have the part of the URL that corresponds to the region. In this case, that would be `county/miami_dade`.

The `relativeUrl` method of `Region` uses the relative URL of the parent and the `urlFragment` to construct the relative URL of the region, in this case, that will be

```tsx
// /us/florida-fl/county/miami_dade
urlJoin(this.state.relativeUrl, `county/miami_dade`)
```

pnavarrc (approved):
Looks good! Just a few notes - Also, maybe increase the version number to `1.1.0` on `packages/regions/package.json` since this is a non-breaking API change.

packages/regions/src/Region.test.ts
smcclure17: ah, good flag, thanks 🙏 I've updated the locations to match the real/expected values in 2972f5565c94fef7795f5adcc0c6611912d72898

PR #14, opened 2022-07-19
Add metro areas to Regions package

smcclure17:
Example outputs: 

```typescript
Region {
  regionId: '14460',
  fullName: 'Boston-Cambridge-Newton metro area, MA-NH',
  shortName: 'Boston-Cambridge-Newton metro',
  abbreviation: 'Boston-Cambridge-Newton metro',
  urlFragment: 'boston_cambridge_newton',
  parent: null,
  population: 4873019
}
```

```typescript
Region {
  regionId: '10420',
  fullName: 'Akron metro area, OH',
  shortName: 'Akron metro',
  abbreviation: 'Akron metro',
  urlFragment: 'akron',
  parent: null,
  population: 703479
}

```

### DIFF
--- a/packages/regions/src/datasets/us/metros.json
+++ b/packages/regions/src/datasets/us/metros.json
@@ -1,0 +1,3155 @@
+[
+  {
+    "name": "Abilene",
+    "fipsCode": "10180",
+    "population": 172060,
+    "stateCode": ["48"],
+    "counties": ["48059", "48253", "48441"]
+  },
+  {
+    "name": "Aguadilla-Isabela",
+    "fipsCode": "10380",
+    "population": 289289,
+    "stateCode": ["72"],
+    "counties": [
+      "72003",
+      "72005",
+      "72011",
+      "72071",
+      "72081",
+      "72099",
+      "72117",
+      "72131",
+      "72141"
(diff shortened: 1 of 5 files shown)


packages/regions/src/datasets/us/metros.json
```diff
@@ -0,0 +1,3155 @@
+[
+  {
+    "name": "Abilene",
+    "fipsCode": "10180",
+    "population": 172060,
+    "stateCode": ["48"],
```
smcclure17, line 6: I imagine if we want to access the states/counties tied to each metro that we can add them as metrics

pnavarrc (approved):
Looks good! - Consider updating the README to describe where the regionId comes from.

packages/regions/src/datasets/us/metros.json
pnavarrc: Same for the counties, we might need to use the metro → county relationship in the future, but I think this is okay for now.

smcclure17:
> Consider updating the README to describe where the regionId comes from.

good flag, thanks! added in 076bf273f6afc69cc8d93c6110edaffe1f349a63

PR #15, opened 2022-07-20
Bump regions package version (1.1.4 -> 1.1.5)

smcclure17:
Increases the package  # so that we can republish with the changes from #14


pnavarrc (approved):
Looks good!

PR #16, opened 2022-07-20
Setup ui-components package

pnavarrc:
Setup ui-components package and implement the `LegendThreshold` (thermometer) component.

image

## Changes

- Setup the `ui-components` package.
- Setup storybook (inside the `ui-components` folder)
- Update the ESLint configuration to include `.storybook` (ESLint ignores folders starting with `.` by default)
- Implement the `LegendThreshold` component.
- Update `plopfile.mjs` to include a generator to create a component

## Notes

- We need to setup the theme in the `ui-components` package as well (and it should be consistent with the theme on the `act-now-template`). Ideally, we want to setup 2 different themes so we can verify that the components are picking the variables correctly.
- We need to improve the accessibility of this component. One way would be to have the screen reader describe each level, for example, it could read "level 1: from 0 to 10, level 2: from 10 to 20, ...). I'm going to leave that for a separated PR.
- Also for a follow-up PR: Setup Chromatic or storybook deployment.


packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
```diff
@@ -0,0 +1,85 @@
...
+  getItemValue,
+  getItemColor,
+  getItemLabel,
+  showLabels = true,
+}: LegendThresholdProps<T>) => {
+  const values = items.map(getItemValue);
+  const scaleRect = scaleBand({ domain: values, range: [0, width] });
+  const rectWidth = scaleRect.bandwidth();
+
+  const clipPathId = uniqueId(`rounded-borders-`);
```
chasulin: hm its not always `rounded-borders` though? if a borderRadius of 0 is provided..

packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+];
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const getItemValue = (item: Item, itemIndex: number) => item.value;
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const getItemColor = (item: Item, itemIndex: number) => item.color;
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const getItemLabel = (item: Item, itemIndex: number) => item.label;
+
+export const Example = Template.bind({});
```
chasulin: maybe something more descriptive, like `DefaultProps` or `Default` or `WithDefaultPropSettings`. something like that

packages/ui-components/.storybook/preview-head.html
```diff
@@ -0,0 +1,4 @@
+<link
+  rel="stylesheet"
+  href="https://fonts.googleapis.com/css2?family=Source+Code+Pro:wght@400;700&family=Roboto:wght@300;400;500;700&display=swap"
```
mikelehen, line 3: How do we imagine managing font dependencies in Act Now projects?  

Will we assume all projects use Roboto and Source Code Pro and therefore we should add them to the template [here](https://github.com/covid-projections/act-now-template/blob/0fe8e3c99408f174b7f40a7b3380a9bdd55d3731/src/pages/_document.tsx#L13)?

Or will it be project dependent and part of setting up the theme is adding font dependencies?  Maybe something we should document somewhere.

packages/ui-components/README.md
```diff
@@ -0,0 +1,13 @@
+# @actnowcoalition/ui-components
+
+> Ui components for Act Now
```
mikelehen: nit: "Ui" => "UI"

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
```diff
@@ -0,0 +1,85 @@
...
+  items: T[];
+  getItemValue: (item: T, itemIndex: number) => number;
+  getItemColor: (item: T, itemIndex: number) => string;
+  getItemLabel: (item: T, itemIndex: number) => string;
+  showLabels?: boolean;
+}
+
+/**
+ * `LegendThreshold` represents a scale with specific cut values that separate
+ * a set of levels. By default, the component shows labels between each level.
```
mikelehen, line 37: Perhaps document that it returns an svg group and therefore needs to be wrapped in an `<svg>` element?

As an aside, why did we do it this way?  Offhand I wouldn't know when to return a self-contained html component vs an svg group.  Should we start an internal "style guide" document that documents design choices like this?

packages/ui-components/src/index.ts
```diff
@@ -0,0 +1,2 @@
+// -- APPEND ITEMS HERE --
```
mikelehen: Not sure what this comment is for?

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
```diff
@@ -0,0 +1,85 @@
+import React from "react";
+import { Group } from "@visx/group";
+import { scaleBand } from "@visx/scale";
+import uniqueId from "lodash/uniqueId";
+import { TickLabel, TickMark } from "./LegendThreshold.style";
+
+export interface LegendThresholdProps<T> {
+  height?: number;
+  barHeight?: number;
```
mikelehen, line 14: Can we document these?  e.g. How do `height` and `barHeight` relate?

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
```diff
@@ -0,0 +1,85 @@
...
+  width?: number;
+  borderRadius?: number;
+  items: T[];
+  getItemValue: (item: T, itemIndex: number) => number;
+  getItemColor: (item: T, itemIndex: number) => string;
+  getItemLabel: (item: T, itemIndex: number) => string;
```
mikelehen: Again, documenting e.g. `getItemValue()` would be useful.  I'm not sure why we need it given we have getItemLabel() and getItemColor().  What is the value used for?

mikelehen (commented):
A few drive-by comments

packages/ui-components/.storybook/preview-head.html
pnavarrc: The typography should come from the theme, so it might be different from project to project. When implementing UI elements, I imagine that we could have 2 different themes (for testing) to confirm that the components use theme variables instead of hardcoded variables.

I discussed this with Josh and he is leaning towards using the same typography at the beginning and open that to customization later, if we need to. I will add a note to the roadmap doc!

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
pnavarrc: Good point! I don't think we have discussed this, but here is my thinking around that:

We will have mainly 2 kinds of components: building blocks and high-level components. 

- **Building blocks**: They are mostly used as parts of a more complex component. They need to be more flexible to account for several variants (with and without labels, etc).
- **High-level components**: They are more "standalone", and likely will take fewer customization options (if the component takes a metric, it will know from the metric whether or not it should include labels).

SVG building blocks might be used together to create a complex SVG visual, sometimes by adding overlays. In this case, if we return an SVG element, it won't render (SVG inside another SVG). For example, to build a metric chart (that shows lines and bars), the structure will be something like:

```tsx
// MetricChart
<svg width={width} height={height} >
  <BarChart />   
  <LineChart />
  <AxisLeft />
  <AxisBottom />
  <TooltipOverlay />
</svg>
```

but for the spark lines, it will be something like:

```tsx
// SparkLines
<svg width={width} height={height}>
  <BarChart />
  <LineChart />
</svg>
```

I think that ideally, high-level components should be wrapped in `div` or `Fragment` elements that are not aware of their context (shouldn't include margins, for example), and that take high-level constructs such as metrics or regions as opposed to lots of props. 

We should definitely open this for discussion though, and also figure out which components should belong to each category.

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
pnavarrc: I think you are right, maybe we don't need this! 🧹

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
pnavarrc: As I was writing this I realized that this particular component might not need to return a group, we don't have any instances where we need this to be part of an SVG element with other graphics. Still, we should discuss!

packages/ui-components/src/index.ts
pnavarrc: It was a marker for the code generator to append the line that exports the newly created component, I think it might be cleaner to just use the default (append at the end). We can later improve it and make sure that the exports are sorted

packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { Story, ComponentMeta } from "@storybook/react";
+import LegendThreshold, { LegendThresholdProps } from "./LegendThreshold";
+
+export default {
+  title: "Components/LegendThreshold",
+  component: LegendThreshold,
+} as ComponentMeta<typeof LegendThreshold>;
+
+interface Item {
+  value: number;
```
mikelehen: I think you can remove `value` now?

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
mikelehen: You gave a 👍 but I still don't see any documentation?

packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+const height = 40;
+const barHeight = 20;
+const borderRadius = 4;
+
+const items: Item[] = [
+  { value: 10, label: "10", color: "#90BE6D" },
+  { value: 20, label: "20", color: "#F9C74F" },
+  { value: 30, label: "30", color: "#F8961E" },
+  { value: 40, label: "40", color: "#E16420" },
+  { value: 50, label: "50", color: "#A10003" },
```
mikelehen: FWIW- It looks like this "50" does not get rendered:

image

I assume this is expected, but it does make this code a bit unintuitive.  Perhaps either:
1. The legend should render the last label if it's provided (even though it would look weird) and make `getItemEndLabel()` be able to return `undefined` so that we can omit it (probably always).
2. The component should document that it will never call `getItemEndLabel()` for the last item, since it won't be rendered.
3. The component could be refactored to accept `colors` and `labels` as two separate arrays with `colors` always being one longer in length than `labels`.

WDYT?

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
```diff
@@ -0,0 +1,95 @@
...
+  width?: number;
+  borderRadius?: number;
+  items: T[];
+  getItemColor: (item: T, itemIndex: number) => string;
+  getItemEndLabel: (item: T, itemIndex: number) => string;
+  showLabels?: boolean;
+}
+
+/**
+ * `LegendThreshold` represents a scale with specific cut values that separate
```
mikelehen: I can guess what "cut values" means, but it's not super clear terminology.  Maybe just say "thresholds" to match the component name?

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
mikelehen: Okay! I'm fine leaving this as is for now, but I agree that it seems like so far we haven't needed to compose the thermometer inside a larger SVG visual AFAIK.

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * **Notes**
+ *
+ * - This component returns an SVG Group element to make it easier to integrate
+ *   with other components inside an SVG element. Make sure to wrap it in an
+ *   SVG element if using it as standalone component.
+ * - The  `height` is the total height of the component (including labels),
+ *   whereas `barHeight` is the height of the colored bars only. When not adding
+ *   bars, make sure that `barHeight` is set to `height` (see Storybook for
+ *   some examples).
```
mikelehen: Ahh, I see you documented the parameters down here.  Any reason not to use normal `/** ... */` doc comments on the actual members of the interface?  I think that'll be more useful (VS Code should show it when you're setting the parameters, etc.).

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
```diff
@@ -0,0 +1,95 @@
...
+ * - Only the labels between two levels are rendered. The last end label is
+ *   not rendered.
```
mikelehen: Oh! I see this is documented as well.  Excellent.  Again, I would put this directly on the `getItemLabel` member.

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
```diff
@@ -0,0 +1,95 @@
...
+ */
+// eslint-disable-next-line @typescript-eslint/no-unnecessary-type-constraint
```
mikelehen: What's up with this?  I assume it's complaining about `extends unknown`?  Can you not just remove that?

Oh, I guess that gives an error because TypeScript tries to parse `<T>` as a JSX tag.  The standard workaround seems to be to add a comma after the type name: https://github.com/microsoft/TypeScript/issues/15713#issuecomment-499474386

So this would be:

```
const LegendThreshold = <T,>({
```

which isn't awesome but I think it's better than needing to disable an eslint warning.

plopfile.mjs
```diff
@@ -42,13 +42,69 @@ const templatePackage = prepareTemplate(`
...
+    "build": "tsc --project ./tsconfig.json"
   },
   "devDependencies": {
     "typescript": "^4.6.4"
   }
 }`);
 
+const templateTSConfig = prepareTemplate(`
+{
+  "extends": "../../tsconfig",
```
mikelehen, line 54: Should we update existing packages with this tsconfig?

mikelehen (approved):
Looks good with minor nits / questions.

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
pnavarrc: oh, I put that in the documentation for the component, moving it here!

.eslintrc.json
```diff
@@ -18,5 +18,6 @@
     "sourceType": "module"
   },
   "plugins": ["react", "@typescript-eslint"],
-  "rules": {}
+  "rules": {},
+  "ignorePatterns": ["!.storybook"]
```
mikelehen: While you're in here maybe we should ignore `dist` too?  I'm seeing a lot of errors from `yarn lint` at the repo root:

```
/Users/mikelehen/src/act-now-packages/packages/regions/dist/cjs/Region.js
  14:26  error  Unsafe usage of ThrowStatement                  no-unsafe-finally
  32:35  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  33:32  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  34:15  error  Require statement not part of import statement  @typescript-eslint/no-var-requires

/Users/mikelehen/src/act-now-packages/packages/regions/dist/cjs/RegionDB.js
  7:31  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  8:16  error  Require statement not part of import statement  @typescript-eslint/no-var-requires

/Users/mikelehen/src/act-now-packages/packages/regions/dist/cjs/datasets/us/counties_db.js
  6:34  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  7:16  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  8:39  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  9:35  error  Require statement not part of import statement  @typescript-eslint/no-var-requires

/Users/mikelehen/src/act-now-packages/packages/regions/dist/cjs/datasets/us/index.js
   7:35  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
   9:37  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  11:35  error  Require statement not part of import statement  @typescript-eslint/no-var-requires

/Users/mikelehen/src/act-now-packages/packages/regions/dist/cjs/datasets/us/metros_db.js
  6:34  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  7:16  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  8:37  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  9:35  error  Require statement not part of import statement  @typescript-eslint/no-var-requires

/Users/mikelehen/src/act-now-packages/packages/regions/dist/cjs/datasets/us/states_db.js
  6:37  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  7:16  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  8:34  error  Require statement not part of import statement  @typescript-eslint/no-var-requires

/Users/mikelehen/src/act-now-packages/packages/regions/dist/cjs/index.js
  21:16  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  23:34  error  Require statement not part of import statement  @typescript-eslint/no-var-requires
  25:14  error  Require statement not part of import statement  @typescript-eslint/no-var-requires

/Users/mikelehen/src/act-now-packages/packages/regions/dist/esm/Region.js
  13:26  error  Unsafe usage of ThrowStatement  no-unsafe-finally
```

packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx
pnavarrc: I added a note in the docstring for the component explaining that the last end label doesn't render (expected). Maybe we can revisit if we create a "threshold" abstraction for the metric?

Also I want to improve this component by making it accessible (ensuring that the screen reader describes each level, but for that levels will need to have names or something? not sure), but I'm not 100% what's the best practice accessibility-wise.

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
pnavarrc: oh this is awesome! I have been looking for this on the internet, thanks!

plopfile.mjs
pnavarrc: I think so (minus the react stuff), will do that on a follow-up PR (based of the recent changes on the regions package, so everything is tree shakeable)

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
pnavarrc: cc @chasulin ^

PR #17, opened 2022-07-25
Enable tree-shaking for regions package

pnavarrc:
Updates the `regions` package to enable [tree shaking](https://webpack.js.org/guides/tree-shaking/), a technique to only include what's being imported from a package instead of the entire package. Note that tree shaking can only be enabled when using ES Modules (`import` and `export`).

Something that some packages do is to build two versions of the package, one using ESM and one using CommonJS, so I added one `tsconfig` files (plus the base one). Maybe we could simplify and simply extend one from the other?

## Testing

I created location pages for states in a [branch in the act-now-templates repository](https://github.com/covid-projections/act-now-template/compare/pablo/location-pages?expand=1), and used `yarn build:analyze` to analyze the bundle size. If tree shaking is enabled, counties and metros shouldn't be included in the final bundle.

### Before (664.92 KB) 🌳 

The bundle includes counties, metros and the entire lodash library. See the commit https://github.com/covid-projections/act-now-template/commit/ba7c7906c4a2873f0c409a2f7de705ef106ead9a

image

### After (322.91 KB) 🌱 🍃 

Now counties, metros and lodash are not included, the bundle size is about half the version without tree shaking. See the commit https://github.com/covid-projections/act-now-template/commit/5158c47bbab4deaf134b3357c202e09951bd0b82

image


packages/regions/tsconfig.node.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.base",
+  "compilerOptions": {
+    "module": "CommonJS",
```
mikelehen: Out of curiosity, is there a particular reason we want / need to generate CommonJS modules?  Or is it just a best practice or something?

I guess maybe it is for node, though node seems to support EcmaScript Modules as well (see e.g. https://www.the-guild.dev/blog/support-nodejs-esm).  But I guess maybe support is newer (v12?) and maybe fragile?  Not 100% sure.

packages/regions/package.json
```diff
@@ -12,13 +12,16 @@
...
+  "module": "dist/esm/index.js",
+  "types": "dist/index.d.ts",
   "files": [
-    "lib"
+    "dist"
   ],
   "scripts": {
-    "build": "tsc --project ./tsconfig.json"
+    "build:browser": "tsc --project ./tsconfig.browser.json",
+    "build:node": "tsc --project ./tsconfig.node.json",
```
mikelehen: Is "browser" and "node" the best way to refer to these builds?  Or should it be build:cjs and build:esm or something?  I don't care strongly if there's a clear precedent here, but it feels like we might be mix/matching two related but different concepts (CJS vs ESM and node vs browser)

mikelehen (approved):
Nice, looks good!  Left a few question comments, but if this is working then it's probably fine to proceed!

packages/regions/tsconfig.node.json
pnavarrc: I think it's just best practice to support projects that are not configured to use CommonJS - one example is `url-join`, which only exposes ESM modules, which made Jest break (with the default config)

mikelehen: Ah!  Good point.  That was weird / annoying!

PR #19, opened 2022-07-26
ui-components: LegendThreshold is now an SVG element

pnavarrc:
Minor updates to the `LegendThreshold` component and `tsconfig` of the `ui-components` package.

## Changes

- The `LegendThreshold` component is now a SVG element (based on PR feedback, also, we haven't used that component within an SVG element before, so exposing it as an SVG feels more appropiate.
- Update the `tsconfig` of the `ui-components` package to generate CommonJS and ESModules (see https://github.com/covid-projections/act-now-packages/pull/17 where this is done for the `regions` package)
- Extends the `LegendThreshold` props to include SVG props, which are passed down to the inner SVG element. This allows the component to be wrapped in `styled` and also to pass SVG attributes (margins, etc) to the component without having to wrap it on a `div` (added an example on the stories as well).

### DIFF
--- a/packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx
+++ b/packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx
@@ -1,7 +1,6 @@
 import React from "react";
 import { Story, ComponentMeta } from "@storybook/react";
 import LegendThreshold, { LegendThresholdProps } from "./LegendThreshold";
-import { styled } from "../../styles";
 
 export default {
   title: "Components/LegendThreshold",
@@ -70,18 +69,3 @@ Squared.args = {
   borderRadius: 0,
   showLabels: true,
 };
-
-const StyledLegendThreshold = styled(LegendThreshold)`
-  line {
-    stroke: ${({ theme }) => theme.palette.grey[50]};
-  }
-  text {
-    fill: ${({ theme }) => theme.palette.grey[50]};
-  }
-`;
-
-export const StyledExample = () => (
-  <div style={{ backgroundColor: "#555", padding: 16 }}>
-    <StyledLegendThreshold {...DefaultProps.args} onClick={() => "clicked"} />
-  </div>
-);
--- a/packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx
+++ b/packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx
@@ -82,6 +82,6 @@ const StyledLegendThreshold = styled(LegendThreshold)`
 
 export const StyledExample = () => (
   <div style={{ backgroundColor: "#555", padding: 16 }}>
-    <StyledLegendThreshold {...DefaultProps.args} />
+    <StyledLegendThreshold {...DefaultProps.args} onClick={() => "clicked"} />
   </div>
 );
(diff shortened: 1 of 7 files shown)


packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx
```diff
@@ -71,3 +70,18 @@ Squared.args = {
...
   showLabels: true,
 };
+
+const StyledLegendThreshold = styled(LegendThreshold)`
+  line {
+    stroke: ${({ theme }) => theme.palette.grey[50]};
+  }
+  text {
+    fill: ${({ theme }) => theme.palette.grey[50]};
+  }
```
mikelehen: Do we actually expect to re-style the component like this?  I'm not sure if there's a best practice here, but this is making assumptions about the implementation of the component (that it uses `line` and `text` SVG elements and that it's okay to override their `stroke` / `fill` properties directly).

Normally it's a best practice not to expose implementation details to the consumers of a software component, but I'm not sure what the best practice for styling React components is.  Naively I'd maybe have `strokeColor` and `fillColor` properties on the component, but that likely isn't ideal either.  Is there a clear precedent from material-ui that we should follow?

packages/ui-components/tsconfig.base.json
```diff
@@ -3,13 +3,11 @@
...
-    "declarationDir": "lib",
+    "declarationDir": "./dist",
     "noEmit": false,
-    "outDir": "lib",
     "rootDir": "src",
-    "module": "commonjs",
     "jsx": "react"
   },
-  "include": ["src/**/*.ts", "src/**/*.tsx", "src/**/*.json"],
+  "include": ["src/**/*.ts", "src/**/*.json"],
```
mikelehen: Hrm.  Why do we not need tsx here?

mikelehen (approved):
Looks good.  One kinda' big question about styling best practices but we can punt on it.

packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx
pnavarrc: Good point, the example is probably not what we want. I think we want to control text and marker colors via the theme (or component props in other situations), not by using styled and targeting internal elements.

We _might_ want to use `styled` in a few limited cases, for example, if we need to change positioning, margins or display properties (let's say `align-self` in a flex container) without wrapping the component on a `div`, but that's probably not worth the trouble.

I actually used styled to confirm that `LegendThreshold` will receive SVG props in addition to the props defined in `LegendThresholdProps`, the idea is that we can add any valid SVG attribute to `LegendThreshold` and it will be applied correctly. This can be used to add `onClick` handlers, ARIA attributes, class names, etc (styled needs to pass a class name for it to work).

I think that you are right that in general, assuming implementation details about components is not the best practice, however, MUI does use internal class names (documented and part of their API) to allow customization of inner components using CSS-in-JS. That being said, we probably don't want (or need) to do that here.

I'm going to remove the styled example and keep the `onClick` example to show that we can pass SVG attributes directly to `LegendThreshold`.

packages/ui-components/tsconfig.base.json
pnavarrc: Apparently  `.ts`, `.tsx`, and `.d.ts` extensions are enabled by default (I removed `.tsx` by mistake actually), I'm going to add it back to restrict the path to `src`, good catch! https://www.typescriptlang.org/tsconfig#include

PR #20, opened 2022-07-27
Initial metrics package with Timeseries class.

mikelehen:


packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -0,0 +1,260 @@
...
+import first from "lodash/first";
+import last from "lodash/last";
+import maxBy from "lodash/maxBy";
+import minBy from "lodash/minBy";
+
+/**
+ * A single point in a timeseries containing a date (which must not contain a
+ * timestamp) and a value of type T.
+ */
+export interface TimeseriesPoint<T = unknown> {
```
mikelehen, line 13: Note that `Timeseries` and `TimeseriesPoint` are generic so you could e.g. have a `Timeseries<string>`.  I am not 100% sure if we need this flexibility.  Most timeseries will be numeric data, and obviously charts will rely on numeric data... 

But this is what I had implemented in the who-demo repo already and I think it may be useful to keep.  GCAT / 3RT used string values for representing some metrics (e.g. "On Track", "Goal Met", etc.).  And we used structured objects for some CAN metrics (`{rt, low, high}`).

If you want to convert to numeric timeseries, you can use the `Timeseries.assertFiniteNumbers()` helper.

packages/metrics/src/Timeseries/timeseries_utils.ts
```diff
@@ -0,0 +1,73 @@
...
+ * returned timeseries will truncate them).
+ * @returns A new Timeseries computed via a rolling average of the provided timeseries.
+ */
+export function rollingAverage(
+  timeseries: Timeseries<number | null>,
+  days = 7,
+  includeTrailingZeros = true
+): Timeseries<number | null> {
+  // TODO(michael): Ideally we could handle gaps in dates but for now just
+  // assert there are none.
```
mikelehen, line 20: I decided it wasn't worth the effort to rewrite this to work with gaps right now, but I think ultimately I'd like all our code to work with sparse timestamps, so I don't love this limitation.

packages/metrics/src/Timeseries/timeseries_utils.ts
```diff
@@ -0,0 +1,73 @@
...
+ * Returns a new Timeseries computed via a rolling average of the provided `timeseries`.
+ *
+ * @param timeseries The timeseries to compute a rolling average of.
+ * @param days The number of days to include in the rolling average.
+ * @param includeTrailingZeros Whether to include trailing zero points (else the
+ * returned timeseries will truncate them).
+ * @returns A new Timeseries computed via a rolling average of the provided timeseries.
+ */
+export function rollingAverage(
+  timeseries: Timeseries<number | null>,
```
mikelehen, line 15: Note that since `rollingAverage()` only works for numeric data, I made it a separate function instead of putting it directly on the `Timeseries<T>` class which might have non-numeric data.  This also makes it tree-shakable.  

Not 100% sure if I love this (harder to find the `rollingAverage()` function than if it was just a method on the class), so holler if you have thoughts / concerns.

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -0,0 +1,283 @@
...
+   * The returned timeseries will cast to `Timeseries<number>` so any subsequent
+   * code doesn't need to deal with non-number values.
+   */
+  assertFiniteNumbers(): Timeseries<number> {
+    this.points.forEach((p) => {
+      // TODO(michael): Won't allowing nulls break the return type?
+      assert(
+        isFinite(p.value),
+        `Found non-numeric (or non-finite) value in timeseries: ${p.value}`
+      );
```
pnavarrc, line 161: It might be useful to also log the date corresponding to the offending point

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -0,0 +1,283 @@
...
+};
+
+/**
+ * An immutable series of dates and values, representing a value that changes
+ * over time.
+ */
+export class Timeseries<T = unknown> {
+  constructor(readonly points: Array<TimeseriesPoint<T>>) {
+    Timeseries.assertDatesHaveNoTimeComponent(points);
+  }
```
pnavarrc, line 49: Maybe it would be useful to sort the points by date when we initialize?

packages/metrics/src/Timeseries/timeseries_utils.ts
pnavarrc: This works for now!

packages/metrics/src/Timeseries/timeseries_utils.ts
```diff
@@ -0,0 +1,73 @@
...
+  timeseries.points.forEach((p) => {
+    assert(
+      !nextDate || nextDate.getTime() === p.date.getTime(),
+      `Timeseries has gaps in dates (${nextDate} is missing)`
+    );
+    const datePlusOne = new Date(p.date);
+    datePlusOne.setUTCDate(datePlusOne.getUTCDate() + 1);
+    nextDate = datePlusOne;
+  });
+}
```
pnavarrc, line 73: The functions here assume that the points in the time series are sorted, maybe we should assert or sort on initialization?

pnavarrc (approved):
Looks good! Added a few non-blocking notes

packages/metrics/src/Timeseries/Timeseries.ts
mikelehen: I like that idea.  Done!

packages/metrics/src/Timeseries/Timeseries.ts
mikelehen: Good call, done!

packages/metrics/src/Timeseries/timeseries_utils.ts
mikelehen: Yep, good call. I added the sorting on initialization.

PR #23, opened 2022-08-01
ui-components: add LegendCategorical

chasulin:
A legend of color blocks and corresponding labels

**Desktop/screens `md` and wider:** 
Screen Shot 2022-08-01 at 5 07 04 PM

**Mobile/screens narrower than `md`:**
Screen Shot 2022-08-01 at 5 07 16 PM


pnavarrc (approved):
Looks good! I suspect that we might need to pass the orientation as prop in the future (as opposed to determine that based on the screen width), but that should be easy to adapt if and when it happens.

chasulin:
> Looks good! I suspect that we might need to pass the orientation as prop in the future (as opposed to determine that based on the screen width), but that should be easy to adapt if and when it happens.

true, updated! @pnavarrc

PR #24, opened 2022-08-02
ui-components: Implement ProgressBar

pnavarrc:
Implements a `ProgressBar` component. 

image

### Notes

- This component could be arguably be called `Meter`, however, it could be confused with the native `<meter>` element, which has additional functionality (it has low and high levels), so I decided to keep the name that we have been using for now, but using the role `meter`, which is a better fit for our use case (see https://www.w3.org/WAI/ARIA/apg/patterns/meter/)
- VoiceOver reads the component as "Percent Vaccinated, 75, level indicator"


packages/ui-components/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,87 @@
+import { scaleLinear } from "@visx/scale";
+import { Group } from "@visx/group";
+import uniqueId from "lodash/uniqueId";
+import React from "react";
+
+export interface ProgressBarProps {
+  /** Border radius of the progress bar */
+  borderRadius?: number;
+  /** Minimum value in the range (in user units) */
+  minValue: number;
```
mikelehen: FWIW- We could probably make this optional and default to zero since that's probably the 98% use case.  Not opposed to keeping it explicit if you prefer though.

packages/ui-components/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,87 @@
...
+  /** Maximum value in the range (in user units) */
+  maxValue: number;
+  /** Value (in user units) */
```
mikelehen: Maybe expand to "Current progress value (in user units)"

packages/ui-components/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,87 @@
+import { scaleLinear } from "@visx/scale";
+import { Group } from "@visx/group";
+import uniqueId from "lodash/uniqueId";
+import React from "react";
+
+export interface ProgressBarProps {
+  /** Border radius of the progress bar */
+  borderRadius?: number;
+  /** Minimum value in the range (in user units) */
```
mikelehen: nit: Since "in user units" applies to all the values and you explain it in more detail on the overall ProgressBar comment, I'm not sure if it's adding much to include it on each property.  But it's not really harmful either, so feel free to leave as-is.

packages/ui-components/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,87 @@
...
+  backgroundColor = "rgba(95, 108, 114, 0.2)",
+  value,
+  color,
+  role = "meter",
+  ...otherSvgProps
+}: ProgressBarProps &
+  Omit<
+    React.SVGProps<SVGSVGElement>,
+    "aria-valuemin" | "aria-valuemax" | "aria-valuenow"
+  >) => {
```
mikelehen, line 50: Wow, nice work with the precise type definitions! 🏆

packages/ui-components/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const xScale = scaleLinear({
+    domain: [minValue, maxValue],
+    range: [0, width],
+  });
+
+  const clipPathId = uniqueId(`rounded-borders-clippath-`);
```
mikelehen, line 56: is the trailing `-` intentional?

mikelehen (approved):
LGTM!

packages/ui-components/src/components/ProgressBar/ProgressBar.tsx
pnavarrc: It is, just so the generated id is like `rounded-borders-clippath-1`, etc.

PR #25, opened 2022-08-02
Update tsconfig files

faihegberg:
Adjust our plop file to auto-generate the updated tsconfig files whenever one creates a new package.


pnavarrc (changes requested):
We also need to update package.json to include `module` and `main`

pnavarrc (approved):
Looks good!

PR #26, opened 2022-08-02
Remove 'us' prefix from the states urlFragment

pnavarrc:
Not the definitive version of how we are going to handle urlParams/routing, but it would be useful to merge this to prototype a few solutions.


mikelehen (approved):
LGTM.  I'd be tempted to also rename `urlFragment` to just `slug` or similar so that Region isn't implying how it should be used.  Though I can't immediately think of other uses besides URLs, so feel free to leave it as-is. 😁

PR #27, opened 2022-08-03
Set up documentation generation with TypeDoc.

mikelehen:
Generates API docs using TypeDoc.  For now you have to do it locally, but I'd like to set this up with a github action that publishes the docs online somewhere.

To test:
```
$ yarn build:all
$ yarn generate-docs
$ open docs/index.html
```

Browse around the docs:
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,3 +18,4 @@ jobs:
       - run: yarn run lint
       - run: yarn workspaces run build
       - run: yarn run test:ci
+      - run: yarn generate-docs
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
     "lint": "eslint .",
     "build:all": "yarn workspaces run build",
     "create-package": "yarn plop package",
-    "create-component": "yarn plop component"
+    "create-component": "yarn plop component",
+    "generate-docs": "yarn typedoc --entryPointStrategy packages --name 'Act Now Packages' ."
   },
   "lint-staged": {
     "*.{ts,tsx}": "eslint --max-warnings=0",
@@ -41,6 +42,7 @@
     "ts-jest": "^28.0.2",
     "ts-node": "^10.7.0",
     "tsconfig-paths": "^4.0.0",
+    "typedoc": "^0.23.10",
     "typescript": "^4.6.4"
   },
   "dependencies": {}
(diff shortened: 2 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -18,3 +18,4 @@ jobs:
       - run: yarn run lint
       - run: yarn workspaces run build
       - run: yarn run test:ci
+      - run: yarn generate-docs
```
mikelehen, line 21: I might move this to a separate job that builds the docs and publishes them somewhere, but this at least makes sure they can be built.

package.json
```diff
@@ -19,7 +19,8 @@
     "lint": "eslint .",
     "build:all": "yarn workspaces run build",
     "create-package": "yarn plop package",
-    "create-component": "yarn plop component"
+    "create-component": "yarn plop component",
+    "generate-docs": "yarn typedoc --entryPointStrategy packages --name 'Act Now Packages' ."
```
mikelehen, line 23: Note that you may need to run `build:all` first (due to how it finds your source code using the "main" entry point of your modules I think), but I didn't want to make CI build the modules twice so I didn't do it automatically here.

packages/metrics/src/Timeseries/index.ts
```diff
@@ -1,4 +1,8 @@
 export { Timeseries } from "./Timeseries";
-export type { NonEmptyTimeseries } from "./Timeseries";
+export type {
+  DateRange,
+  NonEmptyTimeseries,
+  TimeseriesPoint,
+} from "./Timeseries";
```
mikelehen, line 6: We had a few types that ended up referenced by exported code but that were not being exported themselves and so typedoc was throwing errors / warnings.  I have fixed them.

tsconfig.json
```diff
@@ -18,7 +18,8 @@
     "isolatedModules": true,
     "jsx": "preserve",
     "incremental": true,
-    "downlevelIteration": true
+    "downlevelIteration": true,
+    "sourceMap": true
```
mikelehen, line 22: I think TypeDoc finds your source files by looking for a `sourceMappingURL` in the file that is referenced as `main` in your package.json, so I had to enable source map generation.  But we probably want this anyway.  It may make it easier to debug code, etc.

pnavarrc (approved):
Awesome! 📚

PR #28, opened 2022-08-03
UI components: region search (basic)

chasulin:
Implements a simple version of region search (metric un-aware, no colored indicators. only region search).

`RegionSearch` in storybook

### DIFF
--- a/packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
+++ b/packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
@@ -1,0 +1,47 @@
+import React from "react";
+import { Container } from "./RegionSearch.style";
+import { Region } from "@actnowcoalition/regions";
+import { Autocomplete, AutocompleteProps, TextField } from "@mui/material";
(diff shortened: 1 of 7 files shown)


packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
```diff
@@ -0,0 +1,59 @@
...
+const onChange = (item: Region | null) => {
+  if (!item) {
+    return null;
+  } else {
+    window.location.href = item.relativeUrl;
+  }
+};
+
+const RegionSearch: React.FC<
+  RegionSearchProps & AutocompleteProps<Region, false, false, false, "div">
```
faihegberg: `<Region, false, false, false, "div">` seems a bit confusing. Perhaps add a comment here as to what each represents?

chasulin: hovering on `AutocompleteProps` lets you know what each param represents. added some comments as well

faihegberg (approved):

PR #30, opened 2022-08-03
Metrics API

mikelehen:
Includes core metric functionality: metrics, metric catalog, metric data providers, metric data representation classes (MetricData, MultiMetricDataStore, etc.).

For overview see: https://www.dropbox.com/scl/fi/ir46x98uwgchsiais2pp1/Metrics-Overview.paper?dl=0&rlkey=0shjzfe0lzb6y46kcov7s15xc
For API docs see: https://act-now-packages--pr30-mikelehen-metrics-sk-iitqlhzm.web.app/docs/index.html

TODO:
- [x] Add some basic docs to README.md


packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -0,0 +1,286 @@
...
+   * @param region The region to get data for.
+   * @param metrics The metrics to get data for as either strings or `Metric` objects.
+   * @param includeTimeseries Whether to fetch timeseries data or not.
+   * @returns The metric data.
+   */
+  useDataForMetrics(
+    region: Region,
+    metrics: string[] | Metric[],
+    includeTimeseries = true
+  ): DataOrError<MultiMetricDataStore> {
```
mikelehen, line 225: I started to refactor these `use...()` hooks to share more code via a `useAsyncDataFromFetcher()` helper, but I forgot about the JSON.stringify() hack below which made it less straightforward to reuse code.  I think there's a better way to handle that (https://stackoverflow.com/a/66653729).

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -0,0 +1,289 @@
...
+   * @param metrics The metrics to include in the catalog.
+   * @param dataProviders The data providers used to fetch data for metrics.
+   * @param options Additional options for the catalog.
+   */
+  constructor(
+    metrics: MetricDefinition[],
+    dataProviders: MetricDataProvider[],
+    options: MetricCatalogOptions = {}
+  ) {
+    this.metrics = metrics.map((metric) => new Metric({ ...metric }, options));
```
pnavarrc: nit: It could be just `new Metric(metric, options)` right? (unless we need to clone the object?)

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -0,0 +1,289 @@
...
+    this.metricsById = keyBy(this.metrics, "id");
+
+    const referencedDataProviderIds = metrics.map(
+      (m) => m.dataReference?.providerId
+    );
+    const missingDataProviderIds = referencedDataProviderIds.filter(
+      (id) => !dataProviders.find((provider) => provider.id === id)
+    );
```
pnavarrc, line 57: nit: this code could be a bit simpler if `dataProvidersById` is initialized before trying to find the missing data providers

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -0,0 +1,289 @@
...
+        })
+        .catch((error) => {
+          console.error(
+            `Error fetching metric data for ${metric} for ${region.regionId}: ${error}`
+          );
+          return { error };
+        });
+    }, [region, metric]);
+    return { data };
+  }
```
pnavarrc, line 211: For later, it might be useful to have a loading state. 

```tsx
const { isLoading, data, error } = useData(region, metric);

if (isLoading) {
  // render loading state...
}

// handle error or render with data
```
Alternatively, we can always infer the loading state

```tsx
const { data, error } = useData(region, metric)
const isLoading = isNil(data) && isNil(error);

// ...
```

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -0,0 +1,289 @@
...
+    region: Region,
+    metrics: string[] | Metric[],
+    includeTimeseries = true
+  ): DataOrError<MultiMetricDataStore> {
+    const [data, setData] = useState<MultiMetricDataStore>();
+    // TODO(michael): This is a hack. We have to turn metricList into a string
+    // to pass to the useEffect hook since useEffect doesn't do deep equality on
+    // arrays. We probably need to do something similar for regions or else find
+    // another workaround.
+    const metricList = JSON.stringify(metrics);
```
pnavarrc, line 231: Maybe we can rely on sorted metricIds (same for regions)?

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -0,0 +1,289 @@
...
+        .then((metricDataStore) => {
+          setData(metricDataStore);
+        })
+        .catch((error) => {
+          console.error(
+            `Error fetching metric data for ${region.regionId}: ${error}`
+          );
+          return { error };
+        });
+    }, [region, metrics]);
```
pnavarrc: We need to pass `metricList` here instead of `metrics` then right?

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -0,0 +1,289 @@
...
+      const metrics = JSON.parse(metricList);
+      this.fetchDataForMetricsAndRegions(regions, metrics, includeTimeseries)
+        .then((multiRegionMetricDataStore) => {
+          setData(multiRegionMetricDataStore);
+        })
+        .catch((error) => {
+          console.error(`Error fetching metric data: ${error}`);
+          return { error };
+        });
+    }, [metricList, regions]);
```
pnavarrc, line 277: We probably need to implement the hack for `regions` too here

packages/metrics/src/data_providers/MockDataProvider.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    return this.cachedData[cacheKey];
+  }
+}
+
+/**
+ * Generates a timeseries using a sine wave with random magnitude and phase.
+ */
+function mockTimeseries(
```
pnavarrc: Can we export this so we can use it for storybook or testing? 😬

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -0,0 +1,289 @@
...
+        .then((data) => {
+          setData(data);
+        })
+        .catch((error) => {
+          console.error(
+            `Error fetching metric data for ${metric} for ${region.regionId}: ${error}`
+          );
+          return { error };
+        });
+    }, [region, metric]);
```
pnavarrc, line 209: Maybe `region.id` and `metric.id` instead?

pnavarrc (approved):
Looks great! 🙌  Added a few non-blocking suggestions

PR #31, opened 2022-08-03
Explicitly list packages in workspace so we can control the build order.

mikelehen:
The metrics package stopped building when I added a dependency on regions, because yarn was building
metrics before regions. I'm explicitly listing the packages so we can control the build order.


pnavarrc (approved):
Looks good!

PR #32, opened 2022-08-03
Markdown component

faihegberg:
Add markdown component to Act Now Packages.


packages/ui-components/src/components/Markdown/Markdown.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+`;
+
+const StyledTemplate: ComponentStory<typeof StyledMarkdown> = (args) => (
+  <StyledMarkdown {...args} />
+);
+
+export const StyledExample = StyledTemplate.bind({});
+StyledExample.args = {
+  body: "Styled markdown example!",
+};
```
pnavarrc: Consider adding more complex markdown content in the examples (h1-h4, bold, italics, quote, images, code, etc.), and maybe also include things that are enabled by the GFM plugin, since we are including it (tables I think, maybe more?)

pnavarrc: Something like this (but we probably need to combine additional `remarkPlugins` passed as props with the `remarkGfm` plugin):

```tsx
const Markdown: React.FC<ReactMarkdownOptions> = ({ children, ...otherProps }) => (
  <MarkdownBody remarkPlugins={[remarkGfm]} {...otherProps}>
    {children}
  </MarkdownBody>
);
```

Also, consider including the `InlineMarkdown` component that we implemented for GCAT, that comes in handy when we need to format block elements (like headers) that are not divs

pnavarrc (changes requested):
Suggested a few changes to make the component more flexible. Also, consider including the `InlineMarkdown` component so we can use it for non-div elements

packages/ui-components/src/components/Markdown/Markdown.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+  children:
+    "# Heading 1 \n ## Heading 2 \n ### Heading 3 \n I am **bold!** \n\n I am *italicized!*",
+};
+
+// Quote example
+export const QuoteExample = Template.bind({});
+QuoteExample.args = {
+  children:
+    "> I am a nested quote.\n\n>> Including sub-points... \n\n>> - Point 1 is plain. \n\n>> - `Point 2 is code` \n\n>> - Point 3 has an image underneath. \n\n>> ![Cat](https://pbs.twimg.com/profile_images/664169149002874880/z1fmxo00_400x400.jpg)",
+};
```
pnavarrc: If you use backticks \` instead of `"` for children you can just use regular newlines for the examples (see `tableContent` below). Also (nit), maybe add some newlines between statements.

pnavarrc (approved):
Looks good! I made a few minor tweaks

PR #34, opened 2022-08-04
Theming in ui-components package

chasulin:
Sets up MUI theme extensions according to our [figma design system](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=29%3A1035) and adds to typography+palette accordingly. Putting the MUI theme extension declarations in `ui-components/src/index.ts` should result in them showing up in any repo that has the `ui-components` package installed, and therefore should apply the custom theming to that repo. <-- Going to test this out by merging/publishing this PR and testing custom theming in another repo

As of now, custom theme variables are working properly when used to style components in this package (see the custom themed `variant="paragraphSmall"` [here](https://github.com/covid-projections/act-now-packages/pull/34/files#diff-1ff454cd5c955105ef616d90f64d202ff509436f688a484c1546a9eecf72689eR40)). We want the same custom theme variables to be applied to any `act-now-template` project's theme (and we want to control the shape of the custom theme from a single location - [hopefully here](https://github.com/covid-projections/act-now-packages/pull/34/files#diff-e8db90eaf056409aa60f99ae7e58bc599864cd6e284b72fcade7570c4697b1f8R15-R54)).

### DIFF
--- a/packages/ui-components/src/styles/theme/interfaces.ts
+++ b/packages/ui-components/src/styles/theme/interfaces.ts
@@ -1,0 +1,8 @@
+/** Theme interfaces */
+import { TypographyOptions } from "@mui/material/styles/createTypography";
+
+/** Typography */
+export interface ExtendedTypographyOptions extends TypographyOptions {
+  paragraphSmall: React.CSSProperties;
+  paragraphLarge: React.CSSProperties;
+}
--- a/packages/ui-components/src/components/LegendCategorical/LegendCategorical.style.ts
+++ b/packages/ui-components/src/components/LegendCategorical/LegendCategorical.style.ts
@@ -10,7 +10,3 @@ export const Square = styled("div", { shouldForwardProp: isValidProp })<{
   margin-right: ${(props) => props.theme.spacing(1)};
   background-color: ${(props) => props.color};
 `;
-
-export const Label = styled("span")`
-  line-height: 1.2;
-`;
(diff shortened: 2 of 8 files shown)


packages/ui-components/src/styles/theme/interfaces.ts
```diff
@@ -0,0 +1,8 @@
+/** Theme interfaces */
+import { TypographyOptions } from "@mui/material/styles/createTypography";
+
+/** Typography */
+export interface ExtendedTypographyOptions extends TypographyOptions {
```
mikelehen, line 5: I don't care strongly, but if we don't need to export this outside our package (which I think we don't), I think you could just move this into `typography.ts`

Feel free to leave here if you find that cleaner or think this `interfaces.ts` file is going to grow over time and be more substantial.

mikelehen (approved):
One small question, but otherwise looks good!

packages/ui-components/src/styles/theme/interfaces.ts
chasulin: my thought was that we'd declare the type in the external project's typography file as well [(like here)](https://github.com/covid-projections/act-now-packages/pull/34/files#diff-013e9da9022680fe2e069cd9354268b07552a79713cf5dc68906976d0086e723R18)

mikelehen: Oh, right.  They'll need to override the theme and this will help doing that?  Makes sense.

PR #36, opened 2022-08-04
Add Snapshot/Cache structure for MetricCatalog

smcclure17:
Adds the ability to:
- Persist `MultiRegionMultiMetricDataStore`s to JSON (to create a 'snapshot')
- Pass this snapshot into `MetricCatalog`s to allow metrics to be loaded from the snapshot instead of using their `DataProvider`


packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -92,14 +119,28 @@ export class MetricCatalog {
...
   async fetchDataForMetrics(
     region: Region,
     metrics: string[] | Metric[],
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    includeTimeseries = true
+    includeTimeseries = true,
+    useSnapshot = true
   ): Promise<MultiMetricDataStore> {
+    if (this.snapshot !== undefined && useSnapshot) {
+      const dataMap: MetricToDataMap<unknown> = {};
```
smcclure17: Not sure how we should handle types/generics here, but this is probably not it :/

packages/metrics/src/MetricCatalog/MetricCatalog.test.ts
```diff
@@ -31,7 +47,10 @@ describe("MetricCatalog", () => {
           },
         },
       ],
-      dataProviders
+      dataProviders,
+      undefined,
+      snapshot
```
mikelehen: Thoughts on adding `snapshot` to `MetricCatalogOptions` instead of making it another constructor argument?  I don't have strong feelings, but since it's optional, I'm kind of inclined to.

packages/metrics/src/MetricCatalog/MetricCatalog.test.ts
```diff
@@ -31,7 +47,10 @@ describe("MetricCatalog", () => {
...
     );
+    catalog.fetchData(states.findByRegionIdStrict("12"), "the-answer");
```
mikelehen: 1. Is there a reason we're not checking that the returned data is as expected (value 42)?
2. Even if we were checking, I assume this would pass with or without the snapshot, since it would just get fetched from the data provider.  I think it might make sense to add a test that explicitly tests snapshot functionality and in that case specify a metric *without* a data provider, but use a snapshot, and make sure that we get back the correct data from the snapshot.  WDYT?

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -12,6 +12,10 @@ import { MultiMetricDataStore } from "../data/MultiMetricDataStore";
 import { MultiRegionMultiMetricDataStore } from "../data/MultiRegionMultiMetricDataStore";
 import { MetricDefinition } from "../Metric/MetricDefinition";
 import { MetricCatalogOptions } from "./MetricCatalogOptions";
+import { SnapshotJSON } from "../data/MultiRegionMultiMetricDataStore";
+import { Timeseries } from "../Timeseries";
+import { MetricToDataMap } from "../data/MultiMetricDataStore";
+import { MultiRegionMultiMetricDataMap } from "../data/MultiRegionMultiMetricDataStore";
```
mikelehen: nit: We have several imports from "../data/MultiRegionMultiMetricDataStore" Can we combine them?

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -75,14 +84,32 @@ export class MetricCatalog {
    * @param region The region to get data for.
    * @param metric The metric to get data for as either a string or `Metric` object.
    * @param includeTimeseries Whether to fetch timeseries data or not.
+   * @param useSnapshot Whether to try to read metric data from this.snapshot if it exists.
```
mikelehen: Do you have a use case in mind where we would set this to false?

I can imagine wanting control over cache usage but I'm not sure if a simple boolean will be sufficient.  It might be more like "use the cache if it's not more than X hours stale." or something.  So if we don't *need* this functionality, I'm somewhat inclined to leave it out until we have a need and know exactly what it should look like.

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -75,14 +84,32 @@ export class MetricCatalog {
...
     region: Region,
     metric: string | Metric,
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    includeTimeseries = true
+    includeTimeseries = true,
+    useSnapshot = true
   ): Promise<MetricData> {
+    if (this.snapshot !== undefined && useSnapshot) {
+      metric = metric instanceof Metric ? metric : this.getMetric(metric);
+      const metricData = this.snapshot["data"][region.regionId][metric.id];
```
mikelehen: 1. Is there a reason we're doing `["data"]` instead of `.data`?  I would always use the latter for predefined values in an object.
2. I think this will crash if we don't have any data cached for the region.  We probably need to use the optional chaining operator like: `this.snapshot.data[region.regionId]?.[metric.id]` or similar.

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -75,14 +84,32 @@ export class MetricCatalog {
...
+      const timeseriesJSON = metricData["timeseries"]["points"];
```
mikelehen: Again, I'd use `.timeseries.points`.  We also might need some if-checks or optional chaining to guard against timeseries being missing, etc.

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -75,14 +84,32 @@ export class MetricCatalog {
...
+      if (metricData !== undefined) {
+        const timeseries =
+          timeseriesJSON && includeTimeseries
+            ? Timeseries.fromJSON(timeseriesJSON)
+            : undefined;
+        return new MetricData(
+          metric,
+          region,
+          metricData["currentValue"],
```
mikelehen: Again, `metricData.currentValue`

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -92,14 +119,28 @@ export class MetricCatalog {
...
     region: Region,
     metrics: string[] | Metric[],
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    includeTimeseries = true
+    includeTimeseries = true,
+    useSnapshot = true
   ): Promise<MultiMetricDataStore> {
+    if (this.snapshot !== undefined && useSnapshot) {
+      const dataMap: MetricToDataMap<unknown> = {};
+      metrics.forEach(async (metric) => {
```
mikelehen: `async (metric) => { ... }` is an async function which means it will return a `Promise`.  A `Promise` represents an async operation that may either successfully complete (and the promise is "resolved") or fail with an error (and the Promise is "rejected") in the future.  

As written, this forEach loops is going to call the async function once for each metric, which will create a bunch of `Promise` objects that we then completely ignore, since `forEach()` doesn't do anything with return values.  That means that:

1. We might return from `fetchDataForMetric()` before those async functions are done and therefore we may not populate `dataMap` correctly before this function returns.
2. We will ignore any errors that come from those async functions.

The way to fix this is to replace the `forEach` loop with something like:

`await Promise.all(metrics.map(async (metric) => { ... }));`

This takes every metric in `metrics`, uses `map()` to call our async function on it, and returns an array of all the resulting `Promise` objects.  And then `Promise.all()` merges them into a single `Promise` that will be resolved if all of the Promises complete successfully or by rejected if any one errors.  And then `await` will correctly "wait" for that promise.

Note that `await` doesn't actually synchronously wait.  There is no way to synchronously wait in JavaScript (it's 100% non-blocking).  So `await` is just syntactic sugar to help your function Propagate the `Promise` as the eventual result of the function that you're using it from.  If that doesn't make sense, feel free to read up more an async/await, but you can conceptually think of it as "waiting" for the promise to complete. 😁

mikelehen: As an aside, I'm a bit surprised neither https://typescript-eslint.io/rules/no-misused-promises/ nor https://typescript-eslint.io/rules/no-floating-promises/ is catching this.  I wonder if we could improve our eslint config somehow...  I'll try to dig into this later.

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -109,14 +150,27 @@ export class MetricCatalog {
...
     regions: Region[],
     metrics: string[] | Metric[],
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    includeTimeseries = true
+    includeTimeseries = true,
+    useSnapshot = true
   ): Promise<MultiRegionMultiMetricDataStore> {
+    if (this.snapshot !== undefined && useSnapshot) {
+      const dataMap: MultiRegionMultiMetricDataMap<unknown> = {};
+      regions.forEach(async (region) => {
```
mikelehen: Similar feedback here.

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -92,14 +119,28 @@ export class MetricCatalog {
...
    */
   async fetchDataForMetrics(
     region: Region,
     metrics: string[] | Metric[],
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    includeTimeseries = true
+    includeTimeseries = true,
+    useSnapshot = true
   ): Promise<MultiMetricDataStore> {
+    if (this.snapshot !== undefined && useSnapshot) {
```
mikelehen: Sorry, this is confusing because I haven't got these implemented yet, but most likely I'll implement `fetchData()` and `fetchDataForMetrics()` so that they just call `fetchDataForMetricsAndRegions()` so if you want you can just put the snapshot logic in there.

Note that the code will probably be a bit more complicated (it has to deal with multiple metrics and regions), but I'm trying to organize the code so that by default we ask the data providers for "big" chunks of data at a time (multiple metrics and regions if applicable) rather than "small" chunks (1 metric or region) since I think that'll be more efficient.

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -5,6 +5,7 @@ import first from "lodash/first";
 import last from "lodash/last";
 import maxBy from "lodash/maxBy";
 import minBy from "lodash/minBy";
+import { TimeseriesPointJSON } from "../data/MultiRegionMultiMetricDataStore";
```
mikelehen: I'd like to keep `Timeseries` as its own standalone data type that doesn't depend on any metric-related code.  So metrics code can depend on timeseries code but not vice versa.  Can we move `TimeseriesPointJSON` into this file?

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -268,6 +269,36 @@ export class Timeseries<T = unknown> {
...
+
+  /**
+   * Construct a Timeseries instance from JSON timeseries points.
+   *
+   * @param jsonPoints Array of date and value JSON objects to construct timeseries from.
+   * @returns
+   */
+  static fromJSON<T>(jsonPoints: TimeseriesPointJSON[]): Timeseries<T> {
+    const timeseriesPoints: TimeseriesPoint<T>[] = jsonPoints.map((point) => {
+      return { date: new Date(point.date), value: point.value as T };
```
mikelehen: This `as T` is an unsafe cast.  You could do `fromJSON<number>(...)` but the value in the JSON is actually a string and your code will compile fine.  So I think we should make this function *not* generic and have it return `Timeseries<unknown>`.  At some point up the chain the consumer is going to have to call `assertFiniteNumbers()` or something if they want to ensure it is numeric data and deal with it as `Timeseries<number>`.

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -268,6 +269,36 @@ export class Timeseries<T = unknown> {
     // Dates are guaranteed not to have a time component so we just return the ISO date.
     return date.toISOString().split("T")[0];
   }
+
+  /**
+   * Construct a Timeseries instance from JSON timeseries points.
+   *
+   * @param jsonPoints Array of date and value JSON objects to construct timeseries from.
+   * @returns
```
mikelehen: Either remove `@returns` or preferably fill it in (i.e. `@returns The constructed timeseries.` or similar)

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -268,6 +269,36 @@ export class Timeseries<T = unknown> {
...
+   */
+  static fromJSON<T>(jsonPoints: TimeseriesPointJSON[]): Timeseries<T> {
+    const timeseriesPoints: TimeseriesPoint<T>[] = jsonPoints.map((point) => {
+      return { date: new Date(point.date), value: point.value as T };
+    });
+    return new Timeseries<T>(timeseriesPoints);
+  }
+
+  /**
+   * Convert Timeseries instance into serializable JSON format.
```
mikelehen: nit: I'd drop "serializable" or change to "serialized"

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -268,6 +269,36 @@ export class Timeseries<T = unknown> {
     // Dates are guaranteed not to have a time component so we just return the ISO date.
     return date.toISOString().split("T")[0];
   }
+
+  /**
+   * Construct a Timeseries instance from JSON timeseries points.
+   *
+   * @param jsonPoints Array of date and value JSON objects to construct timeseries from.
```
mikelehen: nit: Somebody reading the code or the docs can look at `TimeseriesPointJSON` to see what exactly the array should look like.  So I think "date and value JSON objects" is not super helpful / necessary.  I would instead say "Serialized timeseries points from which to construct a Timeseries" or something.

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -268,6 +269,36 @@ export class Timeseries<T = unknown> {
...
+    const timeseriesPoints: TimeseriesPoint<T>[] = jsonPoints.map((point) => {
+      return { date: new Date(point.date), value: point.value as T };
+    });
+    return new Timeseries<T>(timeseriesPoints);
+  }
+
+  /**
+   * Convert Timeseries instance into serializable JSON format.
+   *
+   * @returns Array of date and value JSON objects containing data from this.points.
```
mikelehen: Again, I'd probably say "Timeseries points serialized as JSON" or something without mentioning the date/value members.

packages/metrics/src/data/MultiRegionMultiMetricDataStore.test.ts
```diff
@@ -0,0 +1,23 @@
...
+      },
+    });
+    const provider = new StaticValueDataProvider();
+    const dataStore = await provider.fetchData(
+      [states.findByRegionIdStrict("12")],
+      [metric],
+      true
+    );
+    dataStore.regionData(states.findByRegionIdStrict("12")).metricData(metric);
+    dataStore.createSnapshot();
```
mikelehen: Thanks for starting the tests!  Can we replace this smoke test with a test that actually validates the resulting snapshot?

FWIW- Smoke tests (that just minimally execute some code to make sure it doesn't crash) aren't terrible but I wouldn't call them a best practice.  I sometimes write them as I'm prototyping as a quick sanity-check but then go back and replace or amend them to do more rigorous testing.  

So I wouldn't necessarily follow my example. :-)

packages/metrics/src/data/MultiRegionMultiMetricDataStore.test.ts
```diff
@@ -0,0 +1,23 @@
...
+      dataReference: {
+        providerId: "static-value",
+        value: 42,
+      },
+    });
+    const provider = new StaticValueDataProvider();
+    const dataStore = await provider.fetchData(
+      [states.findByRegionIdStrict("12")],
+      [metric],
+      true
```
mikelehen: nit: I'm a big fan of commenting literals like this with a comment like `/*useSnapshot=*/true`

packages/metrics/src/data/MultiRegionMultiMetricDataStore.test.ts
```diff
@@ -0,0 +1,23 @@
...
+    );
+    dataStore.regionData(states.findByRegionIdStrict("12")).metricData(metric);
```
mikelehen: nit: maybe extract `states.findByRegionIdStrict("12")` into a `region` variable at the top of the test?

packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
```diff
@@ -1,8 +1,34 @@
...
 import mapValues from "lodash/mapValues";
 import { MultiMetricDataStore } from "./MultiMetricDataStore";
 
+export interface SnapshotJSON {
+  metadata: { createdDate: string; latestDate: string | null };
+  data: RegionDataJSON;
+}
+
+export interface RegionDataJSON {
+  [regionFips: string]: metricDataJSON;
```
mikelehen: nit: regionFips => regionId

packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
```diff
@@ -1,8 +1,34 @@
...
+
+export interface TimeseriesPointJSON {
+  date: string;
+  value: unknown;
+}
+
+interface metricDataJSON {
+  [metricName: string]: {
+    currentValue: unknown;
+    timeseries: { points: TimeseriesPointJSON[] | null };
```
mikelehen: I'm not sure we need `timeseries` and `points` as nested entries.  We could just have `timeseriesPoints`.

In the interest of brevity in the JSON I'd also probably just make the field optional, i.e. `timeseriesPoints?: TimeseriesPointJSON[]`.  That way we don't have to generate a bunch of empty `timeseries` fields.

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -109,14 +150,27 @@ export class MetricCatalog {
...
   async fetchDataForMetricsAndRegions(
     regions: Region[],
     metrics: string[] | Metric[],
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    includeTimeseries = true
+    includeTimeseries = true,
+    useSnapshot = true
   ): Promise<MultiRegionMultiMetricDataStore> {
+    if (this.snapshot !== undefined && useSnapshot) {
+      const dataMap: MultiRegionMultiMetricDataMap<unknown> = {};
```
mikelehen: Given that `createSnapshot()` lives on `MultiRegionMultiMetricDataStore` I would probably create a `MultiRegionMultiMetricDataStore.fromSnapshot()` static method that constructs a `MultiRegionMultiMetricDataStore` from a snapshot for symmetry. And then we would just call that from here.

packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
```diff
@@ -33,4 +57,39 @@ export class MultiRegionMultiMetricDataStore<T = unknown> {
...
+  /**
+   * Preps the contents of this.data to be persisted to JSON file.
+   */
+  createSnapshot(): SnapshotJSON {
+    const metricDataStores = Object.values(
+      this.regionToMultiMetricDataStoreMap
+    );
+    const records: RegionDataJSON = {};
+    const maxDates: (Date | undefined)[] = [];
+    for (let i = 0; i < metricDataStores.length; i++) {
```
mikelehen: nit: You could make this code a bit more concise with `for (const dataStore of Object.values(this.regionToMultiMetricDataStoreMap)) { ...`  or `Object.values(...).forEach(...` to match the code you have below.

packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
```diff
@@ -33,4 +57,39 @@ export class MultiRegionMultiMetricDataStore<T = unknown> {
...
+      const dataStore = metricDataStores[i];
+      const regionFips = dataStore.region.regionId;
```
mikelehen: nit: regionFips => regionId.

For context: We may be dealing with regions that are e.g. countries and this will be an ISO3 code or something.

packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
```diff
@@ -33,4 +57,39 @@ export class MultiRegionMultiMetricDataStore<T = unknown> {
...
+      const metricJsons: metricDataJSON = {};
+
+      Object.values(dataStore.metricToDataMap).forEach((metric) => {
```
mikelehen: nit: metric => metricData ?

packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
```diff
@@ -33,4 +57,39 @@ export class MultiRegionMultiMetricDataStore<T = unknown> {
...
+        if (metric !== undefined) {
+          maxDates.push(metric.timeseries.maxDate());
+          metricJsons[metric.metric.id] = {
```
mikelehen: nit: I'd rename `metric` to `metricData`

packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
```diff
@@ -33,4 +57,39 @@ export class MultiRegionMultiMetricDataStore<T = unknown> {
...
+            currentValue: metric.currentValue,
+            timeseries: { points: metric.timeseries.toJSON() } ?? null,
```
mikelehen: nit: We need to call `hasTimeseries` first I think.  `.timeseries` can throw if there's no timeseries available.

packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
```diff
@@ -33,4 +57,39 @@ export class MultiRegionMultiMetricDataStore<T = unknown> {
...
+    const records: RegionDataJSON = {};
+    const maxDates: (Date | undefined)[] = [];
+    for (let i = 0; i < metricDataStores.length; i++) {
+      const dataStore = metricDataStores[i];
+      const regionFips = dataStore.region.regionId;
+      const metricJsons: metricDataJSON = {};
+
+      Object.values(dataStore.metricToDataMap).forEach((metric) => {
+        if (metric !== undefined) {
+          maxDates.push(metric.timeseries.maxDate());
```
mikelehen: nit: We don't really need to build up an array of maxDates and then call max() at the end, we could just keep track of the current maxDate overall.

mikelehen (changes requested):
Overall this looks great, but I did leave lots of nitty feedback. 😁   Thanks / sorry!

packages/metrics/src/MetricCatalog/MetricCatalog.ts
mikelehen: Actually I think that makes sense.  We're going to be creating `MetricData<unknown>` instances, since we can't know at compile-time what the types are going to be.

packages/metrics/src/MetricCatalog/MetricCatalog.test.ts
smcclure17: Oh, yeah I agree--I think that makes more sense!

packages/metrics/src/MetricCatalog/MetricCatalog.ts
smcclure17: In the prototype in who-demo (https://github.com/covid-projections/who-demo/pull/171) I needed this (or something like it) when creating the `MultiRegionMultiMetricDataStore` that was persisted to create the snapshot because otherwise, the data-store would read from the existing snapshot instead of "refreshing" the data by pulling from the data providers. 

But, I agree that we'll probably want more control than this, so I'll remove it until we have a better picture of what we'll need!

packages/metrics/src/MetricCatalog/MetricCatalog.ts
smcclure17: 1. Ah, no real reason--I think that's how it's generally done in Python, so just an old habit. 
2. Good flag, thanks! I'm refactoring this code into a `MultiRegionMultiMetricDataStore.fromSnapshot()` method, and will be cognizant of this there

packages/metrics/src/MetricCatalog/MetricCatalog.ts
smcclure17: Ah I see--that makes sense (though I should still read more on async stuff 😅). Thanks for this very thorough explanation!! 🙏 🙏

packages/metrics/src/MetricCatalog/MetricCatalog.test.ts
smcclure17: yeah, that's a good idea! In 3089b1a6a72387ef8ab7fb9274ce64d726a016f5 I've updated this test to read from a metric without a value (it still has the `static-value` provider, but no `value`).

packages/metrics/src/MetricCatalog/MetricCatalog.test.ts
```diff
@@ -1,37 +1,63 @@
...
+  };
+  const metricCatalog = new MetricCatalog([metricDefinition], dataProviders, {
+    snapshot: snapshot,
+  });
+  const multiMetricDataStore = new MultiMetricDataStore(region, {
+    [metric.id]: new MetricData(metric, region, 42),
+  });
+  const multiRegionMultiMetricDataStore = new MultiRegionMultiMetricDataStore({
+    [region.regionId]: multiMetricDataStore,
+  });
```
mikelehen, line 45: Given `multiMetricDataStore` and `multiRegionMultiMetricDataStore` are only used in one test, I'd move them inside that test.

packages/metrics/src/MetricCatalog/MetricCatalog.test.ts
```diff
@@ -1,37 +1,63 @@
...
+
+  test("getMetric() returns correct metric, or throws error for unsupported metric.", () => {
+    expect(metricCatalog.getMetric(metric.id)).toStrictEqual(metric);
+    expect(() => {
+      metricCatalog.getMetric("nothing");
+    }).toThrow(`No metric found with id nothing`);
+  });
```
mikelehen, line 52: I'd normally put newlines between tests.

packages/metrics/src/MetricCatalog/MetricCatalog.test.ts
```diff
@@ -1,37 +1,63 @@
...
+  test("fetchDataForMetricsAndRegions() correctly reads from a snapshot.", () => {
+    return expect(
+      // TODO: change to fetchData() when it gets implemented and rip out un-needed code above.
+      metricCatalog.fetchDataForMetricsAndRegions(
+        [region],
+        [metric],
+        /*includeTimeseries*/ false
+      )
+    ).resolves.toEqual(multiRegionMultiMetricDataStore);
```
mikelehen: I think this syntax probably works fine, but I'd have written it using `async` / `await` like the other tests you had.

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -114,9 +120,19 @@ export class MetricCatalog {
   async fetchDataForMetricsAndRegions(
     regions: Region[],
     metrics: string[] | Metric[],
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     includeTimeseries = true
   ): Promise<MultiRegionMultiMetricDataStore> {
+    if (this.snapshot !== undefined) {
+      metrics = metrics.map((metric) => {
+        return metric instanceof Metric ? metric : this.getMetric(metric);
+      });
```
mikelehen: nit: For single-line lambda functions like this, I would probably drop the brackets and the `return`, ie:

```
      metrics = metrics.map((metric) => metric instanceof Metric ? metric : this.getMetric(metric));
```

packages/metrics/src/MetricCatalog/MetricCatalogOptions.ts
```diff
@@ -33,4 +34,10 @@ export interface MetricCatalogOptions {
    * ```
    */
   metricLevelSets?: MetricLevelSet[];
+
+  /**
+   * @param snapshot JSON cache file to read metrics from instead of
```
mikelehen: Because this is a property on an interface rather than a function parameter, you don't need/want `@param snapshot`.  You can just do something like:

```
/** JSON cache file to read metrics from instead of using DataProviders. */
snapshot?: SnapshotJSON;
```

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -6,6 +6,12 @@ import last from "lodash/last";
 import maxBy from "lodash/maxBy";
 import minBy from "lodash/minBy";
 
+/**A single, serialized point in a timeseries containing a date-string and a value. */
```
mikelehen: nit: space after `/**`

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -268,6 +274,38 @@ export class Timeseries<T = unknown> {
...
+   * Construct a Timeseries instance from JSON timeseries points.
+   *
+   * @param jsonPoints Serialized timeseries points from which to construct a Timeseries.
+   * @returns The constructed timeseries.
+   */
+  static fromJSON(jsonPoints: TimeseriesPointJSON[]): Timeseries<unknown> {
+    const timeseriesPoints: TimeseriesPoint<unknown>[] = jsonPoints.map(
+      (point) => {
+        return { date: new Date(point.date), value: point.value as unknown };
+      }
```
mikelehen: nit: Again, I'd normally do this without the brackets / `return`

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -268,6 +274,38 @@ export class Timeseries<T = unknown> {
...
+
+  /**
+   * Convert Timeseries instance into serialized JSON format.
+   *
+   * @returns Timeseries points serialized as JSON.
+   */
+  toJSON(): TimeseriesPointJSON[] {
+    const timeseriesPointJSONs: TimeseriesPointJSON[] = this.points.map(
+      (point) => {
+        return {
```
mikelehen: This can be done without brackets and `return` too, but note that since it's returning an object (`{ date: ..., value: ... }`) you will need an extra set of parens to avoid ambiguity and have TypeScript try to interpret the `{ ... }` as a function body.  So something like: `(point) => ({ date: ..., value: ... })`

packages/metrics/src/data/MetricData.ts
```diff
@@ -70,8 +70,9 @@ export class MetricData<T = unknown> {
    */
   assertFiniteNumbers(): MetricData<number> {
     assert(
-      this.currentValue === null ||
-        (typeof this.currentValue === "number" && isFinite(this.currentValue)),
+      this.currentValue !== null &&
```
mikelehen: So the existing code was intentionally allowing `currentValue` to be null (i.e. missing).  My thought was that `currentValue` is always optional in the case the metric value is missing / unavailable.  E.g. in CAN we would perhaps want to call `removeNils().assertFiniteNumbers()` on say the "casesPerDay" metric.  This would remove null points from the timeseries and ensure that any remaining data is numeric.  This should always be safe to do unless there's a bug in the data pipeline.  But we do still want to allow for missing data (due to lack of data coverage or blacking metrics, etc.).  So `currentValue` could still be null.

That said, I know this code is a bit inconsistent in that it's allowing `null` in `currentValue` but not in the timeseries. 😦 Maybe there's a better way to deal with this, but not sure.

packages/metrics/src/data/MultiRegionMultiMetricDataStore.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  const region = states.findByRegionIdStrict("12");
+  const metric = new Metric({
+    id: "cases_per_100k",
+    dataReference: {
+      providerId: "static-value",
+      value: 42,
+    },
+  });
+  const provider = new StaticValueDataProvider();
+  const expectedSnapshot: SnapshotJSON = {
```
mikelehen: Hrm.  This looks a bit funky.  In general I would only put "common" code that's shared across tests inside of the top-level `describe()` block.  This should generally be for test constants, etc.  You could even move them outside of the `describe()` block to be top-level constants in the file.  I don't know if there's a best practice, but that's what I usually do.

`expectedSnapshot` sounds like it's the snapshot that a particular test is expecting.  I would generally expect different tests to test different things and have different expectations though.  I guess this is the snapshot that you expect to get back for `metric` and `region`.

If it were me I'd actually rename it to `testSnapshot` and maybe also rename the other variables to match (`testRegion`, `testMetric`, etc.) to indicate that they are all test data used by (multiple) tests.  And then finally maybe add a comment here.

```
// The snapshot JSON that corresponds to the data for `testRegion` and `testMetric`.

File: packages/metrics/src/data/MultiRegionMultiMetricDataStore.test.ts [hunk 40]
```diff
@@ -0,0 +1,154 @@
...
+      MultiRegionMultiMetricDataStore.fromSnapshot(
+        expectedSnapshot,
+        [region],
+        [metric],
+        /*includeTimeseries*/ true
+      )
+    ).toEqual(dataStore);
+
+    expectedSnapshot.data[region.regionId][metric.id].timeseriesPoints =
+      undefined;
```
mikelehen: Oh, hrm.  Since `expectedSnapshot` is shared between multiple tests, mutating it like this could be very confusing.  It means that the value of `expectedSnapshot` depends on what tests have already run.  So if you only run a subset of tests or you run tests in a different order, that could cause your test to fail.  One way to fix this would be to create a copy in this test and then modify that.  Since this is a nested object, you'll need a "deep" copy, which you could do with lodash: https://www.geeksforgeeks.org/lodash-_-clonedeep-method/

Another option would be to change `testSnapshot` into a function that generates test snapshots based on parameters, e.g. `makeTestSnapshot(includeTimeseries: boolean): SnapshotJSON` and then you can call it from different tests with different options to get the snapshot you need.

packages/metrics/src/data/MultiRegionMultiMetricDataStore.test.ts
```diff
@@ -0,0 +1,154 @@
...
+      [region],
+      [metric],
+      /*includeTimeseries=*/ false
+    );
+    expect(
+      MultiRegionMultiMetricDataStore.fromSnapshot(
+        expectedSnapshot,
+        [region],
+        [metric],
+        /*includeTimeseries*/ false
```
mikelehen: super nit: When commenting boolean literals, I always include an `=` for clarity.  Else sometimes it's kind of ambiguous.  Does `/*disableLogs*/false` imply that you are specifying `false` to disable logs?  Or does it imply that you are setting `disableLogs` to false, and so longs are enabled?  `/*disableLogs=*/false` is perfectly clear, at least to me. 😁

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -114,9 +120,19 @@ export class MetricCatalog {
...
+      return MultiRegionMultiMetricDataStore.fromSnapshot(
+        this.snapshot,
+        regions,
+        metrics,
+        includeTimeseries
```
mikelehen, line 133: I'm tempted to suggest that we ingest the snapshot once up front in the `MetricCatalog` constructor and turn it into a single `MultiRegionMetricDataStore` that represents the entire snapshot rather than "parsing" the snapshot every time somebody does a fetch.  

But to do that, I think we'd need to have access to all of the `regions` so that we can look up the region ids found in the snapshot.  So that would mean you'd need to pass `regions` into the MetricCatalog constructor or something.

Anyway, I think this is fine for now and we can revisit later.

packages/metrics/src/data/MultiRegionMultiMetricDataStore.test.ts
mikelehen: Oh, also, given we're passing `false` here, why do we also need to mutate `expectedSnapshot` above?  Maybe we don't?  Not sure why we have to do both.

packages/metrics/src/data/MultiRegionMultiMetricDataStore.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  test("assertFiniteNumbers() accepts numeric values", () => {
+    const verifyValueIsValid = (value: number) => {
+      expect(() => {
+        const dataStore = new MultiRegionMultiMetricDataStore({
+          region: new MultiMetricDataStore(region, {
+            metric: new MetricData(metric, region, value),
+          }),
+        });
+        dataStore.assertFiniteNumbers();
+      }).toBeDefined();
```
mikelehen: I'm not sure if this `expect(() => { ... }).toBeDefined()` wrapper is necessary or even is going to work correctly.  My guess is it's not actually executing your code right now.  It's basically asserting that `() => { ... } !== undefined` but it's not actually calling your lambda function.

If you wanted to keep the same format as the above test, I think you could possible do `expect(() => { ... }).not.toThrowError()` and then it would actually execute your function.  But since tests will automatically fail if an error is thrown, you could just remove the `expect(() => { ... }).toBeDefined()` wrapper entirely. 🤷

packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
```diff
@@ -33,4 +54,91 @@ export class MultiRegionMultiMetricDataStore<T = unknown> {
     );
     return new MultiRegionMultiMetricDataStore(regionToMultiMetricDataStoreMap);
   }
+
+  /**
+   * Preps the contents of this.data to be persisted to JSON file.
```
mikelehen: nit: I think `this.data` got renamed (though I might rename it back; it got kinda' annoyingly long).  That said, I would actually try to phrase the comment in terms of the overall `MultiRegionMetricDataStore` object and say something like, "Persists all data into a JSON-compatible snapshot object." or similar.

packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
```diff
@@ -33,4 +54,91 @@ export class MultiRegionMultiMetricDataStore<T = unknown> {
...
+   */
+  toSnapshot(): SnapshotJSON {
+    const metricDataStores = Object.values(
+      this.regionToMultiMetricDataStoreMap
+    );
+    const records: RegionDataJSON = {};
+    const minimumDate = new Date(-8640000000000000);
+    let maxDate = minimumDate; // This is hacky. Sets the "starting" as the lowest possible date.
```
mikelehen: An alternative would be to use `undefined` as a placeholder.  So `let maxDate: Date | undefined;` and then have a special case below like `maxDate = !maxDate || timeseries.maxDate() > maxDate ? timeseries.maxDate() : maxDate;`

mikelehen (approved):
Looks good with some nits and one question / concern about `MetricData.currentValue` and `null` values.

packages/metrics/src/MetricCatalog/MetricCatalog.ts
smcclure17: Yeah, that was my initial thought, too, but wasn't sure how we could deal with "looking up" every metric and region in the snapshot as they're not (at least at the moment) tied one-to-one with `regionDB`s or `MetricCatalog`s. 

I agree that that's probably the ideal behavior, but if we're okay with revisiting this I might leave it as-is for now.

packages/metrics/src/data/MetricData.ts
smcclure17: Ahh, okay, that does make sense! I've reverted this change and added a quick comment for context/clarity: `Null values are treated as finite and will not throw an error.`

packages/metrics/src/data/MultiRegionMultiMetricDataStore.test.ts
smcclure17: Oh, right, yeah we don't need to modify that at all 🤦

packages/metrics/src/data/MultiRegionMultiMetricDataStore.test.ts
smcclure17: Hmm, yeah it looks like you're right. I put a `console.log()` statement in the test to check and it didn't seem to get executed :( 

Using `.not.toThrowError()` seems to execute it correctly and I like how it reads, so I've switched it to this 🙏

packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
smcclure17: ahhh, that's much neater! I had trouble finding a setup that avoided comparing the initial `undefined`/`null` value to the dates. Thanks! 🙏

PR #43, opened 2022-08-09
ui components: export themeConfig object independently

chasulin:


packages/ui-components/src/index.ts
```diff
@@ -1,7 +1,7 @@
 import { Theme } from "@mui/material/styles";
 
 /** Theme variables */
-export * from "./styles"; // styled, theme
+export { themeConfig } from "./styles";
```
chasulin, line 4: since `theme` and `styled` are only for use within the ui-components package, we don't need to expose them/export them

pnavarrc (approved):
Looks good!

PR #45, opened 2022-08-09
Minor Timeseries tweaks and prep for publishing.

mikelehen:
* Added a Timeseries.fromDateRange() that's convenient for creating test timeseries, etc.
* Changed several simple methods on Timeseries into getter properties instead (last(), minValue(), etc.).
* Make Timeseries use time-utils package for formatting ISO dates.
* Changed the version to 0.0.1 in preparation for publishing.
* Changed the default package version (in plopfile) to 0.0.1.
* Also added a couple helpers to the number-format package that were handy.

### DIFF
--- a/packages/number-format/src/index.ts
+++ b/packages/number-format/src/index.ts
@@ -145,3 +145,24 @@ function getOptionsFromPlacesOrOptions(
       }
     : { ...placesOrOptions, ...formattingStyleForOptions };
 }
+
+/**
+ * Type guard that checks if the provided value is a number.
+ *
+ * @param value Value to check.
+ * @returns True if the value is a number.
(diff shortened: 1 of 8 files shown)


packages/number-format/src/index.ts
```diff
@@ -145,3 +145,24 @@ function getOptionsFromPlacesOrOptions(
       }
     : { ...placesOrOptions, ...formattingStyleForOptions };
 }
+
```
mikelehen, line 148: The below added functions are similar to lodash functions, but they function as a type guard as well, so they work better with TypeScript.

pnavarrc (approved):
Looks good!

PR #46, opened 2022-08-09
Add legend threshold vertical

### LINKED ISSUE #60: Legend threshold (vertical)

faihegberg:
This PR adds legend threshold vertical to Act Now Packages.

**Note:** Labels are not implemented in this PR, as per note [here](https://www.dropbox.com/scl/fi/j8laxj6xeyoz12ambmldf/Act-Now-as-a-Platform-Eng.-Vision-and-Q3-Roadmap.paper?dl=0&rlkey=s0mf7m662swlj4zo8dlesd6o3#:h2=LegendThreshold-(vertical)).

close #60

### DIFF
--- a/packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
+++ b/packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
@@ -14,10 +14,7 @@ export type LegendThresholdProps<T> =
  * `LegendThreshold` represents a scale with thresholds that separate
  * a set of levels. By default, the labels between each level are shown.
  */
-const LegendThreshold = <T,>(
-  props: LegendThresholdProps<T> &
-    Omit<React.SVGProps<SVGSVGElement>, keyof LegendThresholdHorizontalProps<T>>
-) => {
+const LegendThreshold = <T,>(props: LegendThresholdProps<T>) => {
   return props.orientation === "horizontal" ? (
     <LegendThresholdHorizontal {...props} />
   ) : (
--- a/packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
+++ b/packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
@@ -1,67 +1,24 @@
 import React from "react";
-import LegendThresholdHorizontal from "./LegendThresholdHorizontal";
-import LegendThresholdVertical from "./LegendThresholdVertical";
+import LegendThresholdHorizontal, {
+  LegendThresholdHorizontalProps,
+} from "./LegendThresholdHorizontal";
+import LegendThresholdVertical, {
+  LegendThresholdVerticalProps,
+} from "./LegendThresholdVertical";
 
-export interface LegendThresholdProps<T> {
-  /** Orientation of the bars */
-  orientation?: "horizontal" | "vertical";
-  /** Height of the component, including the colored bars and labels. */
-  height?: number;
-  /**
-   * Height of the colored bars. When not adding the bars, make sure that
-   * `barHeight` is set to the same value as `height`.
-   */
-  barHeight?: number;
-  /** Width of the component */
-  width?: number;
-  /** Border radius of the colored bars */
-  borderRadius?: number;
-  /** List of items representing the labels */
-  items: T[];
-  /** Function that returns the color of each level */
-  getItemColor: (item: T, itemIndex: number) => string;
-  /**
-   * Note that only the labels between two levels are rendered. The last
-   * end label won't be shown.
-   */
-  getItemEndLabel: (item: T, itemIndex: number) => string;
-  /**
-   * Whether to show the labels or not (true by default). Make sure to set
-   * `barHeight` to `height` when not including the labels.
-   */
-  showLabels?: boolean;
-}
+export type LegendThresholdProps<T> =
+  | LegendThresholdHorizontalProps<T>
+  | LegendThresholdVerticalProps<T>;
 
 /**
  * `LegendThreshold` represents a scale with thresholds that separate
  * a set of levels. By default, the labels between each level are shown.
  */
-const LegendThreshold = <T,>({
-  orientation = "horizontal",
-  height,
-  barHeight,
-  width,
-  borderRadius = 4,
-  items,
-  getItemColor,
-  getItemEndLabel,
-  showLabels = true,
-  ...otherSvgProps
-}: LegendThresholdProps<T> &
-  Omit<React.SVGProps<SVGSVGElement>, keyof LegendThresholdProps<T>>) => {
-  const props = {
-    orientation,
-    height,
-    barHeight,
-    width,
-    borderRadius,
-    items,
-    getItemColor,
-    getItemEndLabel,
-    showLabels,
-    ...otherSvgProps,
-  };
-  return orientation === "horizontal" ? (
+const LegendThreshold = <T,>(
+  props: LegendThresholdProps<T> &
+    Omit<React.SVGProps<SVGSVGElement>, keyof LegendThresholdHorizontalProps<T>>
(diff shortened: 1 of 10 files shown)


packages/ui-components/src/components/LegendThreshold/LegendThresholdHorizontal.tsx
```diff
@@ -0,0 +1,75 @@
+import React from "react";
+import { Group } from "@visx/group";
+import { scaleBand } from "@visx/scale";
+import uniqueId from "lodash/uniqueId";
+import { TickLabel, TickMark } from "./LegendThreshold.style";
+import { LegendThresholdProps } from ".";
+
+/**
+ * `LegendThreshold` represents a scale with thresholds that separate
```
pnavarrc: Consider updating the docstring (LegendThresholdHorizontal)

packages/ui-components/src/components/LegendThreshold/LegendThresholdHorizontal.tsx
```diff
@@ -0,0 +1,75 @@
...
+  height = 40,
+  barHeight = 20,
+  width = 300,
+  borderRadius,
+  items,
+  getItemColor,
+  getItemEndLabel,
+  showLabels,
+  ...otherSvgProps
+}: LegendThresholdProps<T> &
```
pnavarrc: I think we can simplify a little here. The vertical and horizontal versions don't have exactly the same props, so maybe we can do something like this

```tsx
// interfaces.ts
export interface CommonLegendThresholdProps {
  width: number;
  height: number;
  // etc...
}

// LegendThresholdHorizontal.tsx
interface LegendThresholdHorizontalProps extends CommonLegendThresholdProps {
  orientation: 'horizontal';
  barHeight?: number;
  showLabels?: boolean;
  // ...etc
}

// LegendThresholdVertical.tsx
interface LegendThresholdVerticalProps extend CommonLegendThresholdProps {
  orientation: 'vertical';
  // ...etc
}
```

and then

```tsx
type LegendThresholdProps = LegendThresholdHorizontalProps | LegendThresholdVerticalProps;

const LegendThreshold: React.FC<LegendThresholdProps> = (props) => {
  return props.orientation === 'horizontal
    ? <LegendThresholdHorizontal {...props} />
    : <LegendThresholdVertical {...props} />;
}
```

pnavarrc (changes requested):
I think it will be best if we create different props for vertical and horizontal (hopefully they will converge eventually), since `barHeight` and `showLabels` are not relevant for the vertical variant. I added a suggestion on how to go about it

chasulin (commented):
i wonder if we should use the material-ui-ish way of setting props using booleans where possible, rather than 'vertical' vs 'horizontal'. [in LegendCategorical, i just defaulted to horizontal (an optional horizontal prop that defaults to true)](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/components/LegendCategorical/LegendCategorical.tsx#L17) -- i can definitely change legend categorical to match the types you use here. totally up for either way, but we should try to be consistent with conventions like this

and per the point on omitting labels for now, i feel like we can add them with the assumption that they should have some sort of max-width and wrapping behavior. the specifics of the width can change in the future if need be, but its probably a safe assumption to make and start on?

packages/ui-components/src/components/LegendThreshold/LegendThresholdVertical.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { scaleBand } from "@visx/scale";
+import uniqueId from "lodash/uniqueId";
+import { CommonLegendThresholdProps } from ".";
+
+export type LegendThresholdVerticalProps<T> = CommonLegendThresholdProps<T>;
+
+/**
+ * `LegendThresholdVertical` represents a scale with thresholds that separate
+ * a set of levels. By default, the labels between each level are shown.
+ */
```
faihegberg, line 18: Keeping this comment here as labels will be implemented in a follow-up PR.

packages/ui-components/src/components/LegendThreshold/LegendThresholdVertical.tsx
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * Whether to show the labels or not (true by default). Make sure to set
+   * `barHeight` to `height` when not including the labels.
+   */
+  showLabels?: boolean;
+  /**
+   * Note that only the labels between two levels are rendered. The last
+   * end label won't be shown.
+   */
+  getItemEndLabel?: (item: T, itemIndex: number) => string;
```
faihegberg: This will be adjusted as a follow-up PR.

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
```diff
@@ -1,100 +1,27 @@
...
-              )}
-            </Group>
-          );
-        })}
-      </Group>
-    </svg>
+const LegendThreshold = <T,>(
+  props: LegendThresholdProps<T> &
+    Omit<React.SVGProps<SVGSVGElement>, keyof LegendThresholdHorizontalProps<T>>
+) => {
```
pnavarrc: Why are we omitting the props of `LegendThrehsoldHorizontal`? I think it should just be 

```tsx
const LegendThreshold = <T,>(props: LegendThresholdProps<T>) => {
  /* ...*/
}
```

packages/ui-components/src/components/LegendThreshold/LegendThresholdVertical.tsx
```diff
@@ -0,0 +1,73 @@
...
+import uniqueId from "lodash/uniqueId";
+import { CommonLegendThresholdProps } from "./interfaces";
+
+export interface LegendThresholdVerticalProps<T>
+  extends CommonLegendThresholdProps<T> {
+  /**
+   * Note that only the labels between two levels are rendered. The last
+   * end label won't be shown.
+   */
+  getItemEndLabel?: (item: T, itemIndex: number) => string;
```
pnavarrc: Why do we have end labels? For the vertical legend, the props will likely be different (each level will have an interval, see the mocks)

packages/ui-components/src/components/LegendThreshold/interfaces.ts
```diff
@@ -0,0 +1,19 @@
+export interface CommonLegendThresholdProps<T> {
+  /** Orientation of the bars */
+  orientation?: "horizontal" | "vertical";
```
pnavarrc: This is not a common prop (the legend threshold vertical should only receive "vertical")

packages/ui-components/src/components/LegendThreshold/LegendThresholdHorizontal.tsx
```diff
@@ -0,0 +1,92 @@
...
+   * Height of the colored bars. When not adding the bars, make sure that
+   * `barHeight` is set to the same value as `height`.
+   */
+  barHeight?: number;
+  /**
+   * Note that only the labels between two levels are rendered. The last
+   * end label won't be shown.
+   */
+  getItemEndLabel?: (item: T, itemIndex: number) => string;
+}
```
pnavarrc, line 22: Consider adding `orientation="horizontal"` here and the equivalent for the vertical

pnavarrc (changes requested):
Added some comments, types need to be adjusted

packages/ui-components/src/components/LegendThreshold/LegendThresholdVertical.tsx
faihegberg: Adjusting to `getItemLabel` and `getItemSublabel`.

pnavarrc (approved):
Looks good!

PR #47, opened 2022-08-10
ui-components: attempt to fix markdown bug

chasulin:
the markdown component seems to be broken (something having to do with next). error:

Screen Shot 2022-08-10 at 11 54 56 AM

fix attempt: import ReactMarkdown dynamically

### DIFF
--- a/packages/ui-components/package.json
+++ b/packages/ui-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@actnowcoalition/ui-components",
-  "version": "1.5.1",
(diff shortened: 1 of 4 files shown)


mikelehen (approved):
Commenting out looks fine, but I'll be curious to hear how / why it broke.

mikelehen (changes requested):
Actually, I'm not seeing what's broken.  I can successfully run `yarn storybook` and see it in action.  

I did have to run `yarn` first though to update packages, FWIW.

PR #50, opened 2022-08-10
Implements LineChart

pnavarrc:
Implements the LineChart component. See [Chart API Design](https://www.dropbox.com/scl/fi/snrqj20p0uoxr9phynbtu/Chart-API-Design.paper?dl=0&rlkey=3onvrnbe1753bc7v3tpey3do6) for a quick introduction to how we expect to build charts.

The component takes the props of the SVG [`path`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/path) element so we can have the same styling flexibility as the `path` element, but set some reasonable defaults for `stroke` and `strokeWidth` to make sure it renders with default values (by default the line is not visible). It also accepts the props of [`LinePath`](https://airbnb.io/visx/docs/shape#LinePath) (from the visx library) so we can customize some line generation settings.

## Changes

- Implement the `LineChart` as a basic building block for more complex charts.
- Adds the `@visx/mock-data` package to use some mock data as dev dependency. Note: Once we merge #30, we could replace this with the `MockDataProvider`

image

Note: It might be better to pass `x` and `y` instead of the scales (to have some extra flexibility on the accessors), but it feels unnecessary given that we are passing a `Timeseries<number>`, which is guaranteed to have `TimeseriesPoint<number>` and thus `date` and `value`.

### DIFF
--- a/packages/ui-components/src/components/LineChart/LineChart.stories.tsx
+++ b/packages/ui-components/src/components/LineChart/LineChart.stories.tsx
@@ -1,10 +1,10 @@
 import React from "react";
 import { ComponentStory, ComponentMeta } from "@storybook/react";
-import LineChart from ".";
-import { assert } from "@actnowcoalition/assert";
-import { Timeseries, TimeseriesPoint } from "@actnowcoalition/metrics";
 import { scaleLinear, scaleTime } from "@visx/scale";
 import { appleStock } from "@visx/mock-data";
+import { assert } from "@actnowcoalition/assert";
+import { Timeseries, TimeseriesPoint } from "@actnowcoalition/metrics";
+import LineChart from ".";
 
 export default {
   title: "Charts/LineChart",
--- a/packages/ui-components/src/components/LineChart/LineChart.stories.tsx
+++ b/packages/ui-components/src/components/LineChart/LineChart.stories.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { ComponentStory, ComponentMeta } from "@storybook/react";
 import LineChart from ".";
+import { assert } from "@actnowcoalition/assert";
 import { Timeseries, TimeseriesPoint } from "@actnowcoalition/metrics";
 import { scaleLinear, scaleTime } from "@visx/scale";
 import { appleStock } from "@visx/mock-data";
@@ -10,9 +11,8 @@ export default {
   component: LineChart,
 } as ComponentMeta<typeof LineChart>;
 
-const width = 300;
-const height = 100;
-const padding = 4;
+const [width, height] = [600, 400];
+const padding = 20;
 
 const Template: ComponentStory<typeof LineChart> = (args) => (
   <svg width={width} height={height} style={{ border: "solid 1px #eee" }}>
@@ -26,30 +26,30 @@ interface AppleStockPoint {
 }
 
 const formatPoint = (p: AppleStockPoint): TimeseriesPoint<number> => ({
-  date: new Date(`${p.date.substring(0, 10)}T00:00:00Z`),
+  date: new Date(p.date.substring(0, 10)),
   value: p.close,
 });
 
 const timeseries = new Timeseries(appleStock.map(formatPoint));
+assert(timeseries.hasData(), `Timeseries cannot be empty`);
 
-const startDate = timeseries.hasData()
-  ? timeseries.minDate
-  : new Date("2007-04-24");
-const endDate = timeseries.hasData()
-  ? timeseries.maxDate
-  : new Date("2007-04-30");
-
-const minValue = timeseries.hasData() ? timeseries.minValue : 0;
-const maxValue = timeseries.hasData() ? timeseries.maxValue : 10;
+const { minDate, maxDate, minValue, maxValue } = timeseries;
 
 const xScale = scaleTime({
-  domain: [startDate, endDate],
+  domain: [minDate, maxDate],
   range: [padding, width - padding],
 });
+
 const yScale = scaleLinear({
   domain: [minValue, maxValue],
   range: [height - 2 * padding, padding],
 });
 
-export const Example = Template.bind({});
-Example.args = { timeseries, xScale, yScale, stroke: "blue" };
+export const SolidLine = Template.bind({});
+SolidLine.args = {
+  timeseries,
+  xScale,
+  yScale,
+  stroke: "#2a9d8f",
+  strokeWidth: 2,
+};
--- a/packages/ui-components/src/components/LineChart/LineChart.stories.tsx
+++ b/packages/ui-components/src/components/LineChart/LineChart.stories.tsx
@@ -20,17 +20,16 @@ const Template: ComponentStory<typeof LineChart> = (args) => (
   </svg>
 );
 
-interface AppleStockPoint {
-  date: string;
-  close: number;
-}
+// We format the points from appleStock to match TimeseriesPoint<number>
+// so we can use them to initialize Timeseries.
+const points = appleStock.map(
+  (p: { date: string; close: number }): TimeseriesPoint<number> => ({
+    date: new Date(p.date.substring(0, 10)),
+    value: p.close,
+  })
+);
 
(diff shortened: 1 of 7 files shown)


packages/ui-components/src/components/LineChart/LineChart.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+  (p: { date: string; close: number }): TimeseriesPoint<number> => ({
+    date: new Date(p.date.substring(0, 10)),
+    value: p.close,
+  })
+);
+
+const timeseries = new Timeseries(points);
+assert(timeseries.hasData(), `Timeseries cannot be empty`);
+
+const { minDate, maxDate, minValue, maxValue } = timeseries;
```
pnavarrc, line 35: ☝️ This is gold, thanks @mikelehen !

packages/ui-components/src/components/LineChart/LineChart.stories.tsx
```diff
@@ -0,0 +1,62 @@
+import React from "react";
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { scaleLinear, scaleTime } from "@visx/scale";
+import { appleStock } from "@visx/mock-data";
+import { assert } from "@actnowcoalition/assert";
```
mikelehen, line 5: I think we're missing a package.json dependency for `@actnowcoalition/assert`?  I guess it can be a devDependency since we only use it in storybook.  And since storybook seems to work regardless, I guess it technically doesn't matter? 🤷

packages/ui-components/src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,80 @@
+import React from "react";
+import { LinePath } from "@visx/shape";
+import { curveMonotoneX } from "@visx/curve";
+import { ScaleLinear, ScaleTime } from "d3-scale";
+import { Timeseries, TimeseriesPoint } from "@actnowcoalition/metrics";
+import { LinePathProps } from "@visx/shape/lib/shapes/LinePath";
+
+interface LineChartOwnProps {
```
mikelehen: I think we need to export this from the package to get it to show up in the docs.  `yarn generate-docs` actually logs a warning for this:

> ./packages/ui-components/src/components/LineChart/LineChart.tsx:8:0 - warning LineChartOwnProps is referenced by @actnowcoalition/ui-components.LineChartProps but not included in the documentation.

I want to turn on `--treatWarningsAsErrors` so typedoc will fail the build in this case, but we have a couple warnings to fix first and when I looked at it, I got confused by one of them so I temporarily gave up.  But I'll come back to it.

packages/ui-components/src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,80 @@
...
+  /** {@link Timeseries} used to draw the line chart */
```
mikelehen: Just FYI it might not be worth bothering using `@link` for things like this since the docs should automatically link to each property's type.  I usually reserve `@link` for when I'm referring to types that are related but not being directly referenced by the code.

Doesn't hurt anything of course though, so feel free to leave it!

packages/ui-components/src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,80 @@
...
+ * LineChart is a chart building block that return an SVG path element, given
+ * a timeseries and x and y scales.
+ *
+ * The LineChart is intended to be used as building block of more complex
+ * charts, so it doesn't include axes, tooltips or anything else. See
+ * Storybook for a working example.
+ *
+ * @example
+ *    const xScale = scaleLinear({ domain: [minDate, maxDate], range: [0, 200] });
+ *    const yScale = scaleLInear({ domain: [minVal, maxVal], range: [0, 100] });
```
mikelehen: scaleLInear => scaleLinear

packages/ui-components/src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,80 @@
...
+ *
+ *    return (
+ *      <svg>
+ *        <BarChart ... />
+ *        <LineChart
+ *          timeseries={timeseries}
+ *          xScale={}
+ *          yScale={}
```
mikelehen: It looks like we're not referencing the scales you created above?

packages/ui-components/src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,80 @@
...
+ *          stroke="blue"
+ *          strokeDasharray="4 4"
+ *        />
+ *      <svg>
```
mikelehen: `</svg>` ?

packages/ui-components/src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { curveMonotoneX } from "@visx/curve";
+import { ScaleLinear, ScaleTime } from "d3-scale";
+import { Timeseries, TimeseriesPoint } from "@actnowcoalition/metrics";
+import { LinePathProps } from "@visx/shape/lib/shapes/LinePath";
+
+interface LineChartOwnProps {
+  /** {@link Timeseries} used to draw the line chart */
+  timeseries: Timeseries<number>;
+
+  /** Scale to transform point dates to positions on the x-axis */
```
mikelehen: I like how below you say "d3-scale" instead of just "scale"

You also say "pixel positions" instead of just "positions."

packages/ui-components/src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * @returns SVG Path element
+ */
+const LineChart: React.FC<LineChartProps> = ({
+  timeseries,
+  xScale,
+  yScale,
+  stroke = "#000",
+  strokeWidth = 2,
+  shapeRendering = "geometricPrecision",
```
mikelehen, line 60: Mildly curious why this is the default.  https://developer.mozilla.org/en-US/docs/Web/CSS/text-rendering seems to mostly talk about font kerning, but we aren't doing any text rendering, are we? 🤷

packages/ui-components/src/index.ts
```diff
@@ -8,6 +8,8 @@ export { default as LegendThreshold } from "./components/LegendThreshold";
 export type { LegendThresholdProps } from "./components/LegendThreshold";
 export { default as LegendCategorical } from "./components/LegendCategorical";
 export type { LegendCategoricalProps } from "./components/LegendCategorical";
+export { default as LineChart } from "./components/LineChart";
+export type { LineChartProps } from "./components/LineChart";
```
mikelehen, line 12: I'm mildly curious if we should simplify our exports to avoid redundancy.  What I did in my metrics PR is:
1. Don't bother using `default` exports, since we end up only using named exports at the package level.  So mapping an export to `default` and then mapping it back to a named export felt unnecessary.
2. At the package level I just did `export * from "./folder"` rather than specify the exports by name again.

I'm not very confident this makes sense to do everywhere all the time.  Maybe we want some "internal" exports within packages that don't get exported?  But I'm not sure.

Anyway, your code is consistent with the package so far so I wouldn't change anything.  Just food for thought.

mikelehen (approved):
Looks good with a few nits!

packages/ui-components/src/components/LineChart/LineChart.stories.tsx
pnavarrc: good catch! adding as dev dependency

packages/ui-components/src/components/LineChart/LineChart.tsx
pnavarrc: This is shape rendering, circles and curves look nicer https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering

mikelehen: Oops, I guess I ended up in the wrong place with my googling and didn't notice!

PR #61, opened 2022-08-15
Initial CSV data provider

smcclure17:
Adds a data provider for general wide-format CSV files. 

We will probably want to add functionality to handle long-format CSVs soon, but I'm not 100% sure if we'd want to do that in a separate/new data provider, or extend the functionality of this one (by e.g. increasing the number of `CsvDataProviderOptions`).


packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,74 @@
...
+import { Dictionary } from "lodash";
+import {
+  DateFormat,
+  formatUTCDateTime,
+  parseDateString,
+} from "@actnowcoalition/time-utils";
+
+export type DataRow = { [key: string]: unknown };
+
+export function dataRowsToMetricData(
```
smcclure17: Ideally this function could be reused by other data providers that read data in row/record format (like a JSON  provider)

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+    includeTimeseries: boolean
+  ): MetricData<unknown> {
+    const cacheKey = `region:${region.regionId}_metric:${metric.id}`;
+    const metricData = this.cachedData[cacheKey];
+    assert(
+      metricData,
+      `Data for region and metric not found. Be sure populateCache() 
+      has been invoked with expected region and metric.`
+    );
+    if (!includeTimeseries && metricData.hasTimeseries()) {
```
smcclure17: It seems like the `includeTimeseries` arg in `getDataFromCache()` should always match the arg in`populateCache()` (as is the case in `fetchData()`), so we maybe shouldn't have to specify whether we want timeseries data or not, but I've added this in case the args aren't equal.

packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -0,0 +1,112 @@
...
+        dateColumn: dateCol,
+      });
+      return (
+        await provider.fetchData([region], [TEST_METRIC], includeTimeseries)
+      )
+        .regionData(region)
+        .metricData(TEST_METRIC);
+    };
+
+    expect(
```
smcclure17: The line splitting makes these a little hard to parse I think, so I'm willing to refactor these in the name of clarity if we'd like

packages/metrics/src/data/MetricData.ts
```diff
@@ -29,6 +29,13 @@ export class MetricData<T = unknown> {
     return this._timeseries !== undefined;
   }
 
+  /**
+   * Removes timeseries data for this metric and returns a new MetricData object.
+   */
+  dropTimeseries(): MetricData {
```
smcclure17: Maybe this should just be a setter like
```typescript
  dropTimeseries(): void {
    this._timeseries = undefined
  }
```
but I felt there was safety in returning a new object

mikelehen: 💯

I'm a big fan of immutable objects where there is no way to modify them directly.  Instead, any "mutator" functions just create & return a new instance that is independent from the original one.  In general, immutable objects / values make code much easier to understand / reason about (mutability leads to a lot of difficult-to-diagnose bugs).  TBH this is probably the biggest lesson in programming that I've learned since getting my CS degree. 😁

packages/metrics/src/data/MetricData.ts
```diff
@@ -29,6 +29,10 @@ export class MetricData<T = unknown> {
     return this._timeseries !== undefined;
   }
 
+  dropTimeseries(): MetricData<T> {
```
mikelehen, line 38: Can you add a doc comment?  I'm trying to be pretty exhaustive with documenting classes and their public methods / properties.

packages/metrics/src/data_providers/data_provider_utils.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    );
+    const payload = await fetchCsv("url-placeholder");
+    expect(payload[0].cool_metric).toBe("a value that has spaces.");
+  });
+
+  test("fetchCsv() parses booleans", async () => {
+    const mockFetchBoolean = async (value: string | boolean) => {
+      mockFetch(`region,date,cool_metric\n36,2022-08-02,${value}`);
+      const payload = await fetchCsv("url-placeholder");
+      console.log(payload[0].cool_metric);
```
mikelehen: Remove console.log?

packages/metrics/src/data_providers/CsvDataProvider.test.ts
mikelehen: Yeah, that sounds good to me.  I almost always break up expressions when they use `await` since there's a lot of magic that happens under the hood with `await` and it means the code is now asynchronous instead of synchronous.  So for readability I don't want to hide it in a nested expression.

packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -0,0 +1,152 @@
+import { CsvDataProvider } from "./CsvDataProvider";
+import { Metric } from "../Metric";
+import { Region, states } from "@actnowcoalition/regions";
+import { formatUTCDateTime, DateFormat } from "@actnowcoalition/time-utils";
+
+const mockData = `region,cool_metric\n36,150\n12,`;
+const mockTimeseries = `region,date,cool_metric\n36,2022-08-02,150\n12,2022-08-02,`;
```
mikelehen: Hrm.  These stringified CSV files are not super readable...  I think you can do multiline strings like:

```
const mockData = 
`region,cool_metric
36,150
12,`;
```

That might be a little better?

I've also seen people do things like:

```
const mockData = [
  'region,cool_metric',
  '36,150', 
  '12,'
].join('\n');
```
I kinda' prefer that, but I think prettier will turn that into a single line which defeats the purpose a bit. :/

packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -0,0 +1,152 @@
...
+const newYork = states.findByRegionIdStrict("36");
+const florida = states.findByRegionIdStrict("12");
+const testMetric = new Metric({ id: "cool_metric" });
+
+const mockFetchData = async (
```
mikelehen: I wouldn't be able to guess what this function does from the name.  I'd maybe:
1. Rename it to "testFetchingCsvData()" or something?
2. Add a brief comment explaining what it does.

packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -0,0 +1,152 @@
...
+  data: string,
+  includeTimeseries: boolean,
+  dateCol?: string,
+  region: Region = newYork
+) => {
+  global.fetch = jest.fn(() =>
+    Promise.resolve({
+      text: () => Promise.resolve(data),
+    } as unknown as Response)
+  );
```
mikelehen: This is very interesting, but I'm not sue I'm a fan. 😆 

As a general rule, I only resort to mocking dependencies (`fetch()` in this case) when absolutely necessary.  Mocking dependencies creates indirect dependencies between the test code and the production code that are not obvious when you are modifying the code.  For example in this case if we ever changed `CsvDataProvider` to use say [axios](https://axios-http.com/docs/intro) to fetch data instead of native `fetch()` then all of our tests will start failing even though we didn't break anything.

Additionally it's generally a bad practice to modify global state, especially from tests.  In this case we are replacing the global `fetch()` function with a fake version but (unless I'm missing it), we are not "putting it back" afterwards so all subsequent tests are going to get a fake version that won't work as normal.

On one last note, I don't think you need the `jest.fn(...)` wrapper here.  I think you use that when you want to track if/when/how a function is called by the code you are testing. But we don't care, so you could just use a plain (unwrapped) lambda function here.

So what would I do instead?  I'd try to change the production code so that it's more easily testable.  In this case, perhaps we should have other ways to pass data into the `CsvDataProvider` other than by URL.  So maybe add a `data` property as an alternative way to specify the CSV contents instead of `url`.  For now we'll only use it in tests, but that's fine.  We can have the comment say "(especially useful for testing)" or something to explain why it's an option.

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,84 @@
+import { assert } from "@actnowcoalition/assert";
+import { Region } from "@actnowcoalition/regions";
+import { CachingMetricDataProviderBase } from "./CachingMetricDataProviderBase";
+import { Metric } from "../Metric";
+import { MetricData } from "../data";
+import { fetchCsv, DataRow, dataRowsToMetricData } from "./data_provider_utils";
+import { groupBy } from "lodash";
+
+interface CsvDataProviderOptions {
+  /* URL of the CSV file to import from. */
```
mikelehen: For these to be doc comments (i.e. "documentation comments") picked up by TypeDoc, they need to start with `/**` not just `/*`

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+  url: string;
+  /* Name of column containing valid Region IDs. */
+  regionColumn: string;
```
mikelehen, line 18: I think we can/should punt for now, but I could imagine letting the caller alternatively pass in a "regionGetter" function that could "calculate" the region ID for any given row using arbitrary logic.  So if you don't have a clean CSV with region IDs already you could have a little lookup function that handles it.

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+import { Region } from "@actnowcoalition/regions";
+import { CachingMetricDataProviderBase } from "./CachingMetricDataProviderBase";
+import { Metric } from "../Metric";
+import { MetricData } from "../data";
+import { fetchCsv, DataRow, dataRowsToMetricData } from "./data_provider_utils";
+import { groupBy } from "lodash";
+
+interface CsvDataProviderOptions {
+  /* URL of the CSV file to import from. */
+  url: string;
```
mikelehen: Per my comment above, perhaps make `url` optional and allow `data` (feel free to come up with a more explicit name) as an alternative way to pass data in.

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+ */
+export class CsvDataProvider extends CachingMetricDataProviderBase {
+  private readonly url: string;
+  private readonly regionColumn: string;
+  private readonly dateColumn?: string;
+
+  readonly cachedData: { [key: string]: MetricData } = {};
+
+  constructor(options: CsvDataProviderOptions) {
+    super("csv-data");
```
mikelehen: I think we want this ID to be configurable (required even) since you may have different CSV sources that you want to refer to for metrics.  E.g. you might have a CSV with region metadata, another CSV with climate data, etc. and you want to create separate data providers and be able to refer to them from metrics.

So I'd change the constructor to `constructor(providerId: string, options: CsvDataProviderOptions)`

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+    metrics: Metric[],
+    includeTimeseries: boolean
+  ): Promise<void> {
+    const csv: DataRow[] = await fetchCsv(this.url);
+    assert(
+      csv.length > 0 &&
+        csv.every((row) => row[this.regionColumn] !== undefined),
+      "CSV must not be empty and must contain a region column entry for all rows."
+    );
+    const groupedCsvRows = groupBy(csv, (row) => row[this.regionColumn]);
```
mikelehen: So I'm not sure how much configurability we'll want in terms of caching, but for now I think I would err on the side of aggressive caching which means we might as well only fetch the CSV once.  So I would actually cache the `groupedCsvRows` and use that from `getDataFromCache()`.  Then we can get rid of or move the below code.

I set this up so that fetching / caching could be more granular when it makes sense.  Like if every region's data came from a different file (like with the CAN API), then we would only fetch the files necessary, and cache them.  But in the case of this CSV data provider, might as well cache the whole contents.

We'll probably need to revisit caching behavior in the future, but at this point I'm not sure what we will want.

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+    regions.forEach((region) => {
+      metrics.forEach((metric) => {
+        const cacheKey = `region:${region.regionId}_metric:${metric.id}`;
+        this.cachedData[cacheKey] = dataRowsToMetricData(
+          groupedCsvRows,
+          region,
+          metric,
+          includeTimeseries,
```
mikelehen: Related to your comment below, if we are respecting `includeTimeseries` when filling the cache, then we probably should include it as part of the cache key.  Although since `populateCache()` just overwrites the cache every time, it probably doesn't matter.  Normally I'd expect it to only populate the cache if it's not already populated.

packages/metrics/src/data_providers/CsvDataProvider.ts
mikelehen: Yeah, I think as written the cached entry will always match the args here.

FWIW- You don't really need to check `metricData.hasTimeseries()` before calling `metricData.dropTimeseries()`.  It's safe to just always drop it, even if it's not there.

packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      (
+        await mockFetchData(mockData, /*includeTimeseries=*/ false)
+      ).hasTimeseries()
+    ).toBe(false);
+    expect(
+      (
+        await mockFetchData(
+          mockData,
+          /*includeTimeseries=*/ false,
+          undefined,
```
mikelehen: nit: Similar to my recommendation on boolean literals, it can make sense for other literals too.  Like here, `/*dateCol=*/undefined`

Basically it's a way to simulate [named arguments](https://en.wikipedia.org/wiki/Named_parameter) (which JavaScript doesn't support 😢) for code readability purposes.

packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    expect(
+      (
+        await mockFetchData(
+          mockData,
+          /*includeTimeseries=*/ true,
+          undefined,
+          florida
+        )
+      ).timeseries
+        .removeNils()
```
mikelehen: Is this removeNils() necessary / helpful?

packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -0,0 +1,152 @@
...
+          "date"
+        )
+      ).hasTimeseries()
+    ).toBe(false);
+    expect(
+      (
+        await mockFetchData(
+          mockTimeseries,
+          /*includeTimeseries=*/ false,
+          "date"
```
mikelehen: FWIW- All of this repetition (`await mockFetchData(mockTimeseries, /*includeTimeseries=*/ false, "date" )` above and then immediately repeated is another good reason to refactor these tests.  In general the more duplication there are in tests the harder they are to read and to verify that all of the scenarios are being covered.

packages/metrics/src/data_providers/data_provider_utils.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    );
+    expect(() => {
+      dataRowsToMetricData(dataRows, newYork, testMetric, true, "date");
+    }).toThrow();
+    expect(() => {
+      dataRowsToMetricData(dataRows, newYork, testMetric, false, "date");
+    }).toThrow();
+  });
+
+  test("dataRowsToMetricData() fails when region data is missing.", async () => {
```
mikelehen: nit: I had trouble understanding what this meant. Perhaps rephrase to "when there's no data for region."

packages/metrics/src/data_providers/data_provider_utils.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    );
+    expect(() => {
+      dataRowsToMetricData(dataRows, newYork, testMetric, true, "date");
+    }).toThrow();
+    expect(() => {
+      dataRowsToMetricData(dataRows, newYork, testMetric, false, "date");
+    }).toThrow();
+  });
+
+  test("dataRowsToMetricData() succeeds when region data is null.", async () => {
```
mikelehen: Perhaps rephrase, "when metric data is null"

packages/metrics/src/data_providers/data_provider_utils.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    );
+  };
+
+  test("fetchCsv() return null when no cell data is provided", async () => {
+    mockFetch(`region,date,cool_metric\n36,2022-08-02,`);
+    const payload = await fetchCsv("url-placeholder");
+    expect(payload[0].cool_metric).toBe(null);
+  });
+
+  test("fetchCsv() keeps strings intact", async () => {
```
mikelehen: I would have had trouble guessing what this test does from the name.  Perhaps "handles CSV values with strings." or something.

packages/metrics/src/data_providers/data_provider_utils.test.ts
```diff
@@ -0,0 +1,85 @@
...
+describe("fetchCsv()", () => {
+  const mockFetch = async (data: string) => {
+    global.fetch = jest.fn(() =>
+      Promise.resolve({
+        text: () => Promise.resolve(data),
+      } as unknown as Response)
+    );
+  };
+
+  test("fetchCsv() return null when no cell data is provided", async () => {
```
mikelehen: This is kinda' misleading.  `fetchCsv()` doesn't return null.  Maybe "fetchCsv() treats missing values as null" or something.

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,84 @@
...
+      "CSV must not be empty and must contain a region column entry for all rows."
+    );
+    const groupedCsvRows = groupBy(csv, (row) => row[this.regionColumn]);
+    regions.forEach((region) => {
+      metrics.forEach((metric) => {
+        const cacheKey = `region:${region.regionId}_metric:${metric.id}`;
+        this.cachedData[cacheKey] = dataRowsToMetricData(
+          groupedCsvRows,
+          region,
+          metric,
```
mikelehen: As-is, we are passing `metric` in to `dataRowsToMetricData` and then it is using `metric.id` as if it is the CSV column name, but I think we want more flexibility.  The idea is that I can define a metric with an arbitrary ID (used in the frontend and CMS variables, etc.) and then tie that to an arbitrary column in an arbitrary CSV.  So defining a metric I might do:

```
const metrics = [
  {
    id: "population",
    dataReference: {
      providerId: "region-metadata-csv",
      column: "2020_census_population"
    }
  }];

const dataProviders = [
  new CsvDataProvider('region-metadata-csv', {
    url: '.../region-metadata.csv',
    regionColumn: 'region-id'
  }
]);

const catalog = new MetricCatalog(metrics, dataProviders);
```

or whatever.  I think that probably means we need to pass in a `metricColumn` in addition to `metric` or similar.

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+ */
+export function dataRowsToMetricData(
+  rowsDictionary: Dictionary<DataRow[]>,
+  region: Region,
+  metric: Metric,
+  includeTimeseries: boolean,
+  dateKey?: string
+) {
+  const rows = rowsDictionary[region.regionId];
+  assert(rows, `No data found for region ${region.regionId}`);
```
mikelehen: FWIW- I'm not sure if this is the behavior we want, but it's probably better to start out strict and loosen it up if it turns out that's annoying.

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+  if (!dateKey) {
+    assert(
+      rows.length === 1,
+      `Duplicate or no entries for region ${region.regionId} and metric ${metric.id} found.`
```
mikelehen: I'd probably tweak this to be a bit more helpful diagnostically:

```
`Expected exactly 1 entry for region ${region.regionId} and metric ${metric.id} but found: ${rows.length}`
```

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+    );
+    const value = rows[0][metric.id];
+    assert(
+      value !== undefined,
+      "Entry for metric for region ${region.regionId} and metric ${metric.id} found."
```
mikelehen: "Entry for metric for region"?  Also should it say "not found"?

packages/metrics/src/data_providers/data_provider_utils.test.ts
```diff
@@ -0,0 +1,85 @@
...
+import { DataRow, dataRowsToMetricData, fetchCsv } from "./data_provider_utils";
+import { Dictionary, groupBy } from "lodash";
+
+const newYork = states.findByRegionIdStrict("36");
+const testMetric = new Metric({ id: "cool_metric" });
+
+describe("dataRowsToMetricData()", () => {
+  test("dataRowsToMetricData() fails when metric data column is missing.", async () => {
+    const dataRows: Dictionary<DataRow[]> = groupBy(
+      [{ date: "2022-08-01", region: "36", another_metric: 100 }],
```
mikelehen, line 43: Am I missing it or do we not have any tests that create a timeseries with more than 1 data point in it?  I feel like that'd be worth testing.

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+    );
+    const timeseries = includeTimeseries
+      ? new Timeseries([
+          {
+            date: stripTime(new Date()),
+            value: value,
+          },
+        ])
+      : undefined;
```
mikelehen: I would actually just always have the timeseries be `undefined` since we don't know what date the data is from.

I know this isn't what I did with StaticValueDataProvider, but it's more of a toy data provider and it was handy for testing, etc. to have it provide timeseries data too.

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+  assert(
+    rows.every((row) => typeof row[dateKey] === "string"),
+    `Not all rows have a valid date-string in column ${dateKey}.`
+  );
+  const timeseries = new Timeseries(
+    rows.map((row) => {
+      assert(
+        row[metric.id] !== undefined,
+        "Entry for metric for region ${region.regionId} and metric ${metric.id} found."
```
mikelehen: "Entry for metric for region" again...

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+  });
+  return csv.data as DataRow[];
+}
+
+/**
+ * Remove hours minutes and seconds from Javascript Date object.
+ *
+ * @param date Date object to truncate.
+ */
+function stripTime(date: Date): Date {
```
mikelehen, line 84: I created a `Timeseries.isoDateString()` function that does the same thing (but in a different way).  I wonder if we should just move this logic to the time-utils package.  Feel free to do it as a TODO for now (it would need to happen in multiple PRs anyway).

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+  );
+}
+
+/**
+ * Fetches data from a URL and parses it as JSON.
+ *
+ * @param url URL to fetch data from.
+ * @returns Fetched CSV data.
+ */
+export async function fetchCsv(url: string): Promise<DataRow[]> {
```
mikelehen: This seems specific to the CsvDataProvider so I would probably move it to that file.

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+import { Timeseries } from "../Timeseries";
+import { Dictionary } from "lodash";
+import { parseDateString } from "@actnowcoalition/time-utils";
+
+export type DataRow = { [key: string]: unknown };
+
+/**
+ * Parse raw data-rows into MetricData for specified region and metric.
+ *
+ * @param rowsDictionary Region-indexed dictionary of data to transform.
```
mikelehen: Consider renaming to `dataRowsByRegionId` or similar.

mikelehen (changes requested):
I left a lot of feedback (sorry!) but overall this is looking good.  Holler if anything doesn't make sense or you want to punt on some of the items or whatever.  Thanks!

packages/metrics/src/data_providers/CsvDataProvider.ts
smcclure17: Yeah, I agree that sounds useful. I think I'll pass on that for now (in this PR) but it does seem like something we'll want

packages/metrics/src/data_providers/data_provider_utils.ts
smcclure17: Yeah I think it makes sense to move it--I can tackle that after this gets merged

packages/metrics/src/data_providers/CsvDataProvider.test.ts
smcclure17: Yeah good point. I've reworked these tests--I've thinned out a lot of the tests here (in `CsvDataProvider.test.ts`) in favor of doing more testing in `data_uprovider_utils_test.ts`, as `dataRowsToMetricData()` handles a lot of the internal processing used by the CSV provider

packages/metrics/src/data_providers/CsvDataProvider.ts
smcclure17: Ah, yes that makes a lot of sense! thanks for the explanation + example 🙏

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,113 @@
...
+import { DataRow, dataRowsToMetricData } from "./data_provider_utils";
+import { Dictionary, groupBy } from "lodash";
+import Papa from "papaparse";
+
+interface CsvDataProviderOptions {
+  /** URL of a CSV file to import from. */
+  url?: string;
+  /** Name of column containing valid Region IDs. */
+  regionColumn: string;
+  /** Name of column containing valid ISO 8601 date-time values.
```
mikelehen: nit: If you're doing a `/** ... */` comment and it goes to multiple lines, it's typical to leave the first / last lines blank, so:

```
/** 
 * Name of column containing valid ISO 8601 date-time values.
 * dateColumn must be provided if the CSV contains timeseries data, even if
 * timeseries data will not be used (e.g. by setting includeTimeseries=false) */
 */
```

Also, I find the comment about includeTimeseries=false a bit confusing out-of-context.  I'm not sure if it is necessary / helpful here.  Perhaps something like:

> Name of column containing valid ISO 8601 date-time values. Required if the CSV contains timeseries data, else it should not be specified.

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @returns Metric Data for the specified region and metric.
+ */
+export function dataRowsToMetricData(
+  dataRowsByRegionId: Dictionary<DataRow[]>,
+  region: Region,
+  metric: Metric,
+  includeTimeseries: boolean,
+  dateKey?: string
+) {
+  const metricColumn = metric.dataReference?.column;
```
mikelehen: Hrm.  This makes this helper less reusable since it'll basically only work with metrics that use the CsvDataProvider (or other providers that happen to have a `metricColumn` in their data reference), which e.g. probably wouldn't be true for a JSON provider.

So I think it's actually better to explicitly pass in `metricColumn` (or maybe `metricKey`?) as a parameter so that this function doesn't have to make assumptions about the metric.

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,79 @@
...
+      rows.length === 1,
+      `Expected exactly 1 entry for region ${region.regionId} and metric ${metric.id} but found: ${rows.length}`
+    );
+    const value = rows[0][metricColumn];
+    assert(
+      value !== undefined,
+      `No data for region ${region.regionId} and metric ${metric.id} found.`
+    );
+    return new MetricData(metric, region, value, /**_timeseries=*/ undefined);
+  }
```
mikelehen: nit: This is a highly-debated stylistic thing, but I would wrap the rest of this function in an `} else { ... }` so that the control flow is more explicit.

Basically there's a logical assumption that functions execute until the "end" and you can just look at the return call at the end of a function to see what it returns, and work backwards to understand the logic.  But early returns break that assumption.

The one case where I think early returns aid in readability is when there are trivial edge cases / error conditions that can be handled at the very beginning of the function in order to simplify the rest of the code.  But returning from the middle of a function can make it harder to understand the control flow.  I'd rather an explicit `if { ... return ... } else { ... return ... }` so that it's obvious that there's a branch in the control flow.

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,113 @@
...
+ * |TX     |2022-02-01 |12   |45   |
+ * |CA     |2022-02-01 |31   |66   |
+ */
+export class CsvDataProvider extends CachingMetricDataProviderBase {
+  private readonly regionColumn: string;
+  private readonly dateColumn?: string;
+  private readonly url?: string;
+  private readonly csvData?: string;
+
+  private cachedData: Dictionary<DataRow[]> = {};
```
mikelehen: Consider making this a bit more explicit by renaming `cachedData` to `dataRowsByRegionId` and changing the type to ` { [regionId: string]: DataRow[] }`

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  constructor(providerId: string, options: CsvDataProviderOptions) {
+    assert(
+      options.url || options.csvData,
+      "Either a URL or CSV data must be provided to create an instance of CsvDataProvider."
+    );
```
mikelehen, line 49: Nice assert!

packages/metrics/src/data_providers/data_provider_utils.test.ts
```diff
@@ -0,0 +1,214 @@
...
+        newYork,
+        testMetric,
+        /*includeTimeseries=*/ false
+      ).currentValue
+    ).toBe(null);
+  });
+});
+
+describe("fetchCsv()", () => {
+  test("fetchCsv() treats missing values as null", async () => {
```
mikelehen: This isn't a huge deal, but all else being equal I'd rather that we test our public API (the stuff that consumers of our packages will be seeing / using), which would mean we'd be writing tests against the `CsvDataProvider` instead of against the `fetchCsv()` helper function which is really just an internal implementation detail.

That way:
1. Our tests serve dual-purpose as "example" code and document the contracts of the API (what's expected to work / not) more directly.
2. If we refactor the internals of `CsvDataProvider` in the future and e.g. rework / remove the `fetchCsv()` helper (like I am suggesting to split it into two functions) it wouldn't break any tests.

Not a huge deal, and no need to change it in this PR, but something to consider for the future.  

Sometimes it _does_ make sense to write tests against an internal helper function than against the public API.  E.g. If we intend to use `dataRowsToMetricData()` helper for the JSON provider as well, then writing tests directly against `dataRowsToMetricData()` saves us the effort of writing all those tests twice.

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,113 @@
...
+    super(providerId);
+    this.regionColumn = options.regionColumn;
+    this.dateColumn = options.dateColumn;
+    this.url = options.url;
+    this.csvData = options.csvData;
+  }
+
+  async populateCache(): Promise<void> {
+    const csv = await fetchCsv(this.url, this.csvData);
```
mikelehen: nit: This is kind of a weird / overloaded function.  It takes two parameters, only one of which is expected to be set, and then it might fetch data from the internet and then it'll parse either the fetched data or the provided data as a CSV.  

I would split it into something like:
```
const csvText = this.fetchCSV();
const csv = parseCsv(csvText);
```

Also maybe rename `csv` to `csvRows` or `rows` or similar since it's no longer comma-separated-values.

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,113 @@
...
+    assert(csv.length > 0, "CSV must not be empty.");
+    assert(
+      csv.every((row) => row[this.regionColumn] !== undefined),
+      "CSV must contain a region column entry for all rows."
+    );
```
mikelehen, line 71: Nit: As a general rule it's nice to avoid iterating over a large body of data multiple times if you can get by with only doing it once.  So I think we could leverage `groupBy`'s behavior to do this more efficiently:

```
const dataRowsByRegionId = groupBy(csv, (row) => row[this.regionColumn]);
assert(!dataRowsByRegionId['undefined'], `One or more CSV rows were missing a region id value in column ${this.regionColumn}`);
```

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,113 @@
...
+      this.dateColumn
+    );
+    if (includeTimeseries) {
+      assert(
+        metricData.hasTimeseries(),
+        "includeTimeseries set to true but cached data has no timeseries."
+      );
+      return metricData;
+    }
+    return metricData.dropTimeseries();
```
mikelehen: nit: Again I'd put this in an `if { ... } else { ... }`

Also, I'm not sure if this `dropTimeseries()` call is necessary since I think `dataRowsToMetricData()` is respecting the `includeTimeseries` flag already?

That said, it might be simpler to have `dataRowsToMetricData()` _not_ take `includeTimeseries` as a parameter and have it always include the timeseries if you pass `dateColumn`, and then we _would_ want to call `dropTimeseries()` here.  

I don't have a strong preference though.

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+}
+
+/**
+ * Fetches data from a URL or CSV data and parses it as JSON.
+ *
+ * At least one of url and csvData must be provided.
+ * If both url and csvData are provided data is fetched from the url and csvData is ignored.
```
mikelehen: As mentioned above, this is kind of a funky description for a funky function.  A good hint that it might be better to split up into multiple functions.

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,79 @@
...
+export type DataRow = { [key: string]: unknown };
+
+/**
+ * Parse raw data-rows into MetricData for specified region and metric.
+ *
+ * @param dataRowsByRegionId Region-indexed dictionary of data to transform.
+ * @param region Region to parse data for.
+ * @param metric Metric to parse data for.
+ * @param includeTimeseries Whether to create a timeseries.
+ * @param dateKey Key name of the items in the data-rows that contain the date-time values.
```
mikelehen: Perhaps add "To be used when the data-rows represent timeseries data."

That said, I could also see splitting this function into two versions.  One for timeseries data (which requires dateKey) and one for non-timeseries data (which does not take dateKey).  But that'd probably end up with a bit of code duplication between them, so I'm not sure if that's an improvement.

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,79 @@
...
+    assert(
+      rows.length === 1,
+      `Expected exactly 1 entry for region ${region.regionId} and metric ${metric.id} but found: ${rows.length}`
+    );
+    const value = rows[0][metricColumn];
+    assert(
+      value !== undefined,
+      `No data for region ${region.regionId} and metric ${metric.id} found.`
+    );
+    return new MetricData(metric, region, value, /**_timeseries=*/ undefined);
```
mikelehen: nit: `/** ... */` is only for "doc comments" that should be used to generate API docs.  For normal code comments, just `/* ... */`.

That said, because the timeseries is optional, I think you can just omit it. You don't need to explicitly pass `undefined`

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+  assert(
+    rows.every((row) => typeof row[dateKey] === "string"),
+    `Not all rows have a valid date-string in column ${dateKey}.`
+  );
+  const timeseries = new Timeseries(
+    rows.map((row) => {
+      assert(
+        row[metricColumn] !== undefined,
+        `No data for region ${region.regionId} and metric ${metric.id} found.`
```
mikelehen: This error is kind of confusing.  It's okay if there's not data for a particular region+metric combo, but we expect that to be represented as `null` (I think?).  So what this is really catching is if the metric column is missing entirely.  And chances are if it's missing for one region, then it's missing for all regions (since a CSV has the same columns for every row).

So I think maybe this should be something like: `Metric column ${metricColumn}` missing.`

To be a bit more helpful we could do: `Metric column ${metricColumn} missing for row: ${row}` except unfortunately the default string representation of an array/object is just "[Object object]".  One workaround is to do `JSON.stringify(row)` to get a string representation.  The annoying part is that will generate the string even when the assert succeeds, so it's not ideal from a performance perspective.  Probably not the end of the world, but as a rule we don't want asserts to slow down the code in a meaningful way.  So I'd probably leave it out for now.

packages/metrics/src/data_providers/data_provider_utils.ts
mikelehen: Cool?  Mind actually adding a `// TODO: Merge this logic with Timeseries.isoDateString() and move to time-utils package.` comment?  Then we can do it whenever we get around to it...

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -0,0 +1,79 @@
...
+    assert(
+      value !== undefined,
+      `No data for region ${region.regionId} and metric ${metric.id} found.`
+    );
+    return new MetricData(metric, region, value, /**_timeseries=*/ undefined);
+  }
+  assert(
+    rows.every((row) => typeof row[dateKey] === "string"),
+    `Not all rows have a valid date-string in column ${dateKey}.`
+  );
```
mikelehen: Again, I think we could avoid iterating the rows twice by moving this assert into the `rows.map(...)` code below.  Basically add a `assert(row[dateKey], ...)` assert similar to the existing `row[metricColumn]` assert.

packages/metrics/src/data_providers/data_provider_utils.test.ts
```diff
@@ -0,0 +1,214 @@
...
+        newYork,
+        testMetric,
+        /*includeTimeseries=*/ false,
+        /*dateKey=*/ "date"
+      ).currentValue
+    ).toBe(null);
+  });
+});
+
+describe("dataRowsToMetricData() without timeseries", () => {
```
mikelehen: FWIW- This is another hint that this function could be split into two separate ones (one that expects timeseries data, and one that doesn't)

packages/metrics/src/data_providers/data_provider_utils.test.ts
```diff
@@ -0,0 +1,214 @@
...
+        dataRows,
+        newYork,
+        testMetric,
+        /*includeTimeseries=*/ false
+      ).currentValue
+    ).toBe(null);
+  });
+});
+
+describe("fetchCsv()", () => {
```
mikelehen: fetchCsv() is no longer in data_provider_utils.ts so I think these tests should move?

packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      /*includeTimeseries=*/ true,
+      /*dateColumn=*/ "date"
+    );
+    expect(metricDataNoTs.currentValue).toBe(150);
+    expect(metricDataNoTs.hasTimeseries()).toBe(false);
+    expect(metricDataTs.currentValue).toBe(150);
+    expect(metricDataTs.timeseries.last?.value).toBe(150);
+  });
+
+  test("fetchData() fails if timeseries is expected but does not exist in cache.", async () => {
```
mikelehen: The cache is an implementation detail and I think this test is just cehcking that fetchData() fails if timeseries is expected, but the CSV doesn't provide timeseries data (we didn't provide a `dateColumn`), right?

mikelehen (approved):
Looks good!  Thanks for all the fixes.  I did find some more tweaks to suggest, but feel free to push back / punt if you're not a fan.  They're basically all stylistic / readability stuff, I think.

packages/metrics/src/data_providers/data_provider_utils.ts
smcclure17: I see, that makes sense to me--thanks for the context surrounding it. Per the other suggestions, I ended up splitting this into two different functions (one to handle timeseries data and one to handle single datapoints), but I'll keep this in mind for the future! 🙏

packages/metrics/src/data_providers/data_provider_utils.test.ts
smcclure17: Cool, that sounds good. Good points, and again thanks for the context 🙌

packages/metrics/src/data_providers/CsvDataProvider.ts
smcclure17: Yeah, I ended up removing `includeTimeseries` from `dataRowsToMetricData()` and am relying on the `dropTimeseries()` here. Thanks for the suggestion!

packages/metrics/src/data_providers/data_provider_utils.ts
smcclure17: ah, yep sorry!

PR #65, opened 2022-08-15
Implement BarChart component

### LINKED ISSUE #62: BarChart component
Implement the BarChart component, following the model of [`LineChart`](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/components/LineChart/LineChart.tsx) and similar props (`yScale`, `xScale`). Note that the chart below also includes a line chart, ignore that for the time being.

[Figma](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=95%3A4197)

Image

pnavarrc:
Close https://github.com/covid-projections/act-now-packages/pull/62 - Implements BarChart component. 

image

**Edit**: I updated the `BarChart` API to match `LineChart` as suggested by @mikelehen, that was great feedback. I documented the rationale for the updated choices in the [Chart API Design](https://www.dropbox.com/scl/fi/snrqj20p0uoxr9phynbtu/Chart-API-Design.paper?dl=0&rlkey=3onvrnbe1753bc7v3tpey3do6#:uid=247458956670909809575882&h2=Coordinate-System-and-Scales) paper doc to better explain some adjustments that we need to make when combining line and bar charts.

I also updated the example to make it easier to see the bars/offsets

image

### DIFF
--- a/packages/ui-components/src/components/BarChart/BarChart.stories.tsx
+++ b/packages/ui-components/src/components/BarChart/BarChart.stories.tsx
@@ -23,7 +23,9 @@ const fill = "#2a9d8f";
 
 const Template: ComponentStory<typeof BarChart> = (args) => (
   <svg width={width} height={height} style={{ border: "solid 1px #eee" }}>
-    <BarChart {...args} />
+    <Group left={padding} top={padding}>
+      <BarChart {...args} />
+    </Group>
   </svg>
 );
 
@@ -61,22 +63,25 @@ Example.args = {
   timeseries,
   xScale,
   yScale,
-  fill: fill,
+  fill,
+  barWidth: dayWidth - 3,
 };
 
 export const WithLineChart = () => (
   <svg width={width} height={height} style={{ border: "solid 1px #eee" }}>
     <Group top={padding} left={padding}>
       <rect width={innerWidth} height={innerHeight} fill="#eee" />
-      <BarChart
-        timeseries={timeseries}
-        xScale={xScale}
-        yScale={yScale}
-        fill={fill}
-        fillOpacity={0.3}
-        barWidth={dayWidth - 2}
-      />
       <Group left={0.5 * dayWidth}>
+        <Group left={-0.5 * dayWidth}>
+          <BarChart
+            timeseries={timeseries}
+            xScale={xScale}
+            yScale={yScale}
+            fill={fill}
+            fillOpacity={0.3}
+            barWidth={dayWidth - 2}
+          />
+        </Group>
         {timeseries.points.map((p, i) => (
           <circle
             key={`dot-${i}`}
--- a/packages/ui-components/src/components/BarChart/BarChart.stories.tsx
+++ b/packages/ui-components/src/components/BarChart/BarChart.stories.tsx
@@ -1,13 +1,16 @@
 import React from "react";
 import { ComponentStory, ComponentMeta } from "@storybook/react";
-import { scaleLinear, scaleBand } from "@visx/scale";
+import { scaleLinear, scaleTime } from "@visx/scale";
 import { appleStock } from "@visx/mock-data";
+
 import { assert } from "@actnowcoalition/assert";
 import { Timeseries, TimeseriesPoint } from "@actnowcoalition/metrics";
+
 import BarChart from ".";
+import LineChart from "../LineChart";
 
 export default {
-  title: "Components/BarChart",
+  title: "Charts/BarChart",
   component: BarChart,
 } as ComponentMeta<typeof BarChart>;
 
@@ -30,16 +33,15 @@ const points = appleStock.map(
 
 const timeseries = new Timeseries(points).filterToDateRange({
   startAt: new Date("2008-01-01"),
-  endAt: new Date("2008-06-30"),
+  endAt: new Date("2008-03-31"),
 });
 assert(timeseries.hasData(), `Timeseries cannot be empty`);
 
-const { minValue, maxValue } = timeseries;
+const { minDate, maxDate, minValue, maxValue } = timeseries;
 
-const xScale = scaleBand({
-  domain: timeseries.dates,
+const xScale = scaleTime({
+  domain: [minDate, maxDate],
   range: [0, width],
-  padding: 0.2,
 });
 
 const yScale = scaleLinear({
@@ -52,4 +54,27 @@ Example.args = {
   timeseries,
   xScale,
   yScale,
+  fill: "#ddd",
 };
+
+const xScaleLine = scaleTime({ domain: [minDate, maxDate], range: [0, width] });
+const yScaleLine = scaleLinear({
+  domain: [minValue, maxValue],
+  range: [height, 0],
+});
+
+export const WithLineChart = () => (
+  <svg width={width} height={height}>
+    <BarChart
+      timeseries={timeseries}
+      xScale={xScale}
+      yScale={yScale}
+      fillOpacity={0.3}
+    />
+    <LineChart
+      timeseries={timeseries}
+      xScale={xScaleLine}
+      yScale={yScaleLine}
+    />
+  </svg>
+);
--- a/packages/ui-components/src/components/BarChart/BarChart.stories.tsx
+++ b/packages/ui-components/src/components/BarChart/BarChart.stories.tsx
@@ -69,30 +69,34 @@ Example.args = {
 
 export const WithLineChart = () => (
   <svg width={width} height={height} style={{ border: "solid 1px #eee" }}>
-    <Group top={padding} left={padding}>
-      <rect width={innerWidth} height={innerHeight} fill="#eee" />
-      <Group left={0.5 * dayWidth}>
-        <Group left={-0.5 * dayWidth}>
-          <BarChart
-            timeseries={timeseries}
-            xScale={xScale}
-            yScale={yScale}
-            fill={fill}
-            fillOpacity={0.3}
-            barWidth={dayWidth - 2}
-          />
-        </Group>
-        {timeseries.points.map((p, i) => (
-          <circle
-            key={`dot-${i}`}
-            cx={xScale(p.date)}
-            cy={yScale(p.value)}
-            r={4}
-            fill="black"
-          />
-        ))}
-        <LineChart timeseries={timeseries} xScale={xScale} yScale={yScale} />
+    <rect
+      y={padding}
+      x={padding}
+      width={innerWidth}
+      height={innerHeight}
+      fill="#eee"
+    />
+    <Group top={padding} left={padding + 0.5 * dayWidth}>
+      <Group left={-0.5 * dayWidth}>
+        <BarChart
+          timeseries={timeseries}
+          xScale={xScale}
+          yScale={yScale}
+          fill={fill}
+          fillOpacity={0.3}
+          barWidth={dayWidth - 2}
+        />
       </Group>
+      {timeseries.points.map((p, i) => (
+        <circle
+          key={`dot-${i}`}
+          cx={xScale(p.date)}
+          cy={yScale(p.value)}
+          r={4}
+          fill="black"
+        />
+      ))}
+      <LineChart timeseries={timeseries} xScale={xScale} yScale={yScale} />
     </Group>
   </svg>
 );
--- a/packages/ui-components/src/components/BarChart/BarChart.stories.tsx
+++ b/packages/ui-components/src/components/BarChart/BarChart.stories.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { ComponentStory, ComponentMeta } from "@storybook/react";
-import { scaleLinear, scaleTime } from "@visx/scale";
+import { scaleLinear, scaleUtc } from "@visx/scale";
+import { Group } from "@visx/group";
 import { appleStock } from "@visx/mock-data";
 
 import { assert } from "@actnowcoalition/assert";
@@ -15,8 +16,10 @@ export default {
 } as ComponentMeta<typeof BarChart>;
 
 const [width, height] = [600, 400];
-const padding = 5;
-const color = "#2a9d8f";
+const padding = 30;
+const innerWidth = width - 2 * padding;
+const innerHeight = height - 2 * padding;
+const fill = "#2a9d8f";
 
 const Template: ComponentStory<typeof BarChart> = (args) => (
   <svg width={width} height={height} style={{ border: "solid 1px #eee" }}>
@@ -28,27 +31,29 @@ const Template: ComponentStory<typeof BarChart> = (args) => (
 // so we can use them to initialize Timeseries.
 const points = appleStock.map(
   (p: { date: string; close: number }): TimeseriesPoint<number> => ({
-    date: new Date(p.date.substring(0, 10)),
+    date: new Date(`${p.date.replace(/T.*/, "")}`),
     value: p.close,
   })
 );
 
 const timeseries = new Timeseries(points).filterToDateRange({
   startAt: new Date("2008-01-01"),
-  endAt: new Date("2008-03-31"),
+  endAt: new Date("2008-01-31"),
 });
 assert(timeseries.hasData(), `Timeseries cannot be empty`);
 
 const { minDate, maxDate, minValue, maxValue } = timeseries;
 
-const xScale = scaleTime({
+const numDays = 31;
+const dayWidth = Math.floor(innerWidth / numDays);
+const xScale = scaleUtc({
   domain: [minDate, maxDate],
-  range: [padding, width - 2 * padding],
+  range: [0, innerWidth - dayWidth],
 });
 
 const yScale = scaleLinear({
   domain: [minValue, maxValue],
-  range: [0, height],
+  range: [innerHeight, 0],
 });
 
 export const Example = Template.bind({});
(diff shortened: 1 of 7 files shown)


packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { ScaleLinear, ScaleTime } from "d3-scale";
+import { Group } from "@visx/group";
+
+import { Timeseries } from "@actnowcoalition/metrics";
+
+export interface BarChartOwnProps {
+  /** Timeseries used to draw the line chart */
+  timeseries: Timeseries<number>;
+
+  /** Scale to transform point dates to pixel positions on the x-axis */
```
mikelehen: nit: Did we go with d3-scale or similar for LineChart?

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,99 @@
...
+  xScale: ScaleTime<number, number>;
+
+  /** Scale to transform point values to piel position on the y-axis */
```
mikelehen: piel => pixel

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,99 @@
...
+}
+
+export type BarChartProps = BarChartOwnProps & React.SVGProps<SVGRectElement>;
+
+/**
+ * BarChart is a chart to represent a Timeseries as a series of bars, one bar
+ * for each point in the series.
+ *
+ * The horizontal center of the rectangles will be aligned with the date of its
+ * corresponding point, make sure to adjust the chart paddings accordingly.
```
mikelehen: "point, make" => "point. Make" ?

But also I'm not 100% sure what this means.  That I need to have some padding so that the first/last bars don't extend outside the chart bounds?  Maybe make the comment a bit more explicit.

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * By default, the bar width corresponds to the distance between two
+ * consecutive days.
+ *
+ * @example
+ * ```tsx
```
mikelehen, line 33: Oooh, nice.  Didn't know you could specify tsx formatting.

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,99 @@
...
+ * const xScale = scaleTime({ domain: [minDate, maxDate], range: [0, width] });
+ * const yScale = scaleLinear({ domain: [minValue, maxValue], range: [0, height] });
```
mikelehen: Will this end up with an "upside-down" chart?  Like you'd probably want `[height, 0]` instead?

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * BarChart is a chart to represent a Timeseries as a series of bars, one bar
+ * for each point in the series.
+ *
+ * The horizontal center of the rectangles will be aligned with the date of its
+ * corresponding point, make sure to adjust the chart paddings accordingly.
+ *
+ * By default, the bar width corresponds to the distance between two
+ * consecutive days.
```
mikelehen: Consider moving this to the `barWidth` documentation on `BarChartProps` so it's closer to where it is used.

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,99 @@
+import React from "react";
+import { ScaleLinear, ScaleTime } from "d3-scale";
+import { Group } from "@visx/group";
+
+import { Timeseries } from "@actnowcoalition/metrics";
+
+export interface BarChartOwnProps {
```
mikelehen, line 7: I think we need to export this from the module index.ts too...
```
./packages/ui-components/src/components/BarChart/BarChart.tsx:7:0 - warning BarChartOwnProps is referenced by @actnowcoalition/ui-components.BarChartProps but not included in the documentation.

7    export interface BarChartOwnProps {
```

Right now we're not getting any useful docs for the props:

image


Same for LineChartOwnProps:

```
./packages/ui-components/src/components/LineChart/LineChart.tsx:8:0 - warning LineChartOwnProps is referenced by @actnowcoalition/ui-components.LineChartProps but not included in the documentation.

8    export interface LineChartOwnProps {
```

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,99 @@
...
+ *     xScale={xScale}
+ *     yScale={yScale}
+ *   />
+ * </svg>
+ * ```
+ *
+ * @param timeseries Timeseries to represent as a bar chart.
+ * @param xScale d3-scale to transform dates into pixel positions on the x-axis
+ * @param yScale d3-scale to transform values into pixel heights of each bar.
+ * @param barWidth width (in pixels) of each bar.
```
mikelehen, line 49: I'm not sure these `@param` docs are being picked up/used:

image

I am not sure if there's a way to fix that or if we should just document `BarChartProps` well and rely on that.

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,99 @@
...
+    </Group>
+  );
+};
+
+export default BarChart;
+
+// Given a date, it returns the date of the next day
+function nextDay(date: Date): Date {
+  const unixTimeMilliseconds = date.getTime();
+  return new Date(unixTimeMilliseconds + 1000 * 24 * 60 * 60);
```
mikelehen: Note that this does not reflect daylight saving time, which may or may not be desired depending on the usage of `nextDay()`.  For this reason I'd slightly prefer not have this `nextDay()` function. Basically as much as possible it would be nice to avoid doing date math and when we have to, centralize it in one package.

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,99 @@
...
+}) => {
+  // We need minDate to calculate the bar width, so we return early if the
+  // timeseries is empty.
+  if (!timeseries.hasData()) {
+    return null;
+  }
+
+  const { minDate } = timeseries;
+  const rectWidth =
+    barWidth ?? Math.floor(xScale(nextDay(minDate)) - xScale(minDate));
```
mikelehen: FWIW-  I don't think `minDate` or `nextDay()` are exactly needed here, since we're basically just wanting to get the a bar width corresponding to 1 day (24 hours).  So we could do something like:  `xScale(new Date(24*60*60*1000)) - xScale(new Date(0))`

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const [, maxHeight] = yScale.range();
```
mikelehen: nit: I am not 100% sure what `maxHeight` means.  Would just `height` or `chartHeight` make sense?

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  return (
+    <Group left={-0.5 * rectWidth}>
+      {timeseries.points.map((p, i) => {
+        const barHeight = yScale(p.value);
```
mikelehen: Is this assuming that yScale(0) == 0 ?  I think maybe technically this should be `barTop` and we should maybe be calling `yScale(0)` somewhere to get `barBottom` and then using those to draw the bar?  Though FWIW that also assumes that the bars should extend down to "0" in the data coordinate system, which wouldn't work if you had some sort of discontinuous axis (e.g. your y-axis started at 100 instead of 0).  Or am I confused / overthinking this?

I think as-is the code would break if your y range doesn't start at 0, e.g.: 
```
yScale = scaleLinear({
  domain: [minValue, maxValue],
  range: [50, height],
});
```

Though I'm not sure if it ever makes sense to do that.

packages/ui-components/src/components/LineChart/LineChart.tsx
```diff
@@ -29,21 +29,23 @@ export type LineChartProps = LineChartOwnProps &
  * Storybook for a working example.
  *
  * @example
- *    const xScale = scaleLinear({ domain: [minDate, maxDate], range: [0, 200] });
- *    const yScale = scaleLinear({ domain: [minVal, maxVal], range: [0, 100] });
+ * ```tsx
+ * const xScale = scaleTime({ domain: [minDate, maxDate], range: [0, 200] });
+ * const yScale = scaleLinear({ domain: [minVal, maxVal], range: [0, 100] });
```
mikelehen: Per my comment above, will this end up being an upside-down chart?  You'd want `[100, 0]` ?

mikelehen (changes requested):
This basically looks good, but I think there may be a slight bug / weakness with calculating the bottom/tops of bars based on the `yScale`.  Though I could also be confused.

packages/ui-components/src/components/BarChart/BarChart.tsx
pnavarrc: I debated myself a lot over the scale orientation, I think it deserves some standup discussion, I'm going to draw some diagrams to discuss it with the team (also for the bar width offset mentioned above). Thanks for your notes! 🙏

mikelehen:
FYI- When I check out your branch and run `yarn` I end up with diffs to `yarn.lock`.  You might want to re-run it and add the changes to the PR?

packages/ui-components/src/components/BarChart/BarChart.tsx
pnavarrc: That was in the component docs (not the interface), but I think it's better to be more explicit in both the interfaces and component docs, updating LineChart as well 👍

packages/ui-components/src/components/BarChart/BarChart.tsx
pnavarrc: Updated!

packages/ui-components/src/components/BarChart/BarChart.tsx
pnavarrc: Updated to match LineChart!

packages/ui-components/src/components/BarChart/BarChart.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+    />
+    <Group top={padding} left={padding + 0.5 * dayWidth}>
+      <Group left={-0.5 * dayWidth}>
+        <BarChart
+          timeseries={timeseries}
+          xScale={xScale}
+          yScale={yScale}
+          fill={fill}
+          fillOpacity={0.3}
+          barWidth={dayWidth - 2}
```
mikelehen: Any particular reason we use `- 3` above and `- 2` here?

Feels like maybe we should have a `barPadding` constant for consistency / readability. 🤷

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,91 @@
+import React from "react";
+import { ScaleLinear, ScaleTime } from "d3-scale";
+import { Group } from "@visx/group";
+
+import { Timeseries } from "@actnowcoalition/metrics";
+
+export interface BarChartOwnProps {
+  /** Timeseries used to draw the line chart */
```
mikelehen: line => bar

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,91 @@
...
+  timeseries: Timeseries<number>;
+
+  /** d3-scale to transform point dates to pixel positions on the x-axis */
+  xScale: ScaleTime<number, number>;
+
+  /** d3-scale to transform point values to pixel position on the y-axis */
```
mikelehen: position => positions

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,91 @@
...
+  yScale: ScaleLinear<number, number>;
+
+  /** Width of each bar, in pixels. By default, the barWidth will be set to
+   * the distance between 2 consecutive days.
```
mikelehen: nit: I think the typical convention is that when a `/** ... */` goes to multiple lines, you leave the first / last lines blank, so:

```
/**
 * Width of each bar, in pixels. By default, the barWidth will be set to
 * the distance between 2 consecutive days.
 */
```

mikelehen: Also, it looks like the default is actually `2` now.  I'm curious why you changed it, but in any case we should update the doc comment to match!

packages/ui-components/src/components/BarChart/BarChart.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+  <svg width={width} height={height} style={{ border: "solid 1px #eee" }}>
+    <rect
+      y={padding}
+      x={padding}
+      width={innerWidth}
+      height={innerHeight}
+      fill="#eee"
+    />
+    <Group top={padding} left={padding + 0.5 * dayWidth}>
+      <Group left={-0.5 * dayWidth}>
```
mikelehen, line 81: Hrm.  This seems kind of funky.  I guess I'm not 100% sure why we changed it to left-align the bars instead of center-aligning.  Was that in response to my feedback?  I may be missing something.

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,91 @@
...
+ */
+const BarChart: React.FC<BarChartProps> = ({
+  timeseries,
+  xScale,
+  yScale,
+  barWidth = 2,
+  ...rectProps
+}) => {
+  // We need minDate to calculate the bar width, so we return early if the
+  // timeseries is empty.
```
mikelehen: I think this is no longer true?

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,91 @@
...
+    <Group>
+      {timeseries.points.map((p, i) => {
+        const rectY = yScale(p.value);
+        return (
+          <rect
+            key={`bar-${i}`}
+            x={xScale(p.date)}
+            y={rectY}
+            width={barWidth}
+            height={maxHeight - rectY}
```
mikelehen: This doesn't feel right to me.  In theory you should be able to set the yScale range to `[0, innerHeight]` instead of `[innerHeight, 0]` and you'll just end up with an upside-down version of the same chart but instead you get sort of a right-side up chart with "inverted values" (small bars end up big bars and vice versa):

`[innerHeight, 0]`
image


`[0, innerHeight]`
image

In theory I think it should be:
```
y={yStart}
height={rectY - yStart}
```

But it looks like SVG doesn't like negative heights so instead you probably need:

```
y={Math.min(yStart, rectY)}
height={Math.abs(rectY - yStart)}
```

And then you correctly get the upside-down chart:
image

packages/ui-components/src/components/LineChart/LineChart.tsx
```diff
@@ -9,10 +9,10 @@ export interface LineChartOwnProps {
   /** Timeseries used to draw the line chart */
   timeseries: Timeseries<number>;
 
-  /** Scale to transform point dates to pixel positions on the x-axis */
+  /** d3-scale to transform point dates to pixel positions on the x-axis */
   xScale: ScaleTime<number, number>;
 
-  /** Scale to transform point values to piel position on the y-axis */
+  /** d3-scale to transform point values to piel position on the y-axis */
```
mikelehen: nit: position => positions

packages/ui-components/src/components/LineChart/LineChart.tsx
mikelehen: ping.  Should this be `[100, 0]` to be a more useful example?

mikelehen (approved):
LGTM though I think there's still technically a bug with the y coordinate calculation, and I'm not 100% sure on the left-aligned vs. center-aligned change.

packages/ui-components/src/components/BarChart/BarChart.tsx
pnavarrc: typo, should be 1!

pnavarrc: I'm leaning towards a simpler default because the parent component will need to be aware and factor in the `barWidth` when defining the scales and to adjust the position of other charting elements, so I set it to be `2px` just so the bars are visible by default. Updating the docs!

packages/ui-components/src/components/BarChart/BarChart.stories.tsx
pnavarrc: After integrating the `LineChart`, I figured that the amount of offset that we want will depend on the context (whether or not the bar chart is composed with other charts or not, etc), so I think it's better and more explicit to leave that concern to the user (since they will be passing the `barWidth` as parameter anyways)

packages/ui-components/src/components/BarChart/BarChart.tsx
pnavarrc: Oh this is much better! 👏 👏 👏

PR #66, opened 2022-08-15
Update the threshold styling to use theme

pnavarrc:
Updated the text and ticks mark colors to use variables from the theme. The typography is not being picked up so I'm leaving that as a TODO item for now.

### DIFF
--- a/packages/ui-components/src/index.ts
+++ b/packages/ui-components/src/index.ts
@@ -41,14 +41,17 @@ declare module "@mui/material/styles" {
   interface DefaultTheme extends Theme {}
 
   interface Palette {
-    severity: {
+    border: {
+      default: string;
+    };
+    gradient: {
       100: string;
       200: string;
       300: string;
       400: string;
       500: string;
     };
-    gradient: {
+    severity: {
       100: string;
       200: string;
       300: string;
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/index.ts
```diff
@@ -41,14 +41,17 @@ declare module "@mui/material/styles" {
   interface DefaultTheme extends Theme {}
 
   interface Palette {
-    severity: {
+    border: {
```
mikelehen, line 44: Out of curiosity, did "border" come from Josh's design system?  (where's the figma link?). Else should we tell him we added it.  I'm curious how we keep our code in sync with his figma design system so that we're using the same tokens / building blocks.

mikelehen (approved):
Looks good to me!

packages/ui-components/src/index.ts
pnavarrc: Good point, I will reference Figma in the future :) - I think there is a Figma plugin to integrate Storybook (into Figma), but I'm not sure if there is an automated way to sync/update both. Chelsi picked up colors from the [Act Now Design System](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=29%3A1035)

PR #68, opened 2022-08-15
Add labels to vertical legend thresholds

faihegberg:
Add labels and sub labels to vertical legend thresholds.

### DIFF
--- a/packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx
+++ b/packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx
@@ -10,6 +10,7 @@ export default {
 
 interface Item {
   label: string;
+  sublabel?: string;
   color: string;
 }
 
@@ -22,13 +23,7 @@ const horizontalHeight = 40;
 const horizontalBarHeight = 20;
 const horizontalWidth = 300;
 
-// Vertical legend threshold props
-const verticalWidth = 20;
-const verticalHeight = 300;
-
-const borderRadius = 4;
-
-const items: Item[] = [
+const horizontalItems: Item[] = [
   { label: "10", color: "#90BE6D" },
   { label: "20", color: "#F9C74F" },
   { label: "30", color: "#F8961E" },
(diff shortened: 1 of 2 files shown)


packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx
```diff
@@ -97,3 +111,12 @@ VerticalSquared.args = {
...
   borderRadius: 0,
 };
+
+export const VerticalNoLabel = Template.bind({});
+VerticalNoLabel.args = {
+  ...VerticalDefault.args,
+  width: 72 / 5,
+  height: 72,
+  borderRadius: 8,
+  showLabels: false,
```
chasulin: do we need a `showLabels` prop? vs. rendering labels/sublabels if theyre provided, and not rendering them if theyre not

chasulin (commented):
the sublabel styles don't seem too match whats in the design system in [figma](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=6%3A6). also, using `rect`/`text` leads the text to not wrap or be responsive 😕. thoughts on using divs and typography?

packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx
faihegberg: I added `showLabels` as an optional prop. I think it would be good to keep the prop and explicitly label it `true` if labels are provided _and_ we want to display them. For example, if we supply an `Item[]` to the component and each `Item` happens to have `labels/sublabels` prop, but we don't want to display labels/sublabels, declaring `showLabels` as false would be straightforward.

I will adjust `showLabels` to default to `false` though, instead of marking it as just optional.

faihegberg:
> the sublabel styles don't seem too match whats in the design system in [figma](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=6%3A6). also, using `rect`/`text` leads the text to not wrap or be responsive 😕. thoughts on using divs and typography?

For the latter point, it might be tricky cos this is nested inside a SVG. I looked into this a little bit (https://stackoverflow.com/questions/17595813/is-it-possible-to-append-a-div-inside-an-svg-element and https://bl.ocks.org/Jverma/2385cb7794d18c51e3ab) and I can try playing around with using `<foreignobject>` to wrap the `div` in and see how it plays out...

Not sure about `typography`, I'll try that...

faihegberg:
Closing as this will be re-implemented without using SVG instead.

PR #76, opened 2022-08-16
Add typography variants to the theme

### LINKED ISSUE #75: Add typography variants
We need to add typography variants to the packages theme (see `paragraphSmall` and `paragraphLarge` for a reference implementation)

- Update the `ExtendedTypographyOptions` interface (https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/styles/theme/interfaces.ts)
- Update the [`typography`](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/styles/theme/typography.ts) to include the typography variants shown in Figma.

[Act Now Design System (Figma)](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=285%3A20188)

pnavarrc:
Close #75 - Implement typography variants (theme)

image

## Changes

- Updated the font-family to `Inter` for the default theme, and added `Source Code Pro` as monospaced font.
- Added the design system variants to Typography
- Added a story to show Typography variants
- Updated the Storybook config to sort stories, and added the new fonts to the `preview-head.html` file.
- Export the theme with `responsiveFontSizes`, to automatically adjust font sizes under each breakpoint.
- Added styles for `h4` - `h6` (same as `h3`) so we have a good fallback in case we need them

## Notes

- The design system doesn't include specific breakpoints, it only has **mobile**, **tablet** and **desktop**. We need to check with Josh what's the pixel value for these breakpoints.
- The font size for headings is modified based on breakpoints on the design system, I just set the defaults for desktop and added `responsiveFontSizes` to scale them automatically. We can sync with Josh later to confirm whether that's good enough (or adjust as needed).

### DIFF
--- a/packages/ui-components/src/styles/theme/typography.ts
+++ b/packages/ui-components/src/styles/theme/typography.ts
@@ -25,6 +25,9 @@ const typographyConstants = {
   fontSizeBase: "1rem",
   fontSizeDSmall: "1.125rem",
   fontSizeDLarge: "1.5rem",
+  fontSizeH1: "2.75rem",
+  fontSizeH2: "2rem",
+  fontSizeH3: "1.125rem",
 
   lineHeightSmall: 1,
   lineHeightMedium: 1.25,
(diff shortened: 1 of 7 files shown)


packages/ui-components/src/styles/theme/typography.ts
```diff
@@ -1,34 +1,146 @@
...
   lineHeightSmall: 1,
+  lineHeightMedium: 1.25,
   lineHeightBase: 1.5,
 };
 
 const typography: ExtendedTypographyOptions = {
   fontFamily: typographyConstants.fontFamily,
 
+  h1: {
+    fontSize: "2.75rem",
```
chasulin: why not in `typographyConstants`?

chasulin (approved):

PR #77, opened 2022-08-16
Add boolean helper methods to `MetricData` and `Timeseries` classes

smcclure17:


packages/metrics/src/Timeseries/Timeseries.test.ts
```diff
@@ -177,4 +177,32 @@ describe("Timeseries", () => {
...
+    const verifyValueThrowsError = (value: unknown) => {
+      expect(() =>
+        new Timeseries([
+          { date: new Date("2021-02-03"), value },
+        ]).assertBoolean()
+      ).toThrowError("Found non-boolean value in timeseries");
+    };
+    verifyValueThrowsError("false");
+    verifyValueThrowsError("true");
+    verifyValueThrowsError("string");
```
faihegberg: I think one string of either "true" or "false" is probably sufficient.

packages/metrics/src/data/MetricData.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    expect(() => {
+      buildMetricData("not boolean").assertBoolean();
+    }).toThrow();
+  });
+
+  test("convertToBoolean() succeeds on near-boolean data.", () => {
+    expect(buildMetricData("yes").convertToBoolean()).toStrictEqual(
+      buildMetricData(true)
+    );
+  });
```
faihegberg: Not really sure what this means by "near-boolean" data?

faihegberg: Nvm, just saw the comment in `data_utils.ts`. Wonder if there's a better term than "near-boolean" though.

packages/metrics/src/data/data_utils.test.ts
```diff
@@ -0,0 +1,32 @@
+import { parseBoolean } from "./data_utils";
+
+describe("data_utils", () => {
+  test("parseBoolean accepts valid boolean-y values", () => {
+    expect(parseBoolean("true")).toBe(true);
+    expect(parseBoolean("True")).toBe(true);
+    expect(parseBoolean("yes")).toBe(true);
+    expect(parseBoolean("Yes")).toBe(true);
+    expect(parseBoolean(1)).toBe(true);
```
faihegberg, line 9: Perhaps add "1" string as a test case?

faihegberg (approved):

packages/metrics/src/data/MetricData.test.ts
smcclure17: Yeah, I agree it is a little unclear--I'll just replace this with "valid data" here. I think I feel okay leaving it in the other instances because they're followed by examples and I can't think of a better term atm, but agreed it's not very clear.

PR #78, opened 2022-08-16
Add string helper methods to `MetricData` and `Timeseries` classes

smcclure17:


packages/metrics/src/data/MetricData.ts
```diff
@@ -88,6 +88,49 @@ export class MetricData<T = unknown> {
...
+
+  /**
+   * Converts metric data to strings.
+   *
+   * Null values are preserved and are not converted to string.
+   *
+   * @returns This `MetricData` with its data converted to strings, and cast to `MetricData<string>`
+   */
+  convertToString(): MetricData<string> {
+    const toStringPreserveNull = (value: unknown) =>
```
smcclure17: IDK how strict we want to be about what can/cannot be converted to `string`.

 This is pretty laissez-faire, allowing anything to be converted as long as the `String()` call allows it, which I could see causing issues (if we try to convert something that doesn't make sense, e.g. an object). But I also didn't want to restrict functionality too much.

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -206,6 +206,26 @@ export class Timeseries<T = unknown> {
     return this.cast<number>();
   }
 
+  /**
+   * Throws an exception if any values in the timeseries are not of type string.
+   *
+   * Null values are treated as strings and will not throw an error.
```
mikelehen: I think null values should be rejected from the Timeseries (this is what `assertFiniteNumbers()` does).  Otherwise, our function is lying by saying it returns `Timeseries<string>`.  It should really be `Timeseries<string|null>`.

FWIW- In general I'm expecting common usage to be something like `timeseries.removeNils().assertStrings()`

packages/metrics/src/data/MetricData.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  return new MetricData(
+    testMetric,
+    testRegion,
+    value,
+    Timeseries.fromJSON([{ date: "2022-08-01", value: value }])
+  );
+};
+
+describe("MetricData", () => {
+  test("assertStrings() succeeds on string or null data.", () => {
```
mikelehen: As mentioned, I think it should fail if there's a null in the timeseries.

packages/metrics/src/data/MetricData.ts
mikelehen: Eh, TBH I can't think of a use case offhand for convertToString() and would be inclined to leave it out until / unless we need it.

packages/metrics/src/data/MetricData.ts
```diff
@@ -88,6 +88,49 @@ export class MetricData<T = unknown> {
...
+      value === null ? (null as unknown as string) : String(value);
+    const currentValueStr = toStringPreserveNull(this.currentValue);
+    const booleanTimeseries = this._timeseries?.mapValues(toStringPreserveNull);
```
mikelehen: boolean? :-)

packages/metrics/src/data/MetricData.ts
```diff
@@ -88,6 +88,49 @@ export class MetricData<T = unknown> {
...
+  /**
+   * Ensures the metric data is of type string.
+   *
+   * Null values are treated as strings and will not throw an error.
+   *
+   * @returns This `MetricData` cast to `MetricData<string>`.
+   */
+  assertStrings(): MetricData<string> {
+    assert(
+      this.currentValue === null || typeof this.currentValue === "string",
```
mikelehen: Note that *here* we can/should allow `null` because `currentValue` can always be missing.  I know this is kind of weird / confusing.  If you have ideas for how to make this make more sense or want to talk through it, let me know.

mikelehen (approved):
LGTM with some minor nits!

packages/metrics/src/data/MetricData.ts
smcclure17: Yeah, that sgtm--I've gone ahead and removed it!

packages/metrics/src/data/MetricData.ts
smcclure17: oops, I definitely wasn't copy/pasting code 😆

packages/metrics/src/Timeseries/Timeseries.ts
smcclure17: Okay, yeah that makes sense! The dissymmetry with `MetricData.assertStrings()` is a little confusing, but the underlying reason for it makes sense and I can't think of a better approach 🙌

packages/metrics/src/data/MetricData.ts
smcclure17: Cool, that reasoning makes sense! Thanks for the explanations 🙏 No thoughts at the moment, but will try to give it some thought as we work with this further

PR #80, opened 2022-08-17
Add theme colors to Storybook

pnavarrc:
Add theme colors to Storybook.

image

I just picked up colors automatically from the theme, and excluded `mode` which seems to be an internal MUI thing.

### DIFF
--- a/packages/ui-components/src/stories/Colors.stories.tsx
+++ b/packages/ui-components/src/stories/Colors.stories.tsx
@@ -1,19 +1,24 @@
 import React from "react";
-import { Typography, Grid, Box } from "@mui/material";
+import isObject from "lodash/isObject";
+import { Typography, Palette, Grid, Box } from "@mui/material";
 import theme from "../styles/theme";
 
 export default {
   title: "Design System/Colors",
 };
 
-const groups = Object.keys(theme.palette)
-  .filter((name) => typeof theme.palette[name] === "object")
-  .filter((name) => !["mode"].includes(name))
+const paletteGroups = (Object.keys(theme.palette) as (keyof Palette)[])
+  .filter((name) => isObject(theme.palette[name]))
+  .filter((name) => ["mode"].includes(name))
   .map((name) => {
-    const colors = Object.keys(theme.palette[name]).map((colorKey) => ({
-      name: `theme.palette.${name}.${colorKey}`,
-      color: theme.palette[name][colorKey],
-    }));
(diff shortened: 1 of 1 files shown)


packages/ui-components/src/stories/Colors.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { Typography, Palette, Grid, Box } from "@mui/material";
+import theme from "../styles/theme";
+
+export default {
+  title: "Design System/Colors",
+};
+
+const paletteGroups = (Object.keys(theme.palette) as (keyof Palette)[])
+  .filter((name) => isObject(theme.palette[name]))
+  .filter((name) => ["mode"].includes(name))
```
mikelehen: Is this equivalent to `name === 'mode'`? Or am I confused?  If so, then we don't need any of this `Object.keys(...).filter(...)` stuff, right?  We could just basically do `const name = 'mode'` ?  Feel like I may be confused. 😁

mikelehen (changes requested):
Hrm.  I'm not sure if this is working for me.   https://act-now-packages--pr80-pablo-colors-to-stor-hp3f4gs8.web.app/storybook/index.html?path=/story/design-system-colors--all is not rendering anything.  Am I confused?

packages/ui-components/src/stories/Colors.stories.tsx
pnavarrc: That was a bug 😅  — Some of the palette attributes are functions or non-string constants (such as `getContrastText()` and `contrastThreshold`), so I'm filtering out those. The `mode` attribute is used to name themes (`light`, `dark`), so is also not a color, so I'm filtering it out.

pnavarrc: Adding a comment to explain that 👍

packages/ui-components/src/stories/Colors.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+import theme from "../styles/theme";
+
+export default {
+  title: "Design System/Colors",
+};
+
+// Filter out the palette attributes that don't correspond to colors,
+// such as functions or numeric constants. The `mode` attribute is
+// a string to name themes (dark, light, highContrast), but not a color,
+// so we filter that out as well.
```
mikelehen, line 13: Very helpful comment, thanks!

mikelehen (approved):
Looks good!

Issue #63, opened 2022-08-15
Implement chart Axes


pnavarrc:
Implement `AxisLeft` and `AxisBottom`, using the API from `@visx/axis`, but adapting lines and typography to use our theme variables for colors. Don't try to make the date format match the mocks yet - we probably need to think more about that (when to abbreviate month names, when to put the year, what to do on mobile, etc.)

[Figma](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=72%3A2892)

![Image](https://user-images.githubusercontent.com/114084/184691082-ee03e81d-20a7-4739-9abb-1080a1310137.png)

PR #81, opened 2022-08-17
Create axes

faihegberg:
Create horizontal and vertical axes for charts, as per [task](https://github.com/covid-projections/act-now-packages/issues/63).

Notes:
- The date format isn't determined yet (as mentioned in the task ^). So I'm just using an arbitrary date format similar to the [mocks](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=72%3A2892).
- Currently in the mocks, the first label on the x-axis isn't displayed. I managed to configure our labels to hide the first label to match the mocks, but still haven't found a way to hide _only_ the first tick mark. The [docs](https://airbnb.io/visx/docs/axis) explain how one can either show all or hide all tick marks, but not individual ones. Also tried playing around with css selectors but no success. Any ideas are welcome! 😄 
- In storybook, it might look like the `body` component has styling (e.g. font color) that doesn't match the mocks, but this is actually getting overridden by the `text` component styling, which should match the mocks.
- There is currently a merge conflict but currently if I merge in develop, I will run into [this issue](https://actnowcoalition.slack.com/archives/C01JP05V28G/p1661374844128729). So I'm holding off on resolving the conflict for now.


packages/ui-components/src/components/Axis/Axis.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+  domain: [0, 500],
+  range: [height - 2 * padding, 0],
+});
+
+const bottomScale = scaleTime({
+  domain: [new Date("2022-01-01"), new Date("2022-06-01")],
+  range: [0, width - 2 * padding],
+});
+
+const formatDate = (date: Date) => formatDateTime(date, DateFormat.MMM);
```
faihegberg, line 30: Arbitrary formatting until we decide date format for charts.

packages/ui-components/src/components/Axis/Axis.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+    <AxisBottom
+      {...args}
+      left={padding}
+      top={padding}
+      // Approximate (More info: https://airbnb.io/visx/docs/axis#Axis_numTicks)
+      numTicks={5}
+      tickLength={4}
+      tickStroke={theme.palette.border.default}
+      tickFormat={formatDate}
+      tickLabelProps={bottomTickLabelProps}
```
faihegberg: `<AxisBottom>` has built-in `tickLabelProps` which is pretty helpful for styling. It looks like before we were overriding styles using css selectors, but it seems this prop allows you to specify styles directly as part of the `AxisBottom` component.

packages/ui-components/src/components/Axis/Axis.style.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export const leftTickLabelProps: TickLabelProps<{}> = () => ({
+  textAnchor: "end", // Horizontal anchor
+  verticalAnchor: "middle",
+  ...baseTickLabelProps,
+});
+
+export const AxisLeft = styled((props: AxisLeftProps) => (
+  <VxAxisLeft axisClassName={props.className ?? ""} {...props} />
```
faihegberg: Still not 100% sure if and why we need class name here...
I based this off of our axis implementation in the GCAT repo, but I'll nix it if it turns out we don't need this.

pnavarrc: Consider passing all the defaults directly so users don't need to pass all the props (customization should be more of an exception than then rule):

```tsx
export const AxisLeft = styled((props: AxisLeftProps) => (
  <VxAxisLeft 
    axisClassName={props.className ?? ""} 
    tickLength={4}
    tickStroke={theme.palette.border.default}
    tickFormat={formatDate}
    tickLabelProps={() => ({ textAnchor: "end", verticalAnchor: "middle" })}
    {...props} 
  />
```

This way, it will follow the design system by default and allow us override props if we need to.

packages/ui-components/src/components/Axis/Axis.style.tsx
```diff
@@ -0,0 +1,42 @@
...
+  className?: string;
+};
+
+export type AxisBottomProps = React.ComponentProps<typeof VxAxisBottom> & {
+  className?: string;
+};
+
+const baseTickLabelProps = {
+  fill: theme.palette.text.secondary,
+  fontSize: theme.typography.body2.fontSize,
```
pnavarrc: You should use the typography variables from `paragraphSmall`, and remember to add all of them (font-family, font-size, font-weight, etc).

packages/ui-components/src/components/Axis/Axis.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+});
+
+const bottomScale = scaleTime({
+  domain: [new Date("2022-01-01"), new Date("2022-06-01")],
+  range: [0, width - 2 * padding],
+});
+
+const formatDate = (date: Date) => formatDateTime(date, DateFormat.MMM);
+
+const TemplateBottomAxis: ComponentStory<typeof AxisBottom> = (args) => (
```
pnavarrc: Consider adding an example without custom styling, so we can confirm that the default match the mocks

pnavarrc (commented):
Looking good, but left a few notes, ping me to do a second pass!

packages/ui-components/src/components/Axis/Axis.stories.tsx
```diff
@@ -0,0 +1,87 @@
+import React from "react";
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { scaleLinear, scaleTime } from "@visx/scale";
+import { AxisLeft, AxisBottom } from "./Axis.style";
+
+export default {
+  title: "Charts/Axis",
+  component: AxisLeft,
+} as ComponentMeta<typeof AxisLeft>;
```
pnavarrc, line 9: The `component` attribute is optional, consider omiting it if you want to have both axes in the same story file

```tsx
export default {
  title: "Charts/Axis"
}
```

packages/ui-components/src/components/Axis/Axis.style.tsx
```diff
@@ -0,0 +1,63 @@
...
+  fontSize: typography.paragraphSmall.fontSize,
+  fontWeight: typography.paragraphSmall.fontWeight,
+};
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export const bottomTickLabelProps: TickLabelProps<{}> = () => ({
+  textAnchor: "middle", // Horizontal anchor
+  verticalAnchor: "start",
+  ...baseTickLabelProps,
+});
```
pnavarrc: We don't need to export `bottomTickLabelProps`, also, consider simplifying a little and trying to preserve the type signature of the function

```tsx
const bottomTickLabelProps = <T,>(props: TickLabelProps<T>) => ({
  textAnchor: "middle",
  verticalAnchor: "start",
  ...typography.paragraphSmall
})
```

(not sure if that will work as is, you might need to adjust that a little)

packages/ui-components/src/components/Axis/Axis.style.tsx
```diff
@@ -0,0 +1,63 @@
...
+    {...props}
+  />
+))``;
+
+export const AxisBottom = styled((props: AxisBottomProps) => (
+  <VxAxisBottom
+    axisClassName={props.className ?? ""}
+    tickLength={4}
+    tickStroke={theme.palette.border.default}
+    tickFormat={props.tickFormat ?? formatDate}
```
pnavarrc: It might be better to leave the default value for `tickFormat`, that way the calculation of the ticks will be better and will automatically adapt the types depending on the type of the scale. For example, with a numeric linear scale, you won't need to pass `tickFormat`

```tsx
const timeScale = scaleUtc({ 
    domain: [new Date('2000-01-01'), new Date('2020-12-31')], 
    range: [0, 500]
})

// will show years, since the domain extends for 20 years
<AxisBottom scale={timeScale}>

const xScale = scaleLinear({ domain: [0, 1000], range: [0, 500] })

// will format the tick labels correctly without passing a custom tickFormat function
<AxisBottom scale={xScale}>
```

pnavarrc (approved):
Added some final notes, but looks good!

packages/ui-components/src/components/Axis/Axis.style.tsx
faihegberg: As discussed on slack, will implement as...
```
const baseTickLabelProps = {
  fill: typography.paragraphSmall.color,
  fontFamily: typography.paragraphSmall.fontFamily,
  fontSize: typography.paragraphSmall.fontSize,
  fontWeight: typography.paragraphSmall.fontWeight,
};
```
then
```
  <VxAxisBottom
    ...
    tickLabelProps={() => ({
      textAnchor: "middle", // Horizontal anchor
      verticalAnchor: "start",
      ...baseTickLabelProps,
    })}
  />
```

PR #82, opened 2022-08-18
Fix useData() hooks and add testing.

mikelehen:
The existing useData() hooks didn't gracefully handle passing the same metrics / regions on every render if the actual array instances changed, e.g.: This would cause an infinite re-render:

```
catalog.useDataForRegionsAndMetrics([region], [metric])
```

Because the array instances would be differente every time.  The core fix is to add a `useCachedArrayIfEqual()` helper that preserves the original array reference if the contents are the same.

This PR also:
* Adds a test to validate the hook behaves correctly now.
* Enables `plugin:react-hooks/recommended` eslint checks.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,6 +7,7 @@
   "extends": [
     "eslint:recommended",
     "plugin:react/recommended",
+    "plugin:react-hooks/recommended",
     "plugin:@typescript-eslint/recommended"
   ],
   "parser": "@typescript-eslint/parser",
--- a/packages/metrics/src/MetricCatalog/MetricCatalog.ts
+++ b/packages/metrics/src/MetricCatalog/MetricCatalog.ts
@@ -1,5 +1,6 @@
-import { useState, useEffect } from "react";
+import { useState, useEffect, useRef } from "react";
 import groupBy from "lodash/groupBy";
+import isEqual from "lodash/isEqual";
 import keyBy from "lodash/keyBy";
 import uniq from "lodash/uniq";
 
@@ -32,6 +33,15 @@ export class MetricCatalog {
   readonly options: MetricCatalogOptions;
 
   private readonly metricsById: { [id: string]: Metric };
+
+  /**
+   * Count of times that the MetricCatalog has had to fetch data from one or
+   * more data providers (e.g. any of the fetchData() methods were called, or a
+   * useData() hook was called with new metrics / regions).
+   *
+   * Used for debugging / testing.
+   */
+  dataFetchesCount = 0;
 
   /**
    * Constructs a new {@link MetricCatalog} from the given metrics, data providers, and options.
@@ -133,6 +143,8 @@ export class MetricCatalog {
     metrics: Array<string | Metric>,
     includeTimeseries = true
   ): Promise<MultiRegionMultiMetricDataStore> {
+    this.dataFetchesCount++;
+
     const resolvedMetrics = metrics.map((m) =>
       typeof m === "string" ? this.getMetric(m) : m
     );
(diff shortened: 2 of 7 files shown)


packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -255,27 +247,19 @@ export class MetricCatalog {
...
-        })
-        .catch((error) => {
-          console.error(`Error fetching metric data: ${error}`);
-          return { error };
-        });
-    }, [metricList, regions]);
-    return { data };
+    // ESLint complains about calling hooks from classes (React Hook
+    // "useDataForRegionsAndMetrics"cannot be called in a class component.) but
+    // this isn't actually a class component.  It's just a class.
```
mikelehen, line 255: TBH I'm not 100% sure if it's "okay" to have a hook defined inside a class like I'm doing.  Like maybe having components call `catalog.useData(...)` is going to cause issues?  The alternative would be to export the `useDataForRegionsAndMetrics()` function that this method calls and make everybody use that.  But I prefer it being a class method if it'll work.

.eslintrc.json
```diff
@@ -7,6 +7,7 @@
   "extends": [
     "eslint:recommended",
     "plugin:react/recommended",
+    "plugin:react-hooks/recommended",
```
mikelehen, line 10: This enables the same hook-related lint checks that `create-react-app` apps have by default, including making sure that your `useEffect()` calls specify all of their dependencies, etc.

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -194,20 +206,13 @@ export class MetricCatalog {
...
-          setData(data);
-        })
-        .catch((error) => {
-          console.error(
-            `Error fetching metric data for ${metric} for ${region.regionId}: ${error}`
-          );
-          return { error };
-        });
-    }, [region, metric]);
-    return { data };
```
mikelehen, line 210: I realized that these useData() methods can call each other, so the implementation guts only live inside `useDataForRegionsAndMetrics()`

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -287,3 +271,59 @@ export class MetricCatalog {
...
+  const [data, setData] = useState<MultiRegionMultiMetricDataStore>();
+  let resolvedMetrics = metrics.map((m) =>
+    typeof m === "string" ? catalog.getMetric(m) : m
+  );
+
+  // In order to allow people to pass in a new array of regions / metrics that
+  // contain the same regions / metrics as before without triggering additional
+  // fetches, we need this caching trick.
+  resolvedMetrics = useCachedArrayIfEqual(resolvedMetrics);
+  regions = useCachedArrayIfEqual(regions);
```
mikelehen, line 290: While I like this approach better than the `JSON.stringify` or string concatenation approaches, it is a bit unconventional so I ended up writing more comments than code, trying to make it clearer.  :-/ Let me know if you have any suggestions / better ideas!

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -287,3 +271,59 @@ export class MetricCatalog {
...
+ * @param value New array to check against the cached array.
+ * @returns The cached array if its contents is exactly equal to the new value, else the new value.
+ */
+function useCachedArrayIfEqual<T>(value: Array<T>): Array<T> {
+  const ref = useRef(value);
+  if (!isEqual(ref.current, value)) {
+    ref.current = value;
+  }
+  return ref.current;
+}
```
pnavarrc, line 329: Consider adding a few tests for this function so we can have a more explicit demonstration of how it works

pnavarrc (approved):
Looks good!

packages/metrics/src/MetricCatalog/MetricCatalog.ts
mikelehen: Sure, sounds good!

PR #85, opened 2022-08-22
Implement legend threshold as divs

faihegberg:
Cancelling [this PR](https://github.com/covid-projections/act-now-packages/pull/68) and re-implementing without using SVG instead.

### DIFF
--- a/packages/ui-components/src/components/LegendThreshold/LegendThresholdVertical.tsx
+++ b/packages/ui-components/src/components/LegendThreshold/LegendThresholdVertical.tsx
@@ -16,6 +16,10 @@ export interface LegendThresholdVerticalProps<T>
   getItemSublabel?: (item: T, itemIndex: number) => string;
 }
 
+/**
+ * `LegendThresholdVertical` represents a scale with thresholds that separate
+ * a set of levels. By default, the labels between each level are shown.
+ */
 const LegendThresholdVertical = <T,>({
   height = 265,
   width = 12,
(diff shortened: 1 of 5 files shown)


packages/ui-components/src/components/LegendThreshold/LegendThresholdVertical.tsx
```diff
@@ -17,51 +21,47 @@ export interface LegendThresholdVerticalProps<T>
...
-            </Group>
-          );
-        })}
-      </Group>
-    </svg>
+          {showLabels && (
+            <LegendLabelContainer>
+              {getItemLabel && (
+                <>
+                  <LegendLabel>{getItemLabel(item, itemIndex)}</LegendLabel>
```
pnavarrc: We should use our design system directly here instead of styling divs 

```tsx
<>
  <Typography variant="labelSmall">
    {getItemLabel(item, itemIndex)}
  </Typography>
  <Typography variant="paragraphSmall">
    {getItemLabel(item, itemIndex)}
  </Typography>
</>
```

packages/ui-components/src/components/LegendThreshold/LegendThresholdVertical.tsx
```diff
@@ -17,51 +21,47 @@ export interface LegendThresholdVerticalProps<T>
...
-      <defs>
-        <clipPath id={clipPathId}>
-          <rect
-            x={0}
-            y={0}
-            width={width}
-            height={height}
-            rx={borderRadius}
-            ry={borderRadius}
+    <LegendThresholdVerticalWrapper $height={height}>
```
pnavarrc: The wrapper doesn't need the `$height` prop because the children have an explicit height (if you do that you won't need the `shouldForwardProps` part on the styled component)

packages/ui-components/src/components/LegendThreshold/LegendThresholdVertical.tsx
```diff
@@ -17,51 +21,47 @@ export interface LegendThresholdVerticalProps<T>
...
+      {items.map((item, itemIndex) => (
+        <LegendContainer
+          key={getItemColor(item, itemIndex)}
+          $height={legendColorHeight}
+        >
```
pnavarrc: When using emotion we don't need to pass parameters with the `$` prefix. In this case, we can rely on the native styling options of the div

```
<LegendContainer
  key={getItemColor(item, itemIndex)}
  style={{ height: legendColorHeight }}
>
```

Also, I would recommend using the index as key instead of relying on the color, we don't have guarantees that the colors will be unique.

packages/ui-components/src/components/LegendThreshold/LegendThresholdVertical.tsx
```diff
@@ -17,51 +21,47 @@ export interface LegendThresholdVerticalProps<T>
...
+          <LegendColor
+            $width={width}
+            $color={getItemColor(item, itemIndex)}
+            $roundTop={itemIndex === 0 ? borderRadius : 0}
+            $roundBottom={itemIndex === numberOfItems - 1 ? borderRadius : 0}
```
pnavarrc: Same here, consider using `style` for `width` and `color` (`backgroundColor`) and remove the `$` prefix since that's not idiomatic for emotion (`shouldForwardProp: isValidProp` will prevent those props to be passed down to the underlying div`

packages/ui-components/src/components/LegendThreshold/LegendThresholdVertical.tsx
```diff
@@ -17,51 +21,47 @@ export interface LegendThresholdVerticalProps<T>
  * a set of levels. By default, the labels between each level are shown.
  */
 const LegendThresholdVertical = <T,>({
-  height = 300,
-  width = 20,
-  borderRadius,
+  height = 265,
+  width = 12,
```
pnavarrc: Consider renaming this to `barWidth` instead to be more consistent with `LegendThresholdHorizontal` - `width` and `height` should be reserved for the total width and height of the component to avoid confusion.

Also, since this is a div element, we don't need to pass the overall width of the component anymore, since it will just adapt to its container, so consider removing the `width` prop from the common props and make it a prop of the `LegendThresholdHorizontal` only

packages/ui-components/src/components/LegendThreshold/LegendThresholdVertical.tsx
```diff
@@ -17,51 +21,47 @@ export interface LegendThresholdVerticalProps<T>
...
+  borderRadius = 0,
```
pnavarrc, line 26: Consider setting the same default value for `LegendThresholdHorizontal` to make it explicit there too

pnavarrc (changes requested):
Added a few notes, looks pretty good though!

pnavarrc (approved):
Looks good!

PR #88, opened 2022-08-23
UI components: US national maps

chasulin:
US national maps (states + counties). County shapes render with an optional `showCounties` prop set to true. Storybook is currently rendering simple tooltips with the shape's state name (tooltip content is passed in via a `renderTooltip` prop, so anything can be rendered)

🗺 [Stories here](https://act-now-packages--pr88-casulin-us-maps-fi61nzrj.web.app/storybook/index.html?path=/story/components-usnationalmap--states-with-no-fill-color) 🗺


.eslintrc.json
```diff
@@ -18,6 +18,9 @@
     "sourceType": "module"
   },
   "plugins": ["react", "@typescript-eslint"],
-  "rules": {},
+  "rules": {
+    "@typescript-eslint/no-explicit-any": "off",
+    "prefer-const": "off"
```
chasulin, line 24: had to add these rules on behalf of the `any` and the `let` in [the importJson code](https://github.com/covid-projections/act-now-packages/pull/88/files#diff-dbac915eece0d9fad387d37708ae6fbbb41e42d48ee671307ce23c167cba4b5fR1-R3)

packages/ui-components/src/common/hooks/usePromise.ts
```diff
@@ -0,0 +1,42 @@
+import { useEffect, useState } from "react";
```
chasulin, line 1: this is all from CAN

packages/ui-components/src/common/hooks/usePromise.ts
```diff
@@ -0,0 +1,42 @@
...
+        setResult({ pending: false, error: err });
+      });
+
+    return () => {
+      canceled = true;
+    };
+  }, [promise]);
+
+  return result;
+}
```
chasulin, line 42: this is all from CAN

packages/ui-components/src/common/shapefiles/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export type CountiesTopology = typeof import("./counties-10m.json");
+
+function importCountyGeographies(): Promise<CountiesTopology> {
+  return importJson("counties-10m", import("./counties-10m.json"));
+}
+
+export const useCountyGeographies = singletonHook({ pending: true }, () =>
+  usePromise(importCountyGeographies())
+);
```
chasulin, line 20: also from CAN

packages/ui-components/src/common/utils/index.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @param importPromise The promise returned from a import('foo/bar.json') call.
+ */
+export function importJson<T>(
+  cacheToken: string,
+  importPromise: Promise<{ default: T }>
+): Promise<T> {
+  cachedImports[cacheToken] =
+    cachedImports[cacheToken] || importPromise.then((module) => module.default);
+  return cachedImports[cacheToken];
+}
```
chasulin, line 21: also from CAN

.eslintrc.json
```diff
@@ -19,6 +19,9 @@
...
+  },
```
pnavarrc, line 25: Maybe we can disable these rules for the specific lines we need to (or the file) instead of globally?

packages/ui-components/src/components/USNationalMap/USNationalMap.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+});
+
+const colors = alphabetArr.map((letter: string, i: number) =>
+  interpolatePiYG(testLinear(i))
+);
+
+const colorScaleAlpha = scaleOrdinal({
+  domain: alphabetArr,
+  range: colors,
+});
```
pnavarrc, line 58: Consider adding some comments to explain what's going on here (since it was kind of hard to figure out)

pnavarrc (approved):
Added a few notes, looks good!

.eslintrc.json
chasulin: i first used ts-ignore, but then we'd need to add a rule enabling ts-ignore (which we currently don't allow). and then it still wasnt working as expected, so i just added the rules.. 😬 .. strong feelings?

pnavarrc: lets leave it but maybe try to figure this out later! :)

PR #90, opened 2022-08-23
Add `legendCatagories` to `MetricDefinition` and `Metric`

smcclure17:
For eventual use with [`LegendCategorical()`](https://github.com/covid-projections/act-now-packages/tree/develop/packages/ui-components/src/components/LegendCategorical), or similar components

### DIFF
--- a/packages/metrics/src/Metric/Metric.test.ts
+++ b/packages/metrics/src/Metric/Metric.test.ts
@@ -22,6 +22,12 @@ const testLevelSet: MetricLevelSet = {
   ],
 };
 
+// Example of a typical testCategories to be used by metrics.
+const testCategories: MetricCategory[] = [
+  { label: "Not started", color: "red", value: "no" },
+  { label: "Done", color: "green", value: "yes" },
+];
+
 // Example of typical MetricCatalogOptions, including a default metric level set.
 const testCatalogOptions: MetricCatalogOptions = {
   metricLevelSets: [testLevelSet],
@@ -115,18 +121,13 @@ describe("Metric", () => {
     expect(metric.roundValue(0.123)).toBe(0.12);
   });
 
-  test("getCategory() with string value and no match.", () => {
-    const testCategories: MetricCategory[] = [
-      { label: "Not started", color: "red", value: "no" },
-      { label: "Done", color: "green", value: "yes" },
-    ];
+  test("getCategory() fails with no matching value.", () => {
     const metric = new Metric({
       ...testMetricDef,
       categories: testCategories,
     });
-    expect(metric.getCategory("no")).toStrictEqual(testCategories[0]);
     expect(() => {
       metric.getCategory("none");
-    }).toThrow("No matching category");
+    }).toThrow("No matching");
   });
 });
--- a/packages/metrics/src/Metric/Metric.test.ts
+++ b/packages/metrics/src/Metric/Metric.test.ts
@@ -2,6 +2,7 @@ import { Metric } from "./Metric";
 import { MetricLevelSet } from "./MetricLevel";
 import { MetricDefinition } from "./MetricDefinition";
 import { MetricCatalogOptions } from "../MetricCatalog";
+import { MetricCategory } from "./MetricCategory";
 
 // Example of a typical metric with mostly default options.
 const testMetricDef: MetricDefinition = {
@@ -112,5 +113,20 @@ describe("Metric", () => {
       formatOptions: { style: "percent", maximumFractionDigits: 0 },
     });
     expect(metric.roundValue(0.123)).toBe(0.12);
+  });
+
+  test("getCategory() with string value and no match.", () => {
+    const testCategories: MetricCategory[] = [
+      { label: "Not started", color: "red", value: "no" },
+      { label: "Done", color: "green", value: "yes" },
+    ];
+    const metric = new Metric({
+      ...testMetricDef,
+      categories: testCategories,
+    });
+    expect(metric.getCategory("no")).toStrictEqual(testCategories[0]);
+    expect(() => {
+      metric.getCategory("none");
+    }).toThrow("No matching category");
   });
 });
--- a/packages/metrics/src/Metric/Metric.test.ts
+++ b/packages/metrics/src/Metric/Metric.test.ts
@@ -121,6 +121,14 @@ describe("Metric", () => {
     expect(metric.roundValue(0.123)).toBe(0.12);
   });
 
+  test("getCategory() with string values", () => {
+    const metric = new Metric({
+      ...testMetricDef,
+      categories: testCategories,
+    });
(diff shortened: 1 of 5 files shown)


packages/metrics/src/Metric/Metric.ts
```diff
@@ -38,6 +38,12 @@ export class Metric {
   readonly levelSetId: string;
   /** {@inheritDoc MetricDefinition.formatOptions} */
   readonly formatOptions: Intl.NumberFormatOptions;
+  /** {@inheritDoc MetricDefinition.legendCatagories} */
+  readonly legendCatagories?: {
+    label: string;
+    color: string;
```
smcclure17: Not 100% sure if we want to have `color` be required as I could imagine us having legends with non-color markers, (e.g. shapes), but that seemed the most straightforward way to work with the `LegendCategorical()` component

mikelehen: Yeah, we can always change it later!

packages/metrics/src/Metric/Metric.ts
```diff
@@ -38,6 +38,12 @@ export class Metric {
...
+    [extra: string]: unknown;
```
mikelehen: Did you have a specific use case in mind for this?

packages/metrics/src/Metric/MetricDefinition.ts
```diff
@@ -52,6 +52,15 @@ export interface MetricDefinition {
    */
   levelSetId?: string;
 
+  /**
+   * Legends for representing categorical metrics.
+   */
+  legendCatagories?: {
```
mikelehen: Catagories => Categories

If you don't already have it installed, I recommend https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

I wonder if this should just be "categories" instead of "legendCategories".  I think maybe the "legend" is the UI component, and "categories" and "thresholds" are two ways of "grading" / "categorizing" metrics for UI purposes.

packages/metrics/src/Metric/MetricDefinition.ts
```diff
@@ -52,6 +52,15 @@ export interface MetricDefinition {
...
+    label: string;
+    color: string;
```
mikelehen: I was imagining making this a bit more powerful / comprehensive like what we have in GCAT (e.g. https://github.com/covid-projections/gcat/blob/develop/src/cms-content/metrics/metrics.json#L196)

Basically when you define the metric, you can map metric values to legend labels/colors and then `Metric` has logic for coloring metric values based on the defined legend (https://github.com/covid-projections/gcat/blob/49b319164a3e55c894cabd53e18740f02bf4ffa5/src/common/metrics/metric.ts#L172) or using [formatValue](https://github.com/covid-projections/gcat/blob/49b319164a3e55c894cabd53e18740f02bf4ffa5/src/common/metrics/metric.ts#L96) to get the legend label for a value (e.g. for use in tooltips).

WDYT?  This could potentially be a second PR if we like.  As-is it's probably enough to power `LegendCategorical`.

packages/metrics/src/Metric/Metric.ts
```diff
@@ -72,6 +78,7 @@ export class Metric {
     this.levelSetId = def.levelSetId ?? "default";
     this.levelSet = (levelSets || []).find((ls) => ls.id === this.levelSetId);
     this.formatOptions = def.formatOptions ?? DEFAULT_FORMAT_OPTIONS;
+    this.legendCatagories = def.legendCatagories;
```
mikelehen: It occurs to me that maybe we could / should have an assert that prevents you from specifying both categories and thresholds.  Because we can color a map and give you a legend based on categories or thresholds, but you have to choose which one.

packages/metrics/src/Metric/MetricDefinition.ts
```diff
@@ -52,6 +52,15 @@ export interface MetricDefinition {
...
+    [extra: string]: unknown;
+  }[];
```
mikelehen: stylistic nit: With TypeScript you can declare an array as `Foo[]` or `Array<Foo>`.  The former is shorter and nice for concise type definitions, but when the type definition gets longer (especially multiple lines), I would do `Array<...>` instead.

In this case I initially didn't realize you had defined it as an array since I saw: `legendCategories?: { ...` which just looks like an object definition.

mikelehen (approved):
Looks good, though I think we can make the `Metric` class a bit more "category-aware" and provide help coloring / formatting values.  This could be a separate PR though if you want so approving as-is!

packages/metrics/src/Metric/Metric.ts
smcclure17: No not really--I was thinking since `LegendCategorical()` accepted arbitrary parameters maybe we should be able to supply them, but that probably isn't necessary. I've removed it!

packages/metrics/src/Metric/MetricDefinition.ts
smcclure17: oop 😢 😭 thanks--I do have a spell checker installed somehow I must have missed it :/ 

And yeah, that makes sense, I've changed this to just `categories` in e40d0b1c045455854b4e964a26f6f74c35aeaaaa

packages/metrics/src/Metric/MetricDefinition.ts
smcclure17: Ahh, yeah that sounds good. I've made changes in e40d0b1c045455854b4e964a26f6f74c35aeaaaa to hopefully move this closer to that

packages/metrics/src/Metric/Metric.test.ts
```diff
@@ -113,4 +120,22 @@ describe("Metric", () => {
...
+
+  test("getCategory() with string values", () => {
+    const metric = new Metric({
+      ...testMetricDef,
+      categories: testCategories,
+    });
+    expect(metric.getCategory("no")).toStrictEqual(testCategories[0]);
+  });
+
+  test("getCategory() fails with no matching value.", () => {
```
mikelehen, line 132: FWIW- GCAT allows you to specify a "default" category with no value specified.  We might want that eventually, but for now I'm happy to be more strict.

packages/metrics/src/Metric/MetricCategory.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * A category for representing a discrete metric values.
```
mikelehen: nit: values => value (or remove "a")

mikelehen (approved):
Looks great!

packages/metrics/src/Metric/Metric.test.ts
smcclure17: Cool, yeah true--if we're okay with leaving this as is for now I think I will, but I can see that potentially being helpful in the future

PR #91, opened 2022-08-23
 Set max old space size NODE_OPTION to avoid crash.

mikelehen:


pnavarrc (approved):
Looks good, thanks Michael!

PR #94, opened 2022-08-24
Metrics / CsvDataProvider fixes / tweaks.

mikelehen:
Few minor fixes uncovered while integrating with hackathon repo:

* Make react a peerDependency instead of normal dependency to avoid multiple versions of react being included.
* Export `CsvDataProvider` and `CsvDataProviderOptions` so that they can be used outside the package.
* Always preserve the region column as a raw string when parsing CSV data (to avoid turning fips codes into numbers).
* Gracefully return null data instead of throwing an error when there's no data for a particular region + metric combo.
* Handle when we fail to fetch the CSV from the given URL and give a useful error.  Before it was ending up getting parsed as an empty CSV.
* Allow empty lines in CSV file.  Needed this since pandas seems to leave a trailing empty line when you use `to_csv()`.

### DIFF
--- a/packages/metrics/package.json
+++ b/packages/metrics/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@actnowcoalition/metrics",
-  "version": "0.0.5",
+  "version": "0.0.6",
   "description": "Classes for representing metrics and loading metric data",
   "repository": {
     "type": "git",
@@ -26,8 +26,12 @@
   "devDependencies": {
     "@testing-library/react-hooks": "^8.0.1",
     "@types/lodash": "^4.14.182",
+    "react": "^18.2.0",
     "react-test-renderer": "^18.2.0",
     "typescript": "^4.6.4"
+  },
+  "peerDependencies": {
+    "react": "^18.2.0"
   },
   "sideEffects": false,
   "dependencies": {
@@ -37,7 +41,6 @@
     "@actnowcoalition/time-utils": "^1.0.1",
     "@types/papaparse": "^5.3.3",
     "lodash": "^4.17.21",
-    "papaparse": "^5.3.2",
-    "react": "^18.2.0"
+    "papaparse": "^5.3.2"
   }
 }
--- a/packages/metrics/src/data_providers/CsvDataProvider.test.ts
+++ b/packages/metrics/src/data_providers/CsvDataProvider.test.ts
@@ -3,11 +3,13 @@ import { Metric } from "../Metric";
 import { states } from "@actnowcoalition/regions";
 
 const mockCsv = `region,cool_metric
-                  36,150
-                  12,`;
+36,150
+12,`;
+
 const csvTimeseries = `region,date,cool_metric
-                        36,2022-08-02,150
-                        12,2022-08-02,`;
+36,2022-08-02,150
+12,2022-08-02,`;
+
 const newYork = states.findByRegionIdStrict("36");
 const testMetric = new Metric({
   id: "metric",
--- a/packages/metrics/src/data_providers/CsvDataProvider.ts
+++ b/packages/metrics/src/data_providers/CsvDataProvider.ts
@@ -11,7 +11,7 @@ import {
 import { groupBy } from "lodash";
 import Papa from "papaparse";
 
-interface CsvDataProviderOptions {
+export interface CsvDataProviderOptions {
   /** URL of a CSV file to import from. */
   url?: string;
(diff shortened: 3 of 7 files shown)


packages/metrics/package.json
```diff
@@ -26,9 +26,13 @@
   "devDependencies": {
     "@testing-library/react-hooks": "^8.0.1",
     "@types/lodash": "^4.14.182",
+    "react": "^18.2.0",
```
mikelehen, line 29: I made react a `peerDependency` instead of a normal `dependency` which also means I needed to make it a `devDependency` so it's available for our tests.

packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -3,11 +3,13 @@ import { Metric } from "../Metric";
 import { states } from "@actnowcoalition/regions";
 
 const mockCsv = `region,cool_metric
-                  36,150
-                  12,`;
+36,150
+12,`;
```
mikelehen, line 7: When I changed the CSV parser to preserve the region column as a string we ended up with a bunch of spaces prefixed to the regions (e.g. `".                 36"`)

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -118,6 +118,11 @@ export class CsvDataProvider extends CachingMetricDataProviderBase {
   private async fetchCsvText(): Promise<string> {
     assert(this.url, "URL must be specified in order to use fetchCsvText()");
     const response = await fetch(this.url);
+    if (response.status !== 200) {
+      throw new Error(
+        `Failed to fetch CSV data from ${this.url}: ${response.statusText}`
+      );
+    }
```
mikelehen, line 125: FWIW- I am explicitly throwing an error instead of using `assert()` since this is something that could happen during "normal" operation of an app (if the network is down or whatever), whereas I reserve `assert` for programming errors.

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -126,12 +131,41 @@ export class CsvDataProvider extends CachingMetricDataProviderBase {
...
+ * @param regionColumn Name of column containing region IDs. It will be
+ * preserved as a string, even if it has numeric values, to preserve FIPS codes
+ * as strings.
+ * @returns parsed CSV rows.
  */
-function parseCsv(csvText: string): DataRow[] {
+function parseCsv(csvText: string, regionColumn: string): DataRow[] {
   const csv = Papa.parse(csvText, {
     header: true,
-    dynamicTyping: true,
```
mikelehen, line 134: Replacing this with the `sanitizeRows()` call below which is aware of the `regionColumn`.

packages/metrics/src/data_providers/data_provider_utils.test.ts
```diff
@@ -52,41 +52,22 @@ describe("dataRowsToMetricData()", () => {
...
     }).toThrow();
-    expect(() => {
-      dataRowsToMetricData(
-        dataRows,
-        newYork,
-        testMetric,
-        column,
-        /*dateKey=*/ "date"
-      );
-    }).toThrow();
```
mikelehen, line 63: I think this was accidentally duplicated from directly above?  So I removed it.

packages/metrics/src/data_providers/data_provider_utils.test.ts
```diff
@@ -104,15 +85,6 @@ describe("dataRowsToMetricData()", () => {
...
     ).toBe(null);
-    expect(
-      dataRowsToMetricData(
-        dataRows,
-        newYork,
-        testMetric,
-        column,
-        /*dateKey=*/ "date"
-      ).currentValue
-    ).toBe(null);
```
mikelehen, line 115: This was duplicated too, I think?

packages/metrics/src/data_providers/CsvDataProvider.ts
smcclure17: Ah cool, thanks for the distinction!

packages/metrics/src/data_providers/data_provider_utils.test.ts
smcclure17: Hmm, yeah, 😭 thanks and sorry about that!

smcclure17 (approved):
Looks great, thanks! 🙏 

Just out of curiosity, do you have/could you share the data you were testing this with?

PR #97, opened 2022-08-24
Add date truncation functions to time-utils package

smcclure17:
Moves some logic from different pieces of different packages into the `time-utils` package

### DIFF
--- a/packages/metrics/src/data_providers/data_provider_utils.ts
+++ b/packages/metrics/src/data_providers/data_provider_utils.ts
@@ -3,7 +3,10 @@ import { Region } from "@actnowcoalition/regions";
 import { Metric } from "../Metric";
 import { MetricData } from "../data";
 import { Timeseries } from "../Timeseries";
-import { parseDateString } from "@actnowcoalition/time-utils";
+import {
+  parseDateString,
+  stripHoursMinutesSeconds,
+} from "@actnowcoalition/time-utils";
 
 export type DataRow = { [key: string]: unknown };
 
@@ -66,7 +69,7 @@ export function dataRowsToMetricData(
         `Date column must be a string. ${typeof row[dateKey]} found.`
       );
       return {
-        date: stripTime(parseDateString(row[dateKey] as string)),
+        date: stripHoursMinutesSeconds(parseDateString(row[dateKey] as string)),
         value: row[metricKey] as unknown,
       };
     })
@@ -77,16 +80,4 @@ export function dataRowsToMetricData(
     timeseries.last?.value ?? null,
     timeseries
   );
-}
-
-/**
- * Remove hours minutes and seconds from Javascript Date object.
- *
- * TODO: Merge this logic with Timeseries.isoDateString() and move to time-utils package.
- *
- * @param date Date object to truncate.
- */
-function stripTime(date: Date): Date {
-  const truncatedDate = date.toISOString().split("T")[0];
-  return new Date(truncatedDate);
 }
--- a/packages/time-utils/src/index.ts
+++ b/packages/time-utils/src/index.ts
@@ -198,6 +198,9 @@ export function getTimeDiff(date1: Date, date2: Date, unit: TimeUnit) {
 /**
  * Remove hours, minutes and seconds from a Javascript Date object.
  *
+ * @example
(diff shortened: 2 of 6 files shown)


packages/time-utils/src/index.ts
```diff
@@ -194,3 +194,29 @@ export function getTimeDiff(date1: Date, date2: Date, unit: TimeUnit) {
...
+/**
+ * Format a date in ISO format excluding hours, minutes, and seconds.
+ *
+ * @example
+ * truncatedIsoString(new Date(2020, 3, 2, 10, 30))  // "2020-04-02"
+ *
+ * @param date Date to transform.
+ * @returns Date in ISO format with hours, minutes, and seconds removed.
+ */
+export function truncatedIsoString(date: Date): string {
```
smcclure17, line 220: Not sure I love `truncatedIsoString` as a function name, but nothing better immediately came to mind

mikelehen: Yeah...  I kinda' like `isoDateString()` but perhaps it's not clear enough that it only includes the date, not the date/time?  Maybe `isoDateOnlyString()`?  Don't love it, but I think it's clearer than "truncated"

packages/time-utils/src/index.ts
```diff
@@ -194,3 +194,29 @@ export function getTimeDiff(date1: Date, date2: Date, unit: TimeUnit) {
...
+/**
+ * Remove hours, minutes and seconds from a Javascript Date object.
+ *
+ * @example
+ * stripHoursMinutesSeconds(new Date(2020, 3, 2, 10, 30))  // "2020-04-02T00:00:00.000Z"
+ *
+ * @param date Date to modify.
+ * @returns Truncated Date object.
+ */
+export function stripHoursMinutesSeconds(date: Date): Date {
```
mikelehen, line 207: Hrm...  I am not sure if we should actually ever be stripping hours/minutes/seconds off of dates...  or at least not often.

I think maybe instead we should have an `assertDateOnly()` helper that asserts if the hours/min/seconds are all zero, similar to https://github.com/covid-projections/act-now-packages/blob/7cb9890104987ff37cc2ee9b531b95e8d1a1699b/packages/metrics/src/Timeseries/Timeseries.ts#L338

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -66,7 +69,7 @@ export function dataRowsToMetricData(
         `Date column must be a string. ${typeof row[dateKey]} found.`
       );
       return {
-        date: stripTime(parseDateString(row[dateKey] as string)),
+        date: stripHoursMinutesSeconds(parseDateString(row[dateKey] as string)),
```
mikelehen, line 72: As mentioned below I think maybe instead of *removing* the hours/min/seconds we should assert they are all zero.

Though technically we may not need to, since Timeseries will do that automatically.  Or am I confused and there is a reason that we're expecting non-zero time components?

mikelehen (changes requested):
This looks good, but with a couple nits and I think it needs to be split into two PRs for versioning purposes:

1. A PR that adds the functions to time-utils and bumps the version number in time-utils/package.json.  Once that's merged, we can publish it.

2. Update the metrics/package.json to depend on the newly-published version of time-utils, and then actually consume the new functions.

Basically when we're testing locally, we're using the latest code from all packages as if it's all one uniform codebase.  This is nice for testing , but for actual published packages we need to make sure that they are depending on the right versions of the right packages.  As-is, the metrics package depends on time-utils `^1.0.1` which does *not* have the new functions, and so it would potentially be broken.

As a rule of thumb, we should never make a PR that changes code in one package which another package is going to depend on.  cc/ @pnavarrc FYI since we talked about this recently.

packages/metrics/src/data_providers/data_provider_utils.ts
smcclure17: Yeah, that sounds good--I don't actually remember why we're stripping the times here. I think it may have been just to make sure that the loaded timeseries points were valid, but we should probably leave that up to the user/provider of the data

smcclure17:
Good flag @mikelehen--I'll move these into separate PRs, starting with https://github.com/covid-projections/act-now-packages/pull/98 to add the util functions 🙏

PR #100, opened 2022-08-24
Upgrade @actnowcoalition/regions for metrics

pnavarrc:
I tried to use `MetricData(metric, region, value)` when implementing the MetricOverview component, but I couldn't because the version of `@actnowcoalition/regions` in `ui-components` and `metrics` are different. This PR updates `metrics` to use the latest `@actnowcoalition/regions`.

### DIFF
--- a/packages/metrics/package.json
+++ b/packages/metrics/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@actnowcoalition/metrics",
-  "version": "0.0.6",
+  "version": "0.0.7",
   "description": "Classes for representing metrics and loading metric data",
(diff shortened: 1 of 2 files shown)


mikelehen (approved):
Looks good.

I am starting to think that we might want to give all of our packages the same version number and publish them together.

Basically make the publish action let you specify the version number, and then it'll update all the package.json files with that version number, and publish them all.

That way what is published is equivalent to what we are testing / debugging locally.  Not 100% sure though.

pnavarrc:
Agreed, that's probably the simplest way 👍

PR #102, opened 2022-08-25
use time-utils' `isoDateOnlyString()` in metrics package 

smcclure17:
follow up from #98


packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -66,7 +65,7 @@ export function dataRowsToMetricData(
         `Date column must be a string. ${typeof row[dateKey]} found.`
       );
       return {
-        date: stripTime(parseDateString(row[dateKey] as string)),
+        date: new Date(row[dateKey] as string),
```
smcclure17, line 68: Using `new Date()` instead of `parseDateString()` because `parseDateString` was adding non-zero time components to dates (due to timezones) that could change the day/date incorrectly

mikelehen (approved):
Looks great!

PR #103, opened 2022-08-25
Edit map dependencies

chasulin:
Adds in missing useMemo dependencies

### DIFF
--- a/packages/ui-components/src/components/USNationalMap/USNationalMap.tsx
+++ b/packages/ui-components/src/components/USNationalMap/USNationalMap.tsx
@@ -64,7 +64,7 @@ const USNationalMapFC: React.FC<USNationalMapProps> = ({
         {showCounties && <CountiesMap getFillColor={getFillColor} />}
       </>
     );
-  }, [getFillColor]);
+  }, [getFillColor, renderTooltip, showCounties]);
 
   return (
     <ComposableMap projection={projection} height={500}>


mikelehen (approved):
Thanks!  I think this is the right fix, but just FYI it does mean that consumers of the map will probably have to be careful to pass the exact same renderTooltip every time, else the map will re-render.

So e.g. if they're using an inline lambda function inside their component (which will return a new function every time it renders), they may need to use `useMemo()` or similar to memoize the callback.

Fortunately our metric-aware wrapper can handle that transparently so we don't have to think about it. 🤞

chasulin:
> Thanks! I think this is the right fix, but just FYI it does mean that consumers of the map will probably have to be careful to pass the exact same renderTooltip every time, else the map will re-render.
> 
> So e.g. if they're using an inline lambda function inside their component (which will return a new function every time it renders), they may need to use `useMemo()` or similar to memoize the callback.
> 
> Fortunately our metric-aware wrapper can handle that transparently so we don't have to think about it. 🤞

Oooh interesting. I'm not sure I fully understand the lambda function example, but I'm going to merge per your last comment that a wrapper can handle it ⚡️

mikelehen:
@chasulin Oh, actually this was already true for `getFillColor`.  But basically it just means that when you use the component you have to be careful to use useMemo() or useCallback() on your getFillColor / renderTooltip callbacks (like we do in CAN [here](https://github.com/covid-projections/covid-projections/blob/develop/src/components/USMap/USRiskMap.tsx#L18) if you're using a lambda function (i.e. `(fips) => {...}`) inside the component.

PR #107, opened 2022-08-29
Add PureDate class to time-utils package.

mikelehen:
This makes it easier to deal with "pure" dates (with no time component) like we use in Timeseries.

This came in handy when trying to write a 7-day rolling average function.


pnavarrc (approved):
Looks good!

PR #108, opened 2022-08-29
CAN API data provider

smcclure17:


packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -73,7 +73,7 @@ export function dataRowsToMetricData(
   return new MetricData(
     metric,
     region,
-    timeseries.last?.value ?? null,
+    timeseries.removeNils().last?.value ?? null,
```
smcclure17: Made this change as it seems like we probably want the `currentValue` to be the last, non-null value (like we do in CAN). 

Though, I imagine we might want more flexibility here in the future (e.g. the ability to set a "max lookback days" variable like we do for CAN)

packages/metrics/src/data_providers/CanRegionDataProvider.test.ts
```diff
@@ -0,0 +1,120 @@
...
+const testData = {
+  "25017": testApiResponse,
+  "25": testApiResponse,
+  "10420": testApiResponse,
+  USA: testApiResponse,
+};
+
+const dataProviders = [
+  new CanRegionDataProvider(
+    /*providerId=*/ "can-region",
```
mikelehen: FWIW- We could potentially just "hardcode" the providerId in the `CanRegionDataProvider` class (rather than specifying it in the constructor) since I can't think of a reason you'd need to create multiple instances with different names (unlike the CsvDataProvider where you might have multiple CSV files you're working with).

packages/metrics/src/data_providers/CanRegionDataProvider.test.ts
```diff
@@ -0,0 +1,120 @@
...
+};
+const testData = {
+  "25017": testApiResponse,
+  "25": testApiResponse,
+  "10420": testApiResponse,
+  USA: testApiResponse,
+};
+
+const dataProviders = [
+  new CanRegionDataProvider(
```
mikelehen: Out of curiosity are you imagining we'll want / need other variations of CAN data providers?  I'd be slightly tempted to just name this `CanDataProvider` (or `CovidActNowDataProvider` to be a bit more explicit since CAN is mostly an internal acronym).  WDYT?

I know we have aggregate endpoints (states.timeseries.json etc), but since they provide the same data, I'm not sure why we would need to use them.  We *might* want to use them in certain cases for performance reasons, but I think the data provider could just do that automatically.  E.g. If you're requesting a bunch of regions, then it might fetch the aggregated endpoint instead of all of the individual ones.  But realistically I'm not sure we'd ever bother implementing that optimization.

packages/metrics/src/data_providers/CanRegionDataProvider.test.ts
```diff
@@ -0,0 +1,120 @@
...
+      [testCounty, testNation, testState, testMetro],
+      [MetricId.HOSPITAL_BEDS, MetricId.WEEKLY_ADMISSIONS, MetricId.POPULATION],
+      /*includeTimeseries=*/ false
+    );
+    const countyData = data.regionData(testCounty);
+    const nationData = data.regionData(testNation);
+    const stateData = data.regionData(testState);
+    const metroData = data.regionData(testMetro);
+
+    for (const data of [countyData, nationData, stateData, metroData]) {
```
mikelehen: nit: You could save a few lines of code with:
```
for (const region of [testCounty, testNation, testState, testMetro]) {
  const d = data.regionData(region);
  ...
```

packages/metrics/src/data_providers/CanRegionDataProvider.test.ts
```diff
@@ -0,0 +1,120 @@
...
+      [testCounty, testNation, testState, testMetro],
+      [MetricId.HOSPITAL_BEDS, MetricId.WEEKLY_ADMISSIONS],
+      /*includeTimeseries=*/ true
+    );
+    const countyData = data.regionData(testCounty);
+    const nationData = data.regionData(testNation);
+    const stateData = data.regionData(testState);
+    const metroData = data.regionData(testMetro);
+
+    for (const data of [countyData, nationData, stateData, metroData]) {
```
mikelehen: Same nit here.

packages/metrics/src/data_providers/CanRegionDataProvider.test.ts
```diff
@@ -0,0 +1,120 @@
...
+        data.metricData(MetricId.HOSPITAL_BEDS).timeseries.first?.value
+      ).toBe(100);
+      expect(data.metricData(MetricId.WEEKLY_ADMISSIONS).currentValue).toBe(10);
+      expect(
+        data.metricData(MetricId.WEEKLY_ADMISSIONS).timeseries.first?.value
+      ).toBe(10);
+    }
+  });
+
+  test("fetch fails if includeTimeseries is specified for a metric that has no timeseries data.", async () => {
```
mikelehen: FWIW- I'm not sure if we need to be this strict.  I could see it being handy to fetch a bunch of metrics at once, some of which have timeseries and some don't.

I think of `includeTimeseries` mostly as a performance optimization so that the data provider can optionally do less work / cache less data if you don't need the timeseries data.

That said I don't feel very strongly so we could start with this behavior and then change it later if it feels too strict. 🤷

packages/metrics/src/data_providers/CanRegionDataProvider.ts
```diff
@@ -0,0 +1,178 @@
...
+import { counties, metros, Region } from "@actnowcoalition/regions";
+import { Metric } from "../Metric";
+import { assert } from "@actnowcoalition/assert";
+import {
+  dataRowsToMetricData,
+  dataRowToMetricData,
+} from "./data_provider_utils";
+import flatten from "flat";
+import { DataRow } from "./data_provider_utils";
+import fetch from "node-fetch";
```
mikelehen: Not sure if we actually need `node-fetch`?  Though we might if we want this code to work from node later (e.g. to run a script that downloads data and caches it in a snapshot).  We don't import `node-fetch` in the CsvDataProvider.

If we *do* want to use node-fetch, I think we need to add `node-fetch` to the package.json (probably via `yarn add node-fetch` and `yarn add @types/node-fetch`).  I'm guessing it's only working as-is because one of our dependencies already depends on node-fetch or something.

packages/metrics/src/data_providers/CanRegionDataProvider.ts
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * Data provider to ingest data from a Covid Act Now region API endpoint.
```
mikelehen: Consider "from a Covid Act Now region API endpoint." => "from the Covid Act Now API."

packages/metrics/src/data_providers/CanRegionDataProvider.ts
```diff
@@ -0,0 +1,178 @@
...
+  /** Cached CAN API responses indexed by the FIPS codes of the regions fetched. */
+  private flattenedApiJson: { [regionId: string]: DataRow } = {};
+
+  constructor(
+    providerId: string,
+    apiKey?: string,
+    data?: { [regionId: string]: DataRow }
+  ) {
+    if (!data) {
+      assert(apiKey, "API key must be provided if test data is not provided.");
```
mikelehen: nit: Drop the `if` and just do `assert(data || apiKey, ...)` ?

packages/metrics/src/data_providers/CanRegionDataProvider.ts
```diff
@@ -0,0 +1,178 @@
...
+    }
+    super(providerId);
+    this.apiKey = apiKey;
+    this.data = data;
```
mikelehen: nit: Not a huge deal but rather than having this special `data` field we store the data in, could we massage it so we can just store it in `flattenedApiJson` like normal cached data?  Then we don't need to have extra state variables or special logic in `populateCache`.

packages/metrics/src/data_providers/CanRegionDataProvider.ts
```diff
@@ -0,0 +1,178 @@
...
+    metrics: Metric[],
+    includeTimeseries: boolean
+  ) {
+    for (const region of regions) {
+      let json;
+      if (this.data) {
+        json = this.data[region.regionId];
+      } else {
+        const url = this.buildFetchUrl(region, includeTimeseries);
+        const response = await fetch(url);
```
mikelehen: I think we should skip fetching data for regions that are already in the cache.  Though we'll have to somehow make sure we respect `includeTimeseries` so we don't return cached data without a timeseries.  The easiest way is probably to just include it in the cache key, e.g.

```
const cacheKey = `${region.regionId}-${includeTimeseries}`;
if (!this.flattenedApiJson[cacheKey]) {
  // ... fetch json data
  this.flattenedApiJson[cacheKey] = flattenDataForRegion(json);
}
```

Although it'd be nice if we didn't re-fetch non-timeseries data when we already have timeseries data cached.  So we might want a helper for getting data from the cache like:

```
function getCachedData(region: Region, includeTimeseries: boolean) {
  // If we have timeseries data cached, we always use it.
  const cachedTimeseriesData = this.flattenedApiJson[`${region.regionId}-true`];
  if (cachedTimeseriesData) {
    return cachedTimeseriesData);
  } else if (!includeTimeseries) {
    // check for non-timeseries data cached.
    return this.flattenedApiJson[`${region.regionId}-false`];
  }
```

And then use that here and in getDataFromCache().

packages/metrics/src/data_providers/CanRegionDataProvider.ts
```diff
@@ -0,0 +1,178 @@
...
+
+  getDataFromCache(
+    region: Region,
+    metric: Metric,
+    includeTimeseries: boolean
+  ): MetricData<unknown> {
+    const metricKey = metric.dataReference?.column;
+    assert(
+      typeof metricKey === "string",
+      `Metric key must exist and be of type string.`
```
mikelehen: This error won't be super helpful since we say "metric key" but the field we're looking for is "column".  Consider something like ``` `Metrics using ${this.id} data provider must specify the CAN API field to access via the dataReference.column property.` ```

packages/metrics/src/data_providers/CanRegionDataProvider.ts
```diff
@@ -0,0 +1,178 @@
...
+    );
+
+    if (includeTimeseries) {
+      // Get the timeseries prefix from the flat metric key by splitting on the first '.'.
+      // E.g. 'actuals.hospitalBeds.capacity' -> ['actuals', 'hospitalBeds.capacity', '']
+      const metricKeyParts = metricKey.split(/\.(.*)/);
```
mikelehen: nit: Would `metricKey.split('.')` be sufficient, or no?

packages/metrics/src/data_providers/CanRegionDataProvider.ts
```diff
@@ -0,0 +1,178 @@
...
+  } else if (region.regionId.length === 2) {
+    return "state";
+  } else {
+    const countyOrNull = counties.findByRegionId(region.regionId);
+    const metroOrNull = metros.findByRegionId(region.regionId);
+    assert(
+      (countyOrNull && !metroOrNull) || (!countyOrNull && metroOrNull),
+      `Cannot determine region type for region: ${region.regionId}`
+    );
+    return countyOrNull ? "county" : "cbsa";
```
mikelehen: Hrm.  This is a kind of roundabout way to determine if a region is a county or a metro.  I'm also not wild about us importing `counties` and `metros` just to do this check.  It'll bloat the bundle size a bit if e.g. the app only cares about states (or states and counties but not metros).  Ideally the actual app would be the only place we're importing regions.  It makes me wonder if we need some kind of `regionType` on regions...

FWIW I think we could just do `const isCounty = region.parent != null` since counties have parents (the state) but metros don't.  This is arguably more hacky, but I kind of prefer it for the above reasons.

cc/ @pnavarrc as FYI.

packages/metrics/src/data_providers/CanRegionDataProvider.ts
```diff
@@ -0,0 +1,178 @@
...
+ * Flattens the objects inside first-level arrays while leaving the arrays themselves unchanged.
+ * This is used to flatten the data inside the CAN timeseries (e.g. actualsTimeseries) without
+ * flattening the arrays themselves.
+ *
+ * @param json JSON to flatten.
+ * @returns Flattened JSON.
+ */
+function flattenDataForRegion(json: DataRow) {
+  // TODO: write our own flatten function such that we can remove npm-flat as a dependency,
+  // and so that we don't have to iterate through all the array keys like this.
```
mikelehen: Instead of flattening all the JSON and storing it that way, could we store the JSON as-is, but then use lodash [get()](https://lodash.com/docs/4.17.15#get) to look up dot-separated paths in the JSON?  Feel free to punt if you don't feel like going down that path right now, but it feels like it might be easier / cleaner.

mikelehen: Oh.  I guess we need to flatten it for it to work as-is with `dataRowToMetricData()`.  One workaround would be to allow you to pass in either a string column id or a function that gets the column from the row.  So you could equivalently do:

```
dataRowsToMetricData(rows, region, metric, 'cases', 'date')
// or
dataRowsToMetricData(rows, region, metric, row => row.cases, 'date')
```

That's how a lot of lodash functions work, e.g. here's the `sumBy` example from [the lodash docs](https://lodash.com/docs/4.17.15):
```
var objects = [{ 'n': 4 }, { 'n': 2 }, { 'n': 8 }, { 'n': 6 }];
 
_.sumBy(objects, function(o) { return o.n; });
// => 20
 
// The `_.property` iteratee shorthand.
_.sumBy(objects, 'n');
// => 20
```

That said, this is getting to be a bigger refactor, so feel 100% free to stick with using the flatten logic for now!

mikelehen (changes requested):
Nice! I think the caching logic isn't quite leveraging the cache the way it should be, but otherwise this looks great. Left a few other nits but nothing major (feel free to punt on the flatten refactor I alluded to!).

packages/metrics/src/data_providers/CanRegionDataProvider.test.ts
smcclure17: Yeah, I've renamed this to `CovidActNowDataProvider`. 

My thinking was we'd maybe want to ingest the bulk files somehow/at some point, but I agree it makes sense to build that into this provider if we ever feel that it's necessary.

packages/metrics/src/data_providers/CanRegionDataProvider.test.ts
smcclure17: Cool, yeah that makes sense--

I've updated this to just fall back to fetching non-timeseries data if the timeseries data doesn't exist

packages/metrics/src/data_providers/CanRegionDataProvider.ts
smcclure17: Ah, yeah it looks like it's not needed! 

I got [this error](https://github.com/covid-projections/act-now-packages/runs/8098484888?check_suite_focus=true) so I added it to see if it would fix it, but it seems to be working without it, anyway

packages/metrics/src/data_providers/CanRegionDataProvider.ts
smcclure17: Yeah, that's much more clean.

Initially, I wanted it to be separate/independent of the cache so that I had a way of testing the flattening logic, but since I've replaced that logic with lodash `get()` (thanks to your very helpful suggestions below!), that's no longer needed. So, I've updated this to store `data` directly in the cache 🙏

packages/metrics/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actnowcoalition/metrics",
-  "version": "0.0.9",
+  "version": "0.0.10",
```
mikelehen: Just FYI- I ended up bumping this in my PR, so you may want to merge develop and then bump this again.

packages/metrics/src/data_providers/CanRegionDataProvider.ts
smcclure17: I think `metricKey.split('.')` splits on every `.` (e.g. `"some.words.here".split('.') -> ["some", "words", "here"]`) but we need to retain all the `.`s besides the first one so that we have the correct dot-notation field id (because the fields inside the timeseries array can contain nesting)

packages/metrics/src/data_providers/CanRegionDataProvider.ts
smcclure17: Yeah, that's much simpler and saves a lot of imports--updated to use that 🙏 

It is still a little hacky, but I don't know how often we'll need to do something like this (parse a region of multiple potential types), so IDK if it's worth modifying the `Region` class over, but maybe something to keep track of.

packages/metrics/src/data_providers/CanRegionDataProvider.ts
smcclure17: oooh this is very nice! 🎉 Thanks for mentioning this + the examples 🙌 

I've updated this to use `get()`--I was able to just use it in `dataRowToMetricData()` instead of accessing the object keys directly (e.g. switching from `dataObject[key]` --> `get(dataObject, key)`), which allows us to be able to still only pass a string for the metric key and leave everything un-flattened.

Interesting about passing a function or string/key, I'll keep this in mind for potential future use 🙏

packages/metrics/src/data_providers/CanRegionDataProvider.ts
smcclure17: Ah, right, thanks! 😭 🙏

packages/metrics/src/data_providers/CovidActNowDataProvider.ts
```diff
@@ -0,0 +1,164 @@
...
+    metrics: Metric[],
+    includeTimeseries: boolean
+  ) {
+    for (const region of regions) {
+      const cacheKey = `${region.regionId}-${includeTimeseries}`;
+      const timeseriesCacheKey = `${region.regionId}-true`;
+      // If timeseries data exists in the cache, skip the fetch no matter what,
+      // as the timeseries endpoints also contain all of the non-timeseries data,
+      // and we will use this data if necessary via the getCachedData method.
+      if (!this.apiJson[cacheKey] && !this.apiJson[timeseriesCacheKey]) {
```
mikelehen: I think this code is correct but I had to read it twice to make sure.  I think it'd be simpler / shorter to just leverage our existing `getCachedData()` helper.

```
if (!getCachedData(region, includeTimeseries)) { 
  ...
```

packages/metrics/src/data_providers/CanRegionDataProvider.ts
mikelehen: Oh, right!  I missed that we're using the second half of the string below.  Makes sense!

packages/metrics/src/data_providers/CanRegionDataProvider.ts
mikelehen: Oh, right!  Very nice.  Yeah, this came out nicely.  Thanks for going ahead and reworking this.  I definitely prefer without the flattening logic. 😁

mikelehen (approved):
Couple minor things, but looks great.  I like the way this came out! 👏

packages/metrics/src/data_providers/CovidActNowDataProvider.ts
smcclure17: ah, yeah, much nicer! The existing logic is a little clunky as in half the cases `cacheKey` will equal `timeseriesCacheKey`

packages/metrics/package.json
smcclure17: oop, good catch thanks!

Issue #89, opened 2022-08-23
Write RollingAverageDataProvider


mikelehen:
This is a bit broader than just implementing RollingAverageDataProvider. I think we need to come up with a mechanism to define metric data providers that depend on other metrics (or regions).

In my original metrics refactor I had prototyped a "DerivedDataProvider" that would basically let you write a data provider that depends on other metrics.  A simple example is a RollingAverageDataProvider that provides a rolling average metric from a "raw" metric...   The definitions are here: https://github.com/covid-projections/who-demo/blob/82eddb59ba6237cce39ac0c3ce9a007f2ca3c582/src/metrics/data_providers.ts#L32

I think we want something similar, but I'm not sure exactly what it should look like.  Maybe we should just amend the MetricDataProvider.fetchData() function so that it receives the MetricCatalog as an additional argument.  That way it can fetch data for other metrics in the catalog if it wants / needs to.  I'm not sure.  Happy to brainstorm more on this.

mikelehen:
I added `TransformedMetricDataProvider` to act-now-packages and added a `RollingAverageMetricDataProvider` example to the template.

PR #110, opened 2022-08-30
Add TransformedMetricDataProvider.

mikelehen:
TransformedMetricDataProvider is a base class to help implement a MetricDataProvider that provides data by transforming data from another existing metric.  Implementers just need to set the `id` of the data provider and implement the `transformData()` method.

The idea is we can use this to create little data providers that compute daily new cases from cumulative cases, rolling averages, etc.  Example (note that it relies on a `Timeseries.rollingAverage()` method that's coming in a subsequent PR):

```
class RollingAverageMetricDataProvider extends TransformedMetricDataProvider {
  readonly id = "rolling-average";

  transformData(sourceData: MetricData<unknown>, newMetric: Metric, region: Region): MetricData<unknown> {
    const ts = sourceData.timeseries;
    const newTs = ts.rollingAverage({
      days: 7,
      treatMissingDatesAsZero: true
    });
    return new MetricData(newMetric, region, newTs.last?.value ?? null, newTs);
  }
}
```

Some additional incidental changes:
* Renamed MetricCatalog.fetchDataForMetricsAndRegions() to fetchDataForRegionsAndMetrics() for consistency.
* Made MetricCatalog pass itself to MetricDataProviders when fetching data so they can fetch
  dependent data if necessary.
* Added a MultiRegionMultiMetricDataStore.fromRegionsAndMetrics() helper for constructing a MultiRegionMultiMetricDataStore from regions,  metrics, and a MetricData getter callback.

Part of solving #89.


packages/metrics/src/MetricCatalog/MetricCatalog.test.ts
```diff
@@ -88,7 +88,7 @@ describe("MetricCatalog", () => {
     const catalog = new MetricCatalog(testMetricDefs, dataProviders);
 
     // Fetch data for two metrics that come from different data providers.
-    const dataStore = await catalog.fetchDataForMetricsAndRegions(
+    const dataStore = await catalog.fetchDataForRegionsAndMetrics(
```
mikelehen, line 91: I renamed this for consistency with other methods (e.g. useDataForRegionsAndMetrics) and to match the order of the parameters you pass.

packages/metrics/src/MetricCatalog/MetricCatalog.ts
```diff
@@ -186,7 +186,8 @@ export class MetricCatalog {
       const fetchedData = await provider.fetchData(
         regions,
         metrics,
-        includeTimeseries
+        includeTimeseries,
+        this
```
mikelehen, line 190: MetricCatalog now passes itself to `MetricDataProvider`s so they can fetch dependent metric data if necessary.

packages/metrics/src/data_providers/CachingMetricDataProviderBase.ts
```diff
@@ -63,19 +59,11 @@ export abstract class CachingMetricDataProviderBase
...
-      data[region.regionId] = new MultiMetricDataStore(region, regionData);
-    }
-
-    return new MultiRegionMultiMetricDataStore(data);
+    return MultiRegionMultiMetricDataStore.fromRegionsAndMetrics(
+      regions,
+      metrics,
+      (region, metric) =>
+        this.getDataFromCache(region, metric, includeTimeseries)
+    );
```
mikelehen, line 67: Refactored this to use the new helper.

packages/metrics/src/data_providers/MetricDataProvider.ts
```diff
@@ -31,6 +32,7 @@ export interface MetricDataProvider {
   fetchData(
     regions: Region[],
     metrics: Metric[],
-    includeTimeseries: boolean
+    includeTimeseries: boolean,
+    metricCatalog: MetricCatalog
```
mikelehen, line 40: MetricDataProviders now receive the MetricCatalog.

packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
```diff
@@ -50,6 +50,33 @@ export interface MultiRegionMultiMetricDataMap<T> {
 export class MultiRegionMultiMetricDataStore<T = unknown> {
   constructor(private data: MultiRegionMultiMetricDataMap<T>) {}
 
+  /**
```
mikelehen, line 53: This code used to live in `CachingMetricDataProvider` but I needed to do the same sort of thing from `MetricTransformDataProvider` too, so I refactored it into a shared helper.

packages/metrics/src/data_providers/TransformedMetricDataProvider.ts
```diff
@@ -0,0 +1,105 @@
...
+ * via the `sourceMetric` field of the `dataReference` field of the metric.
+ *
+ * @example
+ * ```typescript
+ * class RollingAverageMetricDataProvider extends TransformedMetricDataProvider {
+ *   readonly id = "rolling-average";
+ *
+ *   transformData(sourceData: MetricData<unknown>, newMetric: Metric, region: Region): MetricData<unknown> {
+ *     const ts = sourceData.timeseries;
+ *     const newTs = ts.rollingAverage({
```
mikelehen, line 26: rollingAverage() doesn't actually exist yet but is coming in another PR.

packages/metrics/src/data_providers/MetricDataProvider.ts
smcclure17: super small nit: should we add a `@param`/description for `metricCatalog` here?

smcclure17 (approved):
This looks really great! Very impressed with how easy it will be to add new data providers off of this 🙏 🙌

packages/metrics/src/data_providers/MetricDataProvider.ts
mikelehen: oops, yes!

PR #111, opened 2022-08-30
Add some Timeseries helpers for various calculations.

mikelehen:
Adds some methods to Timeseries for helpful calculations, most notably:
* computeDeltas() - Calculates the deltas between data points (e.g. to turn cumulative cases into daily new cases).
* windowed() - Windowing function that helps with aggregating data (e.g. for rolling average).
* rollingAverage() - Calculates the rolling average timeseries from a timeseries.

Note that:
1. I'm not 100% sure how much of this kind of logic we want to build into Timeseries.  At some point you should probably spin up `python` and `pandas` and build a "data pipeline" if you're doing a lot of computation.  But at the same time, it does seem like it's often handy to do basic transformations in the frontend.  So we'll see how this evolves.
2. The implementations and options provided are kinda' opinionated based on what I think we might want / need, but I'm not super confident so these may evolve over time.
3. I was torn about creating these as standalone functions (`rollingAverage(timeseries)`) vs methods (`timeseries.rollingAverage()`).  The latter is a lot more handy to use, so that's what I did, but it does potentially bloat the Timeseries class with code we might not need for every project.  So if it gets too bloated, we might want to refactor.
4. computeDeltas() and rollingAverage() only really make sense if the timeseries contains numbers, but you could technically call them on a `Timeseries<string>` or whatever.  This is a slight wart by putting them directly on the `Timeseries` class, but I think it's okay.

Part of solving #89.


packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -158,6 +174,29 @@ export class Timeseries<T = unknown> {
     return new Timeseries(this.points.filter(({ value }) => filterFn(value)));
   }
 
+  /**
```
mikelehen, line 177: Added this map() function to help a bit with implementing the other helpers.

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -167,12 +206,14 @@ export class Timeseries<T = unknown> {
...
-      this.points.map((p) => ({
-        date: p.date,
-        value: mapFn(p.value),
-      }))
-    );
+    return this.map((p) => ({
+      date: p.date,
+      value: mapFn(p.value),
+    }));
+  }
```
mikelehen, line 213: Refactored this to use the new map() method.

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -167,12 +206,14 @@ export class Timeseries<T = unknown> {
...
+
+  slice(start: number, end?: number): Timeseries<T> {
+    return new Timeseries(this.points.slice(start, end));
```
mikelehen, line 229: Mostly added this slice() method to make the window tests easier to write.

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -270,6 +311,17 @@ export class Timeseries<T = unknown> {
...
+  /**
+   * The first value in the timeseries, or undefined if the timeseries is
+   * empty.
+   *
+   * You can use {@link Timeseries.hasData} to guard against the timeseries
+   * being empty and ensure this can't return undefined.
+   */
+  get firstValue(): T | undefined {
+    return first(this.points)?.value;
+  }
```
mikelehen, line 336: Added firstValue / lastValue to clean up some examples / tests.

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -281,14 +333,25 @@ export class Timeseries<T = unknown> {
...
+
   /**
    * The minimum (earliest) date in the timeseries.
    *
    * You can use {@link Timeseries.hasData} to guard against the timeseries
    * being empty and ensure this can't return undefined.
    */
   get minDate(): Date | undefined {
-    return minBy(this.points, (p) => p.date)?.date;
+    return this.points[0]?.date;
```
mikelehen, line 367: Because we sort the timeseries in the constructor, we can just use the first / last dates as the min / max date.

packages/metrics/src/Timeseries/Timeseries.ts
```diff
@@ -321,6 +384,15 @@ export class Timeseries<T = unknown> {
...
   }
 
+  /**
+   * Returns the sum of all numeric values in the timeseries.
+   */
+  get sum(): number {
+    return sumBy(this.points, (p) =>
+      typeof p.value === "number" ? p.value : 0
+    );
+  }
```
mikelehen, line 407: This was handy for implementing rolling average.

packages/metrics/src/Timeseries/Timeseries.test.ts
```diff
@@ -198,7 +198,7 @@ describe("Timeseries", () => {
       expect(
         new Timeseries([
           { date: new Date("2021-02-03"), value },
-        ]).assertBoolean().last?.value
+        ]).assertBoolean().lastValue
```
mikelehen, line 201: Added `lastValue` and `firstValue` helpers for convenience.

packages/metrics/src/Timeseries/timeseries_utils.ts
```diff
@@ -1,78 +1,4 @@
-import findLastIndex from "lodash/findLastIndex";
-
-import { assert } from "@actnowcoalition/assert";
-
-import { Timeseries, TimeseriesPoint } from "./Timeseries";
-
-/**
```
mikelehen, line 7: This was an old `rollingAverage()` implementation that only worked on timeseries without gaps.  The new version is more flexible so I'm deleting this one.

packages/metrics/src/Timeseries/Timeseries.ts
smcclure17: should we add a description/documentation for this method here?

smcclure17 (approved):
This looks great to me!! 🙌 

> computeDeltas() and rollingAverage() only really make sense if the timeseries contains numbers, but you could technically call them on a Timeseries<string> or whatever.

Would it make sense to call `assertFiniteNumbers()` in those methods to make it explicit that these only work for `Timeseries<number>`?  I'm not sure if that would add complexity, or if it has any real benefit considering it will (I think) fail anyway if they are called on a `Timeseries<string>`

packages/metrics/src/Timeseries/Timeseries.ts
mikelehen: Ooops, yes! Thanks for catching / flaging.

mikelehen:
> Would it make sense to call assertFiniteNumbers() in those methods to make it explicit that these only work for Timeseries<number>? I'm not sure if that would add complexity, or if it has any real benefit considering it will (I think) fail anyway if they are called on a Timeseries<string>

It looks like I was doing that (`.removeNils().assertFiniteNumbers()`) in `computeDeltas()` but not `rollingAverage()`.  I think was feeling hesitant to do it because I am a bit worried about the wasted performance of calling those over and over.  But I think it is better / cleaner so I went ahead and did it.  I may look into making `Timeseries` "remember" if those methods have already been called and turning subsequent ones into no-ops, so I can stop worrying about the performance impact.

PR #113, opened 2022-08-30
Create grid component

### LINKED ISSUE #64: Implement GridLines
Implement components `GridRows` and `GridColumns` based of `@visx/grid`, but adapting the style to match our theme variables.

Image

[Figma](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=72%3A2892)

faihegberg:
Close #64


packages/ui-components/src/components/Grid/Grid.stories.tsx
```diff
@@ -0,0 +1,54 @@
+import React from "react";
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { scaleLinear } from "@visx/scale";
+import { GridRows, GridColumns } from "./Grid.style";
+import { theme } from "../../styles";
+
+export default {
+  title: "Charts/Grid",
+} as ComponentMeta<typeof GridRows>;
```
pnavarrc, line 9: Same as before, consider not passing the type if the stories are for more than one component (or take a look at https://storybook.js.org/docs/react/writing-stories/stories-for-multiple-components - maybe there is a better pattern that we haven't found yet? I haven't read the article myself though)

The problem with passing the type of only one of the components is that the docs tab in Storybook is going to show the types of one component only.

packages/ui-components/src/components/Grid/Grid.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+  >
+    <GridColumns {...args} />
+  </svg>
+);
+
+export const Columns = TemplateColumns.bind({});
+Columns.args = {
+  scale: yScale,
+  height,
+};
```
pnavarrc, line 51: Consider adding examples that include the Axis component so we can confirm that they align when using the same scales. Also, note that you can use args from other stories to avoid re-creating the parameters each time

```tsx
import { BottomAxis } from "../Axis/Axis.stories.tsx";

// ...
Columns.args = {
  scale: BottomAxis.args.scale,
  // ...
}
```

There might be some examples in the GCAT repo

packages/ui-components/src/components/Grid/Grid.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const TemplateColumns: ComponentStory<typeof GridColumns> = (args) => (
+  <svg
+    width={width}
+    height={height}
+    style={{ border: `1px solid ${theme.palette.border.default}` }}
+  >
+    <GridColumns {...args} />
+  </svg>
+);
```
pnavarrc, line 45: Why are the columns only going so far?

image

packages/ui-components/src/components/Grid/Grid.style.tsx
```diff
@@ -0,0 +1,18 @@
...
+} from "@visx/grid";
+import React from "react";
+
+export type GridRowsProps = React.ComponentProps<typeof VxGridRows>;
+
+export type GridColumnsProps = React.ComponentProps<typeof VxGridColumns>;
+
+export const GridRows = styled((props: GridRowsProps) => (
+  <VxGridRows stroke={theme.palette.border.default} {...props} />
+))``;
```
pnavarrc: It might be better to customize using styled instead, that way the theme will be taken from the ThemeProvider context instead of being taken from the imported theme variable

```tsx
export const GridRows = styled((props: GridRowsProps) => (
  <VxGridRows {...props} />
))`
line.visx-line {
  stroke: ${({theme}) => theme.palette.border.default};
}
`;
```

This is also true for other components, but trickier to implement in some cases.

pnavarrc (commented):
Looking pretty good, added some notes. Ping me when you need a second pass!

packages/ui-components/src/components/Grid/Grid.stories.tsx
faihegberg: Looks like the doc is geared towards storybook displaying multiple components which are _nested_ with each other. I'll remove the type declaration as you suggested though.

packages/ui-components/src/components/Grid/Grid.stories.tsx
faihegberg: As discussed in our call, because the arguments for the row and column templates were backwards 😅 Fixing!

pnavarrc (commented):
Looks good, remember to merge develop in and to increase the version number before publishing!

pnavarrc (approved):

PR #115, opened 2022-08-30
Enable Markdown and USNationalMap components.

### LINKED ISSUE #51: Markdown bug
_Description in progress_

When we try to use the Markdown component using the template (or hackathon) repository, Next.js throws an error (ESM). It looks like this error is caused by Next not handling ESM-only packages correctly (react-markdown is ESM-only).

We need to find a workaround to be able to implement and use the Markdown component.

# Notes

- https://github.com/vercel/next.js/issues/30330#issuecomment-952847838

mikelehen:
Fixes #51

### DIFF
--- a/packages/ui-components/package.json
+++ b/packages/ui-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@actnowcoalition/ui-components",
-  "version": "1.5.6",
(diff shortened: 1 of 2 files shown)


pnavarrc (approved):
Thanks @mikelehen !

PR #116, opened 2022-08-30
Setup `MetricCatalogProvider` and `useMetricContext` hook

pnavarrc:
Implements the `MetricCatalogProvider` component and corresponding `useMetricContext` hook. I added some stories to demonstrate how to use the provider and hook, and added a decorator so the metric catalog context is available for Storybook.

### DIFF
--- a/packages/ui-components/src/components/MetricCatalogContext/MetricCatalogContext.tsx
+++ b/packages/ui-components/src/components/MetricCatalogContext/MetricCatalogContext.tsx
@@ -1,0 +1,68 @@
+import { MetricCatalog } from "@actnowcoalition/metrics";
+import React, { createContext, useContext } from "react";
+
+const defaultMetricCatalog = new MetricCatalog([], []);
+
+const MetricCatalogContext = createContext<MetricCatalog>(defaultMetricCatalog);
+
+/**
+ * The `MetricCatalogProvider` component and `useMetricCatalog` hook provide
+ * convenient  access to the app-level `metricCatalog` instance. It relies on
+ * React context, so make sure that `MetricCatalogProvider` is a parent of
+ * the component that calls `useMetricCatalog`.
+ *
+ * The best way to do this in Next.js applications is to wrap the page
+ * component in _app.tsx.
+ *
+ * @example
+ * ```tsx
+ * // _app.tsx
+ * import { metricCatalog } from "src/common/metrics"
+ *
+ * // ...
+ * <ThemeProvider theme={theme}>
+ *   <MetricCatalogProvider metricCatalog={metricCatalog}>
+ *     <Component {...pageProps} />
+ *   </MetricCatalog>
+ * </ThemeProvider>
+ * ```
+ *
+ * Any component that wants to use the metric catalog will usually need to have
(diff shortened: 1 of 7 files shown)


packages/ui-components/src/components/MetricCatalogContext/MetricCatalogContext.tsx
```diff
@@ -0,0 +1,68 @@
...
+ * interface MetricAwareProps {
+ *   region: Region;
+ *   metrics: Metric[];
+ * };
+ *
+ * const MetricAware: React.FC<MetricAwareProps> = ({ region, metrics }) => {
+ *    const metricCatalog = useMetricCatalog();
+ *    const { data, error } = metricCatalog.useDataForMetrics(region, metrics);
+ *   // ...render component
+ * ```
```
mikelehen, line 45: Lovely docs, thank you!

packages/ui-components/src/index.ts
```diff
@@ -91,3 +91,8 @@ export type { AxisLeftProps, AxisBottomProps } from "./components/Axis";
 // export { default as Markdown } from "./components/Markdown";
 // Note (8/29/22) - Same problem with d3-geo, commenting out until fixed.
 // export { default as USNationalMap } from "./components/USNationalMap";
+
+export {
+  MetricCatalogProvider,
+  useMetricCatalog,
```
mikelehen: nit: Any reason not to just do `export * from ...` ?

I feel like explicitly exporting them from one-level is enough.  Exporting them explicitly at every level just feels cumbersome. 😁   But I don't have strong feelings.

packages/ui-components/src/stories/mockMetricCatalog.ts
```diff
@@ -0,0 +1,62 @@
...
+const testMetricDefs: MetricDefinition[] = [
+  {
+    id: MetricId.PI,
+    name: "Pi",
+    extendedName: "Pi - The ratio of a circle's circumference to its diameter",
+    dataReference: {
+      providerId: "static",
+      value: 3.141592653589793,
+    },
+    formatOptions: { minimumFractionDigits: 5 },
```
mikelehen: Just checking, would maximum be more interesting than minimum?

packages/ui-components/src/stories/mockMetricCatalog.ts
```diff
@@ -0,0 +1,62 @@
...
+    name: "Cases Per 100k (mock)",
+    extendedName: "Cases per 100k population (using mock data)",
+    dataReference: {
+      providerId: "mock",
+      startDate: "2020-01-01",
+    },
+  },
+];
+
+export const metricCatalogB = new MetricCatalog(metricDefsB, dataProviders);
```
mikelehen, line 64: nit: Since this will only ever be relevant to one story (MetricCatalogContext.stories.tsx), I would maybe move it there?

As-is, I could see somebody adding more metrics to this file for the purposes of testing components and get confused about which (or both) catalog they need to add them to.  Not a huge deal though if you aren't a fan.

mikelehen (commented):
Awesome!  Couple optional nits, but this looks good!

mikelehen (approved):

PR #119, opened 2022-08-31
USNationalMap using canvas

chasulin:
Re-implements `USNationalMap` using canvas to improve performance


packages/ui-components/src/components/Maps/NewUSNationalMap/StatesMap.tsx
```diff
@@ -0,0 +1,47 @@
...
+    <svg width={width} height={height}>
+      {statesGeographies.features.map((geo) => {
+        const stateFips = `${geo.id}`;
+        const state = states.findByRegionIdStrict(stateFips);
+        return (
+          <Tooltip key={stateFips} title={renderTooltip(stateFips)}>
+            <g>
+              <a href={state.relativeUrl}>
+                {!showCounties && (
+                  // @ts-ignore (Chelsi) - figure out why 'd' prop is being problematic
```
pnavarrc: I think `d` is fine, but TS doesn't trust that `geoPath(geo)` will return a string, maybe doing `d={geoPath(geo) ?? ""}` will solve the problem.

packages/ui-components/src/components/Maps/NewUSNationalMap/StatesMap.tsx
```diff
@@ -0,0 +1,47 @@
...
+                  <StateShape d={geoPath(geo)} fill={getFillColor(stateFips)} />
+                )}
+                {/* @ts-ignore (Chelsi) - figure out why 'd' prop is being problematic */}
+                <StateOverlay data-track="map-state" d={geoPath(geo)} />
```
pnavarrc: I think the data-track thing is not useful anymore

packages/ui-components/src/components/Maps/NewUSNationalMap/StatesMap.tsx
```diff
@@ -0,0 +1,48 @@
...
+}) => {
+  return (
+    <svg width={width} height={height}>
+      {statesGeographies.features.map((geo) => {
+        const stateFips = `${geo.id}`;
+        const state = states.findByRegionIdStrict(stateFips);
+        return (
+          <Tooltip key={stateFips} title={renderTooltip(stateFips)}>
+            <g>
+              <a href={state.relativeUrl}>
```
pnavarrc: states shouldn't have `relativeUrl` anymore, that should come from the app-level `getRegionUrl` function. Also, the `a` element could be `next/link`, maybe it should also come from the app? we can punt for later too

packages/ui-components/src/components/Maps/NewUSNationalMap/StatesMap.tsx
```diff
@@ -0,0 +1,48 @@
...
+import Tooltip from "@mui/material/Tooltip";
+import { StateOverlay, StateShape } from "./NewUSNationalMap.style";
+import { statesGeographies } from "../../../common/geo-shapes";
+import { states } from "@actnowcoalition/regions";
+
+const StatesMap: React.FC<{
+  width: number;
+  height: number;
+  geoPath: GeoPath;
+  renderTooltip: (fips: string) => React.ReactElement | string;
```
pnavarrc: Consider renaming `fips` -> `regionId`

pnavarrc (approved):
Looks good in general, we need to make the `StateMap` a bit more flexible so it can use `next/link` and the app-level `getRegionRelativeUrl` instead of `region.relativeUrl`

PR #128, opened 2022-08-31
Implement MetricDot component

### LINKED ISSUE #121: MetricDot
Implement the metric dot component. See [ActNow Design System](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System) and [Figma - Monkeypox](https://www.figma.com/file/OG0YkSXfO6YNPfq21jRq42/Monkeypox) for examples in context

Image

pnavarrc:
Close #121 

image

## Changes

- Implement MetricDot component
- Update @actnowcoalition/metrics to the latest version so we can use `metric.getColor`
- Drive-by: Declare and export the `LabelIconProps` interface

## Notes

- We need to do a bit of research before we make this component accessible, so punting on that for now.

### DIFF
--- a/packages/ui-components/src/components/MetricDot/MetricDot.tsx
+++ b/packages/ui-components/src/components/MetricDot/MetricDot.tsx
@@ -1,0 +1,45 @@
+import React from "react";
+import { Region } from "@actnowcoalition/regions";
+import { Metric } from "@actnowcoalition/metrics";
+import { useMetricCatalog } from "../MetricCatalogContext";
+import { Dot, PlaceholderDot } from "./MetricDot.style";
+
+export interface MetricDotProps {
+  /** Region for which we want to represent the current level */
+  region: Region;
+  /** Metric for which we want to represent the current level */
+  metric: Metric | string;
+}
+
+/**
+ * The MetricDot component shows a colored dot that represents current level
+ * or category for a given metric and region. The component is still rendered
+ * for metrics that don't have levels or categories to keep spacing and
+ * alignment consistent across metrics.
+ */
+const MetricDot: React.FC<MetricDotProps> = ({
+  region,
+  metric: metricOrId,
+}) => {
+  const metricCatalog = useMetricCatalog();
+  // TODO: Maybe add a method to the metricCatalog?
+  const metric =
+    typeof metricOrId === "string"
+      ? metricCatalog.getMetric(metricOrId)
+      : metricOrId;
+
+  if (!metric.levelSet || !metric.categories) {
+    return <PlaceholderDot />;
+  }
+
+  const { data, error } = metricCatalog.useData(region, metric);
+  if (error || !data) {
+    return <PlaceholderDot />;
+  }
+
+  // TODO: Find out accessibility best practices for this component
+  const color = metric.getColor(data.currentValue);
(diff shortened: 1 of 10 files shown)


packages/ui-components/src/components/MetricDot/MetricDot.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export interface MetricDotProps {
+  /** Region for which we want to represent the current level */
+  region: Region;
+  /** Metric for which we want to represent the current level */
+  metric: Metric | string;
+}
+
+/**
+ * The MetricDot component shows a colored dot that represents current level
```
mikelehen: nit: "represents current" => "represents the current"

packages/ui-components/src/components/MetricDot/MetricDot.tsx
```diff
@@ -0,0 +1,45 @@
...
+ * or category for a given metric and region. The component is still rendered
+ * for metrics that don't have levels or categories to keep spacing and
+ * alignment consistent across metrics.
+ */
+const MetricDot: React.FC<MetricDotProps> = ({
+  region,
+  metric: metricOrId,
+}) => {
+  const metricCatalog = useMetricCatalog();
+  // TODO: Maybe add a method to the metricCatalog?
```
mikelehen, line 25: I've been thinking this too.  I think maybe I'll just make `getMetric()` accept `string|Metric`

packages/ui-components/src/components/MetricDot/MetricDot.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const metric =
+    typeof metricOrId === "string"
+      ? metricCatalog.getMetric(metricOrId)
+      : metricOrId;
+
+  if (!(metric.levelSet || metric.categories)) {
```
mikelehen, line 31: I wonder if we should have a `metric.isColorable` property or maybe make `metric.getColor()` have some "default" behavior (return "transparent"?) or something.  But this is fine for now.

packages/ui-components/src/stories/mockMetricCatalog.ts
```diff
@@ -28,14 +29,31 @@ const testMetricDefs: MetricDefinition[] = [
...
 ];
-export const metricCatalog = new MetricCatalog(testMetricDefs, dataProviders);
+
+const metricLevelSets = [
+  {
+    id: "cases_mock",
+    levels: [
+      { id: "low", color: theme.palette.severity[100] },
+      { id: "medium", color: theme.palette.severity[200] },
+      { id: "high", color: theme.palette.severity[500] },
```
mikelehen: nit: Consider adding `name` like we talked about?  You don't need it for `MetricDot` but we probably will for other components.

mikelehen (approved):
Nice! Couple optional nits, but looks good!

packages/ui-components/src/components/MetricDot/MetricDot.tsx
pnavarrc: I think `metric.isColorable` might be better, transparent kind of works in this case but probably not in general. In some cases we will need to know about levels and categories (to be able to render the legend for the metric)

PR #129, opened 2022-08-31
Implement MetricValue component

### LINKED ISSUE #127: Implement MetricValue component
[Figma - MetricValue](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=8%3A2798)

Image

pnavarrc:
Close #127 

image

## Changes

- Implement `MetricValue` component

### DIFF
--- a/packages/ui-components/src/components/MetricValue/MetricValue.tsx
+++ b/packages/ui-components/src/components/MetricValue/MetricValue.tsx
@@ -38,7 +38,7 @@ const MetricValue: React.FC<MetricValueProps> = ({
     <Stack direction="row" spacing={1} alignItems="center">
       <MetricDot region={region} metric={metric} />
       <Typography variant={variant}>
-        {metric.formatValue(data?.currentValue, "---")}
+        {metric.formatValue(data.currentValue, "---")}
       </Typography>
     </Stack>
   );
--- a/packages/ui-components/src/components/MetricValue/MetricValue.tsx
+++ b/packages/ui-components/src/components/MetricValue/MetricValue.tsx
@@ -1,0 +1,47 @@
+import React from "react";
+import { Stack, Typography, TypographyProps } from "@mui/material";
+import { Metric } from "@actnowcoalition/metrics";
+import { Region } from "@actnowcoalition/regions";
+import { useMetricCatalog } from "../MetricCatalogContext";
+import MetricDot from "../MetricDot";
+
+export interface MetricValueProps {
+  /** Region for which we want to show the metric value */
+  region: Region;
+  /** Metric for which we want to show the metric value  */
+  metric: Metric | string;
+  /**
+   * Typography variant used to render the metric value. We use
+   * dataEmphasizedLarge by default.
+   * */
+  variant?: TypographyProps["variant"];
+}
+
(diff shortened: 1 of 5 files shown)


packages/ui-components/src/components/MetricValue/MetricValue.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  if (!data || error) {
+    return <Typography variant={variant} />;
+  }
+
+  return (
+    <Stack direction="row" spacing={1} alignItems="center">
+      <MetricDot region={region} metric={metric} />
+      <Typography variant={variant}>
+        {metric.formatValue(data?.currentValue, "---")}
```
mikelehen: You shouldn't need `?` after `data` (you handle `!data` above), right?

packages/ui-components/src/components/MetricValue/MetricValue.tsx
```diff
@@ -0,0 +1,47 @@
...
+  metric: metricOrId,
+  variant = "dataEmphasizedLarge",
+}) => {
+  const metricCatalog = useMetricCatalog();
+  const metric =
+    typeof metricOrId === "string"
+      ? metricCatalog.getMetric(metricOrId)
+      : metricOrId;
+
+  const { data, error } = metricCatalog.useData(region, metric);
```
mikelehen, line 31: FWIW it occurs to me that we could move the MetricCatalogContext stuff into the metrics package and then have the metrics package export a `useData()` hook that internally calls `useMetricCatalog()`.  That would save a line of code in all of these components. 🤷

packages/ui-components/src/index.ts
```diff
@@ -93,6 +93,7 @@ export { default as Markdown } from "./components/Markdown";
 export * from "./components/Maps";
 export { default as LabelIcon } from "./components/LabelIcon";
 export type { LabelIconProps } from "./components/LabelIcon";
-
 export { default as MetricDot } from "./components/MetricDot";
 export type { MetricDotProps } from "./components/MetricDot";
+export { default as MetricValue } from "./components/MetricValue";
+export type { MetricValueProps } from "./components/MetricValue";
```
mikelehen, line 99: Not a big deal, and I won't mention it again since I think I did on a previous code review, but just flagging that we could do `export * from ...` and save some repetition.

mikelehen (approved):
Looks good!

packages/ui-components/src/index.ts
pnavarrc: Actually that's a great idea, let's discuss it tomorrow during standup to make it a standard thing! 🙌

packages/ui-components/src/components/MetricValue/MetricValue.tsx
pnavarrc: I thought about that as well, but it will add react as dependency to the metrics package, not sure if we want to do that

mikelehen: react actually already is a dependency, since `metricCatalog.useData()` is built with react hooks already...

But arguably maybe all of the hook logic should be pulled over to ui-components instead?  Not sure!  Maybe another standup discussion topic.

PR #130, opened 2022-09-01
Implement `OptionSelector` component

### LINKED ISSUE #122: Implement Options component
Implement the metric dot component. See [ActNow Design System](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System) and [Figma - Monkeypox](https://www.figma.com/file/OG0YkSXfO6YNPfq21jRq42/Monkeypox) for examples in context

Image

pnavarrc:
Close #122 - Implement OptionSelector

**Update**: After the code review and some thinking, I decided to just style ToggleButton and ToggleButtonGroup and not implementing the OptionSelector component. Most likely the parent component will need to manage the state, so creating a component that manages the state internally creates unnecessary complexity. See https://github.com/covid-projections/act-now-packages/pull/130#issuecomment-1234804225

---

[Figma - Option Selector](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=95%3A4197)

image

image

## Notes

- I didn't implement the vertical version following the mocks because it feels a bit inconsistent when compared with the horizontal option and also because we haven't used that version so far. Will punt on that until we can sync with Josh. I did customize the component enough to make it look okay though.
- This component is basically `ToggleButton`, but is a bit more convenient and manages the state internally.


packages/ui-components/src/components/OptionSelector/OptionSelector.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export const WithIcons = Template.bind({});
+WithIcons.args = {
+  "aria-label": "Alignment",
+  options: iconOptions,
+  selectedOptionId: "center",
+  size: "small",
+  onChange: (event: React.MouseEvent<HTMLElement>, optionId: string) => {
+    console.log(`Selected ${optionId}`);
+  },
```
mikelehen: nit: would it make sense to put the onChange handler in the DefaultProps?

packages/ui-components/src/components/OptionSelector/OptionSelector.tsx
```diff
@@ -0,0 +1,96 @@
...
+ * the user can select only one of them.
+ *
+ * Accessibility Notes
+ *
+ * The `OptionSelector` component should have an accessibility label attached
+ * to it, use either `aria-label` or `aria-labelledby` (if the label is provided
+ * on a different element).
+ * Each option should have a label. If `buttonContent` doesn't contain text
+ * (when using only an icon, for example), make sure to provide `ariaLabel` to
+ * make the component accessible.
```
mikelehen: Good call, with the accessibility notes!

packages/ui-components/src/components/OptionSelector/OptionSelector.tsx
```diff
@@ -0,0 +1,96 @@
...
+ */
+const OptionSelector: React.FC<OptionSelectorProps> = ({
+  options,
+  selectedOptionId,
+  onChange,
+  ...otherToggleButtonProps
+}) => {
+  assert(options.length > 0, "OptionSelector must have at least one option");
+  const [firstOption] = options;
```
mikelehen: Feel free to disregard since this is 100% stylistic, but this syntax is less intuitive to me than just `options[0]`.  And since we only use `firstOption` once, I would probably just do `options[0].id` below, which seems pretty straightforward / readable to me.  Up to you though!

packages/ui-components/src/components/OptionSelector/OptionSelector.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const [optionId, setOptionId] = useState(firstOption.id);
+
+  // If the user pass a selectedIOptionId, we update the state to select the
```
mikelehen: selectedIOptionId => selectedOptionId

packages/ui-components/src/components/OptionSelector/OptionSelector.tsx
```diff
@@ -0,0 +1,96 @@
...
+}
+
+export interface OptionSelectorProps extends ToggleButtonGroupProps {
+  /** List of options to choose from */
+  options: Option[];
+  /**
+   * Optional ID of the initially selected option. The first option will be
+   * selected by default
+   */
+  selectedOptionId?: string;
```
mikelehen: Perhaps rename to `defaultOptionId` since this won't reflect the actual selected option after the user manually clicks on an option, right?  It's only used before the user interacts?

packages/ui-components/src/components/OptionSelector/OptionSelector.tsx
```diff
@@ -0,0 +1,96 @@
...
+  ...otherToggleButtonProps
+}) => {
+  assert(options.length > 0, "OptionSelector must have at least one option");
+  const [firstOption] = options;
+  const [optionId, setOptionId] = useState(firstOption.id);
+
+  // If the user pass a selectedIOptionId, we update the state to select the
+  // corresponding item. This effect will run only on mount/unmount because
+  // we are passing an empty dependencies array. See
+  // https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects
```
mikelehen: Hrm.  This feels kind of hacky and I'm not sure why we're doing it?  Does it cause a bug if we include `selectedIOptionId` in the effect dependencies?

All else being equal, I think ideally we *should* re-render if `selectedIOptionId` changes, at least if the user hasn't already interacted with the UI (if they have, then it's debatable).

One way to model the state of this component which might help would be:
1. Rename `optionId` to `userSelectedOptionId` and have it be undefined by default. We'll only set it when the user actually clicks on an option.
2. Have this component do something like `const selectedOptionId = userSelectedOptionId ?? defaultOptionId;`

If we do that, then I think we don't need this effect at all?  And the component will automatically update if defaultOptionId changes before the user has selected an option (which might be the behavior we want?  Not sure.).

In general I'm wary of breaking the rules of hooks and skipping dependencies, etc.  I think that will usually result in components that don't behave quite normally.

packages/ui-components/src/components/OptionSelector/OptionSelector.tsx
```diff
@@ -0,0 +1,96 @@
...
+  }, []);
+
+  const handleChangeOption = (
+    event: React.MouseEvent<HTMLElement>,
+    newOption: string
+  ) => {
+    setOptionId(() => {
+      onChange && onChange(event, newOption);
+      return newOption;
+    });
```
mikelehen: How is this different / better than:

```
onChange && onChange(event, newOption);
setOptionId(newOption);
```
?

packages/ui-components/src/components/OptionSelector/OptionSelector.tsx
```diff
@@ -0,0 +1,96 @@
...
+ * Accessibility Notes
+ *
+ * The `OptionSelector` component should have an accessibility label attached
+ * to it, use either `aria-label` or `aria-labelledby` (if the label is provided
+ * on a different element).
+ * Each option should have a label. If `buttonContent` doesn't contain text
+ * (when using only an icon, for example), make sure to provide `ariaLabel` to
+ * make the component accessible.
+ */
+const OptionSelector: React.FC<OptionSelectorProps> = ({
```
mikelehen: FWIW- This name feels too generic to me.  When I think of "OptionSelector" I think of the `<select>` tag or a dropdown list or something, whereas this component is a more specific visual than that (using toggle buttons).  I am not sure what to call it though.  ToggleOptions? ToggleSelector?   Idunno.  Feel free to leave it as-is if you're not wild about changing the name.

packages/ui-components/src/styles/theme/components.ts
```diff
@@ -40,6 +40,38 @@ const components: ThemeOptions["components"] = {
...
+  MuiToggleButtonGroup: {
+    styleOverrides: {
+      root: {
+        ...typography.paragraphSmall,
+        border: `solid 1px ${palette.border.default}`,
+      },
+      groupedHorizontal: {
+        border: "none",
+        color: typography.paragraphSmall.color,
+        textTransform: "none",
```
mikelehen, line 52: I feel like border / color / textTransform could probably be moved to "root" and shored with vertical orientation too.

packages/ui-components/src/styles/theme/components.ts
```diff
@@ -40,6 +40,38 @@ const components: ThemeOptions["components"] = {
...
+        borderBottom: `solid 2px transparent`,
+        ":not(:first-of-type)": {
+          borderLeft: "none",
+        },
```
mikelehen, line 56: I think this isn't necessary since we're setting border to none already?

packages/ui-components/src/styles/theme/components.ts
```diff
@@ -40,6 +40,38 @@ const components: ThemeOptions["components"] = {
...
+      },
+      groupedVertical: {
+        borderLeft: "none",
+        borderRight: "none",
+        ":first-of-type": {
+          borderTop: "none",
+        },
+        ":last-of-type": {
+          borderBottom: "none",
+        },
```
mikelehen, line 75: Per my other note, why not just drop the borders entirely for consistency with horizontal?  Don't feel strongly though.  🤷 

image

mikelehen (changes requested):
Generally looks good but a few nits and a question about the effect dependencies thing.

> Will punt on that until we can sync with Josh.

Do we have a good way to keep track of things we need Josh to be aware of / weigh in on?  A github/trello board or a running doc or something?  Just feels like we could have friction later if there's a divergence from the mocks and the eng implementations that isn't clearly communicated.

packages/ui-components/src/components/OptionSelector/OptionSelector.tsx
pnavarrc: Makes sense 👍

packages/ui-components/src/styles/theme/components.ts
pnavarrc: I tried that, but it doesn't really work. The `root` styles apply to the underlying div that represents the `ToggleButtonGroup`; the styles under `grouped`, `groupedHorizontal` and ``groupedVertical` style the `ToggleButton` items under it. Maybe I can consolidate a little but some of the styles need a certain level of specificity.

packages/ui-components/src/styles/theme/components.ts
pnavarrc: It is necessary because the default styles are set using the `:not(:first-of-type)` pseudo-class, so they have higher specificity than styles set on the default class

packages/ui-components/src/styles/theme/components.ts
pnavarrc: Maybe! I feel like it will need a 2px border (let's say on the right side) to be visually closer to the horizontal version.

packages/ui-components/src/styles/theme/components.ts
mikelehen: Oh, I see.  Yeah, I misunderstood how this works. 😬 

I guess I'd kinda' advocate for duplicating these to the vertical one then, for consistency?  But I don't feel super strongly, especially since we aren't trying to match the vertical mocks or anything.

packages/ui-components/src/styles/theme/components.ts
mikelehen: Oh, hrm.  I see...  I guess there's a 1px transparent border that we are overriding or something.

How do you figure that sort of thing out?  Is it documented?  Or you're just poking around in dev tools to figure out what you need to override?

pnavarrc:
Thanks for the review and all the feedback @mikelehen! After giving it some thought, I think that this component might be a bit redundant and that customizing `ToggleButtonGroup` and `ToggleButton` might be enough.

The main use case that I wanted to cover is allowing client code to have an arbitrary option selected by default. This should be easy to do by passing the `defaultOptionId` as prop and using it to set the initial state, but the prop and the internal state can get out of sync (the prop will be ignored to favor the internal state).

One solution is to use the `useEffect(() => setSelectedOption(defaultOptionId), [])` pattern that runs only on mount, which is what I did here. The pattern of using `useEffect(() => {}, [])` to apply the effect only once is fairly well known, but triggers the ESLint rule.

What's making me think that this component might not be worth the trouble is that we need state management on the parent component anyways, because we want to change something else when the user selects an option. For example, if we use it with the US map component, we will have something like this

```tsx
// 
enum LocationType {
  STATES = "states",
  COUNTIES = "counties"
}
const [locationType, setLocationType] = useState(LocationType.STATES);
const options = [
  { id: LocationType.STATES, buttonContent: "States" },
  { id: LocationType.COUNTIES, buttonContent: "Counties" },
];

<OptionSelector
  options={options}
  defaultOptionId={locationType}
  onChange={(e, value) => setLocationType(value)}
/>

<USMap
  metric={metric}
  locationType={locationType}
/>
```

The `OptionSelector` component makes the code a bit more concise, but the parent will still need to have a state, and we need to update the parent state each time that `OptionSelector` changes its internal state. Adding the `defaultOptionId` to the mix makes things a bit more complicated and both states might get out of sync.

Using `ToogleButtonGroup` and `ToggleButton` directly might be more explicit and reduce the chances of making a mistake managing the state, and makes it easier to pass other props to `ToggleButton` without extending the `Option` interface (`disabled`, `color`, etc)

```tsx
<ToggleButtonGroup exclusive value={locationType}  
  onChange={(e, value) => setLocationType(value)}
>
  <ToggleButton value={LocationType.STATES}>States</ToggleButton> 
  <ToggleButton value={LocationType.COUNTIES}>Counties</ToggleButton> 
</ToggleButtonGroup>

<USMap
  metric={metric}
  locationType={locationType}
/>
```

So my proposal is not to remove `OptionSelector` and add stories for `ToggleGroupButton` and `ToggleGroup`. Thoughts? cc @mikelehen

mikelehen:
@pnavarrc Ah, that's an excellent point!  Yeah, that makes total sense to me.  If we find using ToggleGroup* directly ends up cumbersome / boilerplate-y, we can go back and introduce this component again once it's clear what benefit it will provide.

pnavarrc:
Removed the OptionSelector component and added stories for MUI's `ToggleButton` and `ToggleButtonGroup` as discussed.

packages/ui-components/src/styles/theme/components.ts
faihegberg: I generally agree with Michael here. I think not having a border on the vertical version looks better (subjective opinion of course). So I don't feel strongly.

faihegberg (approved):
image
The size of the text toggle button group seems to change sightly when one toggles it. Not a biggie but it might be good to see if we can keep the size fixed?

pnavarrc:
The borders between items are removed in the vertical version - going to keep the font-size/weight for now and wait for Josh to be back to confirm / improve on the vertical version. Thanks for the review @faihegberg !

PR #131, opened 2022-09-01
`MetricLegendCategorical` component

### LINKED ISSUE #124: Implement MetricLegendCategorical component
See [Figma - Metric Legend Categorical](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=13%3A948)

Image

smcclure17:
Close #124 

Example: 
image


yarn.lock
```diff
@@ -3625,6 +3625,20 @@
     "@types/d3-shape" "^1.3.1"
     d3-shape "^1.0.6"
 
+"@visx/grid@^2.12.2":
```
smcclure17: Not sure what prompted this change--I can revert it if we'd like?

packages/ui-components/src/components/MetricLegendCategorical/MetricLegendCategorical.tsx
```diff
@@ -0,0 +1,59 @@
...
+const getItemLabel = (item: MetricCategory) => item.label;
+
+const MetricLegendCategorical = ({
+  metric,
+  supportingText,
+  orientation,
+  titleVariant = "dataEmphasizedSmall",
+}: MetricLegendCategoricalProps) => {
+  const metricCatalog = useMetricCatalog();
+  const resolvedMetric =
```
smcclure17: this can be streamlined once #133 is merged! 🙏

packages/ui-components/src/components/MetricLegendCategorical/MetricLegendCategorical.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+} from "./MetricLegendCategorical";
+import { Story } from "@storybook/react";
+
+export default {
+  title: "Components/MetricLegendCategorical",
+  component: MetricLegendCategorical,
+} as ComponentMeta<typeof MetricLegendCategorical>;
+
+const Template: Story<MetricLegendCategoricalProps> = (args) => (
+  <MetricLegendCategorical {...args} />
```
pnavarrc:
```tsx
<div style={{width: 200, border: 'border 1px #ddd'}}>
   <MetricLegendCategorical {...args} />
</div>
```

packages/ui-components/src/components/MetricLegendCategorical/MetricLegendCategorical.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+} from "./MetricLegendCategorical";
+import { Story } from "@storybook/react";
+
+export default {
+  title: "Components/MetricLegendCategorical",
+  component: MetricLegendCategorical,
+} as ComponentMeta<typeof MetricLegendCategorical>;
+
+const Template: Story<MetricLegendCategoricalProps> = (args) => (
+  <div style={{ width: 200, border: "border 1px #ddd" }}>
```
mikelehen: I think this was @pnavarrc's code suggestion, but should that be `border: "solid 1px #ddd"`?  I don't think "border" is a valid border-style

packages/ui-components/src/components/MetricLegendCategorical/MetricLegendCategorical.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { assert } from "@actnowcoalition/assert";
+import { useMetricCatalog } from "../MetricCatalogContext";
+import { MetricCategory } from "@actnowcoalition/metrics";
+import { Stack, Typography } from "@mui/material";
+
+export interface MetricLegendCategoricalProps {
+  /** Metric which we want to display categories for. */
+  metric: Metric | string;
+  /** Supporting text to give context to categories. */
+  supportingText: string;
```
mikelehen, line 13: FWIW- I wonder if it'd make sense for metrics to have a `description` and for us to use that as the default text rather than require passing in `supportingText` (but it could still be optional so you can override it).  Or maybe we should repurpose `extendedName`?  In general I'm not sure exactly how many name / description type properties we want metrics to have. 😆 

Not for this PR, but something to think about.

packages/ui-components/src/components/MetricLegendCategorical/MetricLegendCategorical.tsx
```diff
@@ -0,0 +1,54 @@
...
+  supportingText,
+  orientation,
+}: MetricLegendCategoricalProps) => {
+  const metricCatalog = useMetricCatalog();
+  metric = metricCatalog.getMetric(metric);
+  const items = metric.categories;
+  assert(
+    items,
+    "Metric must define categories in order to use MetricLegendCategorical" +
+      `No categories found for metric ${metric.toString()}`
```
mikelehen: I think there will be a missing space after these get concatenated.

Also nit: You don't need to explicitly call `.toString()`.  JavaScript will do that automatically.

yarn.lock
mikelehen: Most likely somebody else added a dependency and forgot to run `yarn` before committing.  So we should go ahead and check this in.  You could double-check by running `yarn` from `develop` and make sure you get the same diff there.  If in doubt, you can always nuke your changes to `yarn.lock` and then run `yarn`, and whatever it changes should generally be checked in.

mikelehen (approved):
Nice, looks good!

packages/ui-components/src/components/MetricLegendCategorical/MetricLegendCategorical.stories.tsx
pnavarrc: oops, right 😅  - it was to test that the labels wrap if the container is narrow

yarn.lock
smcclure17: Cool, thanks

removing the changes and re-running `yarn` gave the same results, so I think it is expected 🙏

packages/ui-components/src/components/MetricLegendCategorical/MetricLegendCategorical.tsx
smcclure17: ooh cool! and fixed

packages/ui-components/src/components/MetricLegendCategorical/MetricLegendCategorical.stories.tsx
smcclure17: ah, thanks!

packages/ui-components/src/components/MetricLegendCategorical/MetricLegendCategorical.tsx
smcclure17: Yeah, I wasn't immediately sure what to use for `supportingText`--Pablo and I landed on passing in a string (that could be sourced from the CMS or something) as opposed to sourcing it from the metric (with e.g. `metric.extendedName`) but I could definitely see us wanting to pull in content from the metric itself at some point

PR #133, opened 2022-09-01
Make getMetric() accept Metric objects.

mikelehen:


packages/metrics/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actnowcoalition/metrics",
-  "version": "0.0.11",
+  "version": "0.0.12",
```
smcclure17: I bumped the package version to 0.12 in #132, so I think we'll need to bump it again here

smcclure17 (approved):
looks great!

packages/metrics/package.json
mikelehen: Thanks, done!

PR #134, opened 2022-09-01
Downgrade d3-geo and remove remark-gfm.

mikelehen:
This seems to avoid the `ERR_REQUIRE_ESM` errors we've been seeing.

Note that for some reason whenever I try to test ui-components from the hackathon repo with `yarn link` (even without making any changes), the hackathon repo works fine.  So I am unable to test this PR with `yarn link`.  So instead I hacked around.  

Basically:
1. Check out `pablo/add-metric-context` branch in hackathon repo.
2. Run `yarn && yarn dev` and make sure it fails with `ERR_REQUIRE_ESM` error.
3. Manually edit yarn.lock so that ui-components depends on d3-geo "^2.0.2" instead of "^3.0.1" and remove "d3-geo@^3.0.1:" entry.
4. Run `yarn`.
5. Confirm that node_modules/d3-geo/package.json is now 2.0.2 instead of 3.0.1.
6. From act-now-packages/packages/ui-components run `yarn build && cp -r dist/* ~/src/hackathon-july-2022/node_modules/@actnowcoalition/ui-components/dist`
7. From hackathon repo run `yarn dev`

Result: It works.

Since this is hacky, I'm not 100% sure this will solve the issue, but I'm optimistic.

### DIFF
--- a/packages/ui-components/package.json
+++ b/packages/ui-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@actnowcoalition/ui-components",
-  "version": "1.5.10",
+  "version": "1.5.11",
   "description": "UI components for Act Now",
   "repository": {
     "type": "git",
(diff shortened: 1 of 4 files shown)


pnavarrc:
Looks good! Thanks for fixing this again 🤞

pnavarrc (approved):

PR #135, opened 2022-09-01
Update export convention for ui-components

pnavarrc:
Update the export convention for ui components.

## Changes
- In component stories, import components and types from the component folder `index.ts` file  (`import { Component } from ".";`) instead of `import Component from "./Component.tsx"` to ensure that we use the same named exports as client code would. 
- Update `plopfile.mjs` to generate components that use named exports.


faihegberg (approved):
LGTM!

PR #136, opened 2022-09-01
Use new getMetric(metricOrId) overload to clean up some code.

mikelehen:


yarn.lock
```diff
@@ -2,19 +2,6 @@
...
-  integrity sha512-5raAJzeFRSkxuXCKwkqMaqNGS4vpKPtTigusjdQgCoMfLGHqIZuNCVjjgstzPDiUF6E/ah1Rp8LPSAThFNAxbg==
-  dependencies:
-    "@actnowcoalition/assert" "^1.0.0"
-    "@actnowcoalition/number-format" "^1.1.1"
-    "@actnowcoalition/regions" "^1.3.0"
-    "@actnowcoalition/time-utils" "^1.1.0"
-    "@types/papaparse" "^5.3.3"
-    lodash "^4.17.21"
-    papaparse "^5.3.2"
-
```
mikelehen, line 17: Not 100% sure why this was here.  But I get the same diff running `yarn` from `develop` branch.

pnavarrc (approved):
Looks good! Do we need to publish both packages?

mikelehen:
The change to add the `getMetric(metricOrId)` functionality is already published.

The changes in this PR are just internal cleanup (no behavioral effect), so it doesn't really matter if/when we publish the changes.  For that reason I'm not going to bother bumping the version number or publishing right now.  They can get published with some future changes. 🤷

PR #137, opened 2022-09-01
Implement `MetricOverview` component

### LINKED ISSUE #95: Implement MetricOverview component
We need to implement the MetricOverview component. It's probably worth discussing with the group what the best API for the component is as well as some behaviours:

**Open Questions**

- Should it take `Metric`, `MetricData` or both? 
- Should it load the timeseries async to load the current values faster?
- Does it make sense to load the metric def. statically (metric name, description) and current value and timelines async?
- Thoughts on naming for metric-aware components (let's say, for the metric-aware version of `LegendThreshold`?)

Example from the [Monkeypox](https://www.figma.com/file/OG0YkSXfO6YNPfq21jRq42/Monkeypox?node-id=0%3A1) mocks:

Image

**Mobile**

Image

**Desktop**

Image

[Design System - Figma](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=8%3A2798)

pnavarrc:
Close #95 - Implement MetricOverview component

**Vertical**

image

**Horizontal**

image

## Notes

- The styling for paragraphSmall was not being applied correctly, I added a note in `.storybook/preview.tsx` (it seems that the order of the decorators matters, even if they don't use the same context)
- I added the prop `orientation` to the component so we can use version both on mobile or desktop (instead of assuming that we will always want horizontal on mobile).

### DIFF
--- a/packages/ui-components/.storybook/preview.tsx
+++ b/packages/ui-components/.storybook/preview.tsx
@@ -10,9 +10,6 @@ import { theme } from "../src/styles";
  * if they use different contexts. I noticed that if the ThemeProvider is the
  * outer provider, the styles for variant="paragraphSmall" on Typography are
  * not correctly applied.
- *
- * I combined both providers into one decorator to make the nesting order more
- * explicit.
  */
 const appDecorator = (Story) => (
   <MetricCatalogProvider metricCatalog={metricCatalog}>
--- a/packages/ui-components/.storybook/preview.tsx
+++ b/packages/ui-components/.storybook/preview.tsx
@@ -5,17 +5,21 @@ import { MetricCatalogProvider } from "../src/components/MetricCatalogContext";
 import { metricCatalog } from "../src/stories/mockMetricCatalog";
 import { theme } from "../src/styles";
 
-// Wraps stories with the MUI Theme provider
-const themeDecorator = (Story) => (
-  <ThemeProvider theme={theme}>
-    <CssBaseline />
-    <Story />
-  </ThemeProvider>
-);
-
-const metricCatalogDecorator = (Story) => (
+/**
+ * Note (Pablo): It seems that the order of the context providers matter, even
+ * if they use different contexts. I noticed that if the ThemeProvider is the
+ * outer provider, the styles for variant="paragraphSmall" on Typography are
+ * not correctly applied.
+ *
+ * I combined both providers into one decorator to make the nesting order more
+ * explicit.
+ */
+const appDecorator = (Story) => (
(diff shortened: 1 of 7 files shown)


packages/ui-components/.storybook/preview.tsx
```diff
@@ -5,17 +5,21 @@ import { MetricCatalogProvider } from "../src/components/MetricCatalogContext";
...
-
-const metricCatalogDecorator = (Story) => (
+/**
+ * Note (Pablo): It seems that the order of the context providers matter, even
+ * if they use different contexts. I noticed that if the ThemeProvider is the
+ * outer provider, the styles for variant="paragraphSmall" on Typography are
+ * not correctly applied.
+ *
+ * I combined both providers into one decorator to make the nesting order more
+ * explicit.
```
mikelehen: I'd maybe drop the "I combined both providers into one decorator ..." sentence since it doesn't really make sense unless you know what the code looked like before your change.

Also, yikes...   This is weird.  I can't think of any reason that ThemeProvider and MetricCatalogProvider would interfere with each other or be sensitive to nesting order.  FWIW- I tried swapping them as well as checking out `develop` and didn't notice paragraphSmall behave differently, so whatever the issue is seems to be subtle.

packages/ui-components/src/components/MetricOverview/MetricOverview.tsx
```diff
@@ -0,0 +1,63 @@
...
+  metricChart,
+  supportingText,
+  orientation = "vertical",
+}) => {
+  const metricCatalog = useMetricCatalog();
+  const metric = metricCatalog.getMetric(metricOrId);
+
+  if (orientation === "vertical") {
+    return (
+      <Stack direction="column" spacing={1}>
```
mikelehen, line 37: I just want to say that it's pretty impressive / cool that we can build these sorts of components without any .style.ts file with ad-hoc css. 😄

mikelehen (approved):
Looks great!

packages/ui-components/.storybook/preview.tsx
pnavarrc: Maybe swapping the providers in one decorator is not enough, the issue might arise when there are 2 decorators in a specific order? I kind of want to create a minimal repro case and report it to Storybook (if they still don't know about this)

packages/ui-components/src/components/MetricOverview/MetricOverview.tsx
pnavarrc: yeah, pretty cool eh? 😎

PR #139, opened 2022-09-02
Pass `StackProps` to `MetricValue` component

pnavarrc:
By default, the `MetricValue` component will extend to fill the entire width of its container. This PR extends the types of the `MetricValue` props and passes them to the underlying stack component to allow manipulating the alignment of the dot relative to the label

Passing the right props allow to change the component width to `fit-content` and align the component to the right (we might need that if we are showing multiple metric values vertically)

image

On a table, we might want to align the number to the right and the dot to the left

image

I added a few examples in Storybook to confirm that it works well.

### DIFF
--- a/packages/ui-components/src/components/MetricValue/MetricValue.stories.tsx
+++ b/packages/ui-components/src/components/MetricValue/MetricValue.stories.tsx
@@ -23,3 +23,30 @@ DefaultVariant.args = {
 
 export const DataTabular = Template.bind({});
 DataTabular.args = { ...DefaultVariant.args, variant: "dataTabular" };
+
+export const AlignedRight = () => (
+  <div
+    style={{
+      width: 300,
+      border: "dashed 1px #eee",
+      display: "flex",
+      justifyContent: "flex-end",
+    }}
+  >
+    <MetricValue
+      region={washingtonState}
+      metric={MetricId.MOCK_CASES}
+      style={{ width: "fit-content" }}
+    />
+  </div>
+);
+
+export const SpaceBetween = () => (
+  <div style={{ width: 300, border: "dashed 1px #eee" }}>
+    <MetricValue
+      region={washingtonState}
+      metric={MetricId.MOCK_CASES}
+      justifyContent="space-between"
+    />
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/components/MetricValue/MetricValue.stories.tsx
```diff
@@ -23,3 +23,30 @@ DefaultVariant.args = {
...
 DataTabular.args = { ...DefaultVariant.args, variant: "dataTabular" };
+
+export const AlignedRight = () => (
+  <div
+    style={{
+      width: 300,
+      border: "dashed 1px #eee",
+      display: "flex",
+      justifyContent: "flex-end",
+    }}
```
mikelehen, line 34: Mostly for my own curiosity, do we have a preference between doing `<div style={...}>` vs `<Box width=... border=... ...>` ?

packages/ui-components/src/components/MetricValue/MetricValue.stories.tsx
```diff
@@ -23,3 +23,30 @@ DefaultVariant.args = {
...
+  >
+    <MetricValue
+      region={washingtonState}
+      metric={MetricId.MOCK_CASES}
+      style={{ width: "fit-content" }}
```
mikelehen, line 39: Since we use `justifyContent=` directly below, is there a reason we're not using `width=` directly here?

mikelehen (approved):
Nice! Looks good to me.

packages/ui-components/src/components/MetricValue/MetricValue.stories.tsx
pnavarrc: For stories it doesn't matter too much, I have been thinking on having a few reusable decorators to wrap components that need a fixed width or SVG container

packages/ui-components/src/components/MetricValue/MetricValue.stories.tsx
pnavarrc: Good suggestion!

PR #140, opened 2022-09-06
Color map by metric

chasulin:
[Map colored by mock metric data](https://act-now-packages--pr140-casulin-metric-map-nxfuao32.web.app/storybook/index.html?path=/story/components-usnationalmap--metric-aware-states)

Adds `MetricUSNationalMap`, which colors the national map by a given metric, using `getMetricMapFillColor` util. Also reorganizes the map code/files a bit.


packages/ui-components/src/common/utils/importJson.ts
```diff
@@ -0,0 +1,21 @@
+/** Cache used by importJson(). */
```
chasulin, line 1: just moved this util [from index.ts](https://github.com/covid-projections/act-now-packages/pull/140/files#diff-dbac915eece0d9fad387d37708ae6fbbb41e42d48ee671307ce23c167cba4b5fL1-L21) into its own file

packages/ui-components/src/common/utils/getMetricMapFillColor.ts
```diff
@@ -0,0 +1,28 @@
...
+): string {
+  const usRegions = new RegionDB([
+    ...states.all,
+    ...counties.all,
+    ...metros.all,
+  ]);
+
+  const region = usRegions.findByRegionIdStrict(regionId);
+
+  const metric =
```
smcclure17: small nit: I think `getMetric()` can now take either strings or metrics, so I think we can do: `const metric = metricCatalog.getMetric(metricOrId)`

packages/ui-components/src/common/utils/getMetricMapFillColor.ts
```diff
@@ -0,0 +1,28 @@
...
+    typeof metricOrId === "string"
+      ? metricCatalog.getMetric(metricOrId)
+      : metricOrId;
+
+  const { data } = metricCatalog.useData(region, metric);
+  const currentValue = data?.currentValue;
+
+  return currentValue ? metric.getColor(currentValue) : "white";
```
smcclure17, line 24: Just for my understanding, since `useData()` is a hook/is async, will the "loading" state for a region be the default (`"white"`)?

And, is it worth checking to see if an error is returned from `useData()`, or would we rather the map render even if there's an error fetching data for a region/metric?

smcclure17 (approved):
Looks great to me! 🎉 

Just left one nit and a small question

packages/ui-components/src/common/utils/getMetricMapFillColor.ts
chasulin: i _think_ this should be safe enough as is 😬. if `data` (and therefore `currentValue`) resolves to undefined, the function will just fall back to returning `white`

packages/ui-components/src/common/utils/getMetricMapFillColor.ts
chasulin: oh amazing!

packages/ui-components/src/common/utils/getMetricMapFillColor.ts
smcclure17: cool :) works for me!

PR #150, opened 2022-09-06
Implement MetricLegendThreshold

### LINKED ISSUE #123: Implement MetricLegendThreshold component
Implement the metric legend component. See [Figma - MetricLegend](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=13%3A948)

Image

smcclure17:
close #123

Stories: https://act-now-packages--pr150-smcclure17-metric-le-xgnil9gw.web.app/storybook/index.html?path=/story/components-metriclegendthreshold--horizontal-default

### DIFF
--- a/packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.stories.tsx
+++ b/packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.stories.tsx
@@ -7,7 +7,7 @@ import MetricLegendThreshold, {
 import { Story } from "@storybook/react";
 
 export default {
-  title: "Components/MetricLegendCategorical",
+  title: "Components/MetricLegendThreshold",
   component: MetricLegendThreshold,
 } as ComponentMeta<typeof MetricLegendThreshold>;
 
--- a/packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.stories.tsx
+++ b/packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.stories.tsx
@@ -13,7 +13,9 @@ export default {
 } as ComponentMeta<typeof MetricLegendThreshold>;
 
 const Template: Story<MetricLegendThresholdProps> = (args) => (
-  <MetricLegendThreshold {...args} />
+  <div style={{ width: 400, border: "1px solid #ddd" }}>
+    <MetricLegendThreshold {...args} />
+  </div>
 );
 
 // Horizontal legend threshold props
--- a/packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.stories.tsx
+++ b/packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.stories.tsx
@@ -12,9 +12,7 @@ export default {
 } as ComponentMeta<typeof MetricLegendThreshold>;
 
 const Template: Story<MetricLegendThresholdProps> = (args) => (
-  <div style={{ width: 200, border: "1px solid #ddd" }}>
-    <MetricLegendThreshold {...args} />
-  </div>
+  <MetricLegendThreshold {...args} />
 );
 
 // Horizontal legend threshold props
@@ -38,6 +36,15 @@ HorizontalWithoutLabels.args = {
   ...HorizontalDefault.args,
   height: horizontalBarHeight,
   showLabels: false,
+};
+
+export const HorizontalOnlySideLabels = Template.bind({});
+HorizontalOnlySideLabels.args = {
+  ...HorizontalDefault.args,
+  height: horizontalBarHeight,
+  showLabels: false,
+  startLabel: "lower",
+  endLabel: "higher",
 };
 
 export const HorizontalRounded = Template.bind({});
--- a/packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.stories.tsx
+++ b/packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.stories.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import { Typography } from "@mui/material";
 import { ComponentMeta } from "@storybook/react";
 import { MetricId } from "../../stories/mockMetricCatalog";
 import {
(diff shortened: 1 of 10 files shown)


packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
```diff
@@ -0,0 +1,86 @@
...
+    ? { items, getItemColor, getItemLabel }
+    : { items, getItemColor, getItemLabel, getItemSublabel };
+
+  return (
+    <Stack spacing={isHorizontal ? 2 : 3}>
+      <Stack spacing={0.5}>
+        <Typography variant="labelLarge">{metric.name}</Typography>
+        <Typography variant="paragraphSmall">{props.supportingText}</Typography>
+      </Stack>
+      <Stack direction={isHorizontal ? "row" : "column"} spacing={1}>
```
smcclure17: There are no mocks of a vertical thermometer with sidelabels, so it's possible we don't want that functionality. If that's the case, I can add an assert to make sure that if `start`/`endLabel`s are given that `orientation` must be horizontal, but I also don't think it hurts to leave in this functionality? Again, open to thoughts

packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
```diff
@@ -0,0 +1,86 @@
...
+  /** Width of the component. */
+  width?: number;
+  /**
+   * Height of the bars. When not adding the bars, make sure that
+   * `barHeight` is set to the same value as `height`.
+   */
+  barHeight?: number;
+}
+
+export interface MetricLegendThresholdVerticalProps
```
smcclure17, line 47: Initially I had this (and the horizontal version) inheriting from the non-metric-component props like, 

```tsx
export interface MetricLegendThresholdVerticalProps
  extends CommonMetricLegendThresholdProps,
    Omit<
      LegendThresholdVerticalProps<MetricLevel>,
      "getItemColor" | "getItemLabel" | "getItemSublabel" | "items"
    > {}
```
but I felt that was a little opaque. I've opted for explicitly defining the props here, but this does mean that if the underlying component gets changed then these props will need to be changed too, which is probably not ideal, so idk. Open to suggestions

packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+interface CommonMetricLegendThresholdProps {
+  /** Metric to display thresholds for. */
+  metric: Metric | string;
+  /** Whether to show level labels. Does not affect start/endLabels */
+  showLabels?: boolean;
+  /** Optional label for the left or top side of the thermometer. */
+  startLabel?: string;
+  /** Optional label for the right or bottom side of the thermometer. */
+  endLabel?: string;
```
pnavarrc: Consider changing the type of `startLabel` and `endLabel` to be `React.ReactNode`, so we can support formatted text (for example, this is what we use for the homepage risk map)

image

packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
```diff
@@ -0,0 +1,93 @@
...
+}
+
+export interface MetricLegendThresholdHorizontalProps
+  extends CommonMetricLegendThresholdProps {
+  /** Orientation of the bars. */
+  orientation: "horizontal";
+  /** Width of the component. */
+  width?: number;
+  /**
+   * Height of the bars. When not adding the bars, make sure that
```
pnavarrc: It should be "when not adding the labels" right?

packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { ComponentMeta } from "@storybook/react";
+import { MetricId } from "../../stories/mockMetricCatalog";
+import {
+  MetricLegendThreshold,
+  MetricLegendThresholdProps,
+} from "./MetricLegendThreshold";
+import { Story } from "@storybook/react";
+
+export default {
+  title: "Components/MetricLegendThreshold",
```
pnavarrc: Consider moving this component to be under "Metrics"

```tsx
export default {
  title: "Metrics/MetricLegendThreshold",
  // ...
}
```

and maybe other metric-aware components if there are any

packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
```diff
@@ -0,0 +1,93 @@
...
+    >
+      <Stack spacing={0.5}>
+        <Typography variant="labelLarge">{metric.name}</Typography>
+        <Typography variant="paragraphSmall">{props.supportingText}</Typography>
+      </Stack>
+      <Stack direction={isHorizontal ? "row" : "column"} spacing={1}>
+        {props.startLabel && (
+          <Typography variant="paragraphSmall">{props.startLabel}</Typography>
+        )}
+        <LegendThreshold<MetricLevel> {...props} {...derivedProps} />
```
pnavarrc: Passing the props of `MetricLegendThreshold` to the `LegendThreshold` component will make it re-render unnecessarily, consider extracting the props that the `LegendThreshold` component doesn't need before passing them down

```tsx
export const MetricLegendThreshold: React.FC<MetricLegendThresholdProps> = ({
  supportingText,
  startLabel,
  endLabel,
  ...legendThresholdProps
}) => {

 // ...
  return (
    <LegendThreshold<MetricLevel> 
      {...legendThresholdProps}
      {...derivedProps}
    />
  );
}
```

also, just a suggestion, I would recommend having a single conditional to make this block easier to read (this might be a bit repetitive, and not sure if it will be actually better, feel free to ignore)

```tsx

if (props.orientation === "horizontal") {
  return (
    <Stack spacing={0.5}>
      {/* ... */}
    </Stack>
  )
}

return (
  {/* vertical */}
);
```

pnavarrc (approved):
Added a few notes, looks good!

PR #159, opened 2022-09-08
Add `MetricScoreOverview`

### LINKED ISSUE #126: Implement MetricScoreOverview component
[Figma - Overview score](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=9%3A689)

Image

smcclure17:
close https://github.com/covid-projections/act-now-packages/issues/126

view [stories](https://act-now-packages--pr159-smcclure17-metric-sc-g2nc8ye7.web.app/storybook/index.html?path=/story/metrics-metricscoreoverview--default)


packages/ui-components/src/components/MetricScoreOverview/MetricScoreOverview.style.ts
```diff
@@ -0,0 +1,22 @@
+import { styled } from "../../styles";
+import InfoOutlinedIcon from "@mui/icons-material/InfoOutlined";
+
+export const Container = styled("div")`
+  max-width: 240px;
+`;
+
+export const StyledInfoIcon = styled(InfoOutlinedIcon)`
```
smcclure17: Copied this from CAN, with minor tweaks to make it match the mocks

packages/ui-components/src/components/MetricScoreOverview/MetricScoreOverview.tsx
```diff
@@ -0,0 +1,75 @@
...
+      height={72}
+      borderRadius={8}
+      barWidth={12}
+    />
+  );
+  const metricValue = <MetricValue region={region} metric={metric} />;
+
+  // Using color for 'contrast' text color style.
+  const tooltipTypography = (
+    <Typography variant="paragraphSmall" color="#F6F6F6">
```
smcclure17: I couldn't find any info in the mocks on the style of the tooltip text, so using `paragraphSmall` with the font color from the `contrast` text style.
image

packages/ui-components/src/components/MetricScoreOverview/MetricScoreOverview.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export interface MetricScoreOverviewProps {
+  /** Region for which we want to show the metric overview */
+  region: Region;
+  /** Metric for which we want to show the metric overview */
+  metric: Metric | string;
+  /** Optional supporting text for the metric */
+  supportingText?: string;
+  /** Optional text to fill tooltip. If undefined, info tooltip is not rendered. */
+  tooltipText?: string;
```
pnavarrc: Consider renaming `tooltipText` → `tooltipTitle` to be consistent with the props of `InfoTooltip`. It might be good to change its type to match too to `React.ReactNode` so users can pass either a string or formatted text (let's say, a rendered markdown block)

packages/ui-components/src/components/MetricScoreOverview/MetricScoreOverview.tsx
pnavarrc: It might be better to just pass `tooltipText` as is to `InfoTooltip`, that way the typography of the tooltip is consistent with other usages of `InfoTooltip` (we should try to keep customizations of the tooltip content to a minimum)

packages/ui-components/src/components/MetricScoreOverview/MetricScoreOverview.tsx
```diff
@@ -0,0 +1,75 @@
...
+  return (
+    <Container>
+      <Stack direction={"column"} spacing={2}>
+        <Stack direction={"row"} spacing={1} alignItems={"center"}>
+          {metricLegendThreshold}
+          <Stack direction={"column"} spacing={1}>
+            <Typography variant="paragraphSmall">
+              {metric.name}
+              {tooltipText && infoIcon}
+            </Typography>
```
pnavarrc: A few notes: 

- The icon and the text are not correctly aligned, consider using `Stack` inside `Typography` to align them vertically
- If you use `color="inherit"` on the `InfoIcon`, the icon will take the same color as the parent `Typography` element
- Since `tooltipText` is optional, we should not render the icon if `tooltipText` is missing

```tsx
<Typography variant="paragraphSmall" color="secondary.light">
  <Stack direction="row" alignItems="center" spacing={1}>
    <span>{metric.name}</span>
    {tooltipText && (
      <InfoTooltip title={tooltipText}>
        <InfoIcon color="inherit" fontSize="small" />
      </InfoTooltip>
    )}
  <Stack>
</Typography>
```

packages/ui-components/src/components/MetricScoreOverview/MetricScoreOverview.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { Stack, Typography } from "@mui/material";
+import { InfoTooltip } from "../InfoTooltip";
+
+export interface MetricScoreOverviewProps {
+  /** Region for which we want to show the metric overview */
+  region: Region;
+  /** Metric for which we want to show the metric overview */
+  metric: Metric | string;
+  /** Optional supporting text for the metric */
+  supportingText?: string;
```
pnavarrc: I know this is in the mocks, but it might be better to take it out. It will be easy enough to add text below the component (or anywhere else) if we need to

packages/ui-components/src/components/MetricScoreOverview/MetricScoreOverview.style.ts
pnavarrc: I think we should remove all the styling so we can rely more on the MUI system - we didn't know better back then 😂

packages/ui-components/src/components/MetricScoreOverview/MetricScoreOverview.style.ts
```diff
@@ -0,0 +1,22 @@
+import { styled } from "../../styles";
+import InfoOutlinedIcon from "@mui/icons-material/InfoOutlined";
+
+export const Container = styled("div")`
+  max-width: 240px;
```
pnavarrc: Let's not give the component a max-width, the parent component can take care of that.

packages/ui-components/src/components/MetricScoreOverview/MetricScoreOverview.tsx
```diff
@@ -0,0 +1,75 @@
...
+  /** Optional text to fill tooltip. If undefined, info tooltip is not rendered. */
+  tooltipText?: string;
+}
+
+export const MetricScoreOverview: React.FC<MetricScoreOverviewProps> = ({
+  region,
+  metric,
+  supportingText,
+  tooltipText,
+}) => {
```
pnavarrc, line 26: Suggestion: We can allow more layout variations on the component if we allow the user to pass props to the outer `Stack` component.

```tsx
<Stack direction="row" spacing={2} {...otherStackProps} >
  <LegendThreshold />
  <Stack direction="column">
    <Typography variant="paragraphSmall">
      {/* metric name and icon */}
    </Typography>
    <MetricValue region={region} metric={metric} />
  </Stack>
</Stack>
```

if we do that, we can set a max-width, justify the content inside, increase or reduce the spacing, or even move the thermometer to the right side, all without changing the API

```tsx
<MetricScoreOverview 
  direction="row-reverse"
  sx={{maxWidth: 240 }}
/>
```

See the [`MetricValue`](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/components/MetricValue/MetricValue.tsx#L8) component (and stories) for an example

packages/ui-components/src/components/MetricScoreOverview/MetricScoreOverview.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const metricCatalog = useMetricCatalog();
+  metric = metricCatalog.getMetric(metric);
```
pnavarrc: Consider creating a new variable to avoid mutating the prop

packages/ui-components/src/components/MetricScoreOverview/MetricScoreOverview.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const metricLegendThreshold = (
+    <MetricLegendThreshold
+      orientation="vertical"
+      metric={metric}
+      showLabels={false}
+      includeOverview={false}
+      height={72}
+      borderRadius={8}
```
pnavarrc: Maybe borderRadius should be 6? (half of `barWidth`)

packages/ui-components/src/components/MetricScoreOverview/MetricScoreOverview.tsx
```diff
@@ -0,0 +1,75 @@
...
+    </InfoTooltip>
+  );
+
+  return (
+    <Container>
+      <Stack direction={"column"} spacing={2}>
+        <Stack direction={"row"} spacing={1} alignItems={"center"}>
+          {metricLegendThreshold}
+          <Stack direction={"column"} spacing={1}>
+            <Typography variant="paragraphSmall">
```
pnavarrc: The color of the text doesn't match the mocks, I think we need to do

```tsx
<Typography variant="paragraphSmall" color="secondary.light">

</Typography>
```

pnavarrc (commented):
Added a few suggestions, looking good so far!

packages/ui-components/src/components/MetricScoreOverview/MetricScoreOverview.tsx
smcclure17: Oh, that's very elegant! 🤩

packages/ui-components/src/components/MetricScoreOverview/MetricScoreOverview.tsx
smcclure17: Ah awesome--thanks 🙏 

I didn't realize you could put so much inside a `<Typography/>` tag 😅

pnavarrc:
Sorry @smcclure17  - one more 😅, the `InfoIcon` should be inside an `IconButton`, that way the tooltip can be opened by using the keyboard only (pressing the <kbd>tab</kbd> key to focus on the inner button and open the tooltip)

```tsx
<InfoTooltip title={tooltipText}>
  <IconButton>
    <InfoIcon fontSize="small" color="inherit" />
  </IconButton>
</InfoTooltip>
```

See https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/ for a more complete description of the expected behaviours of an accessible tooltip (I think everything else is fine, MUI takes care of the `role`)

packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
```diff
@@ -71,7 +74,14 @@ export const MetricLegendThreshold: React.FC<MetricLegendThresholdProps> = ({
...
   );
-  if (legendThresholdProps.orientation === "horizontal") {
+  const isHorizontal = legendThresholdProps.orientation === "horizontal";
+  const derivedProps = isHorizontal
+    ? { items, getItemColor, getItemLabel }
+    : { items, getItemColor, getItemLabel, getItemSublabel };
+
+  if (!includeOverview) {
+    return <LegendThreshold {...derivedProps} {...legendThresholdProps} />;
+  } else if (isHorizontal) {
```
pnavarrc: nit: This is more of a coding style preference, but I'm going to describe it in case others find it useful.

I think that sequences with `if`, `else if` and `else` are better suited for exclusive options on the same variable

```tsx
if (color === 'green') {
 ...
} else if (color === 'yellow') {
 ...
} else {
 ...
}
```

In this case, `includeOverview` and `isHorizontal` are independent variables, so they don't quite belong to the same conditional sequence. I think this might be more readable

```tsx
if (!includeOverview) {
  return <LegendThreshold ... />
}

if (isHorizontal) {
  ...
} else {
  ...
}
```

But that's more of a personal preference than a good practice, feel free to keep whichever option you find more clear!

pnavarrc (approved):
Looks good! Feel free to merge once the threshold is reversed!

packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
smcclure17: Ah, yeah I like that more! Makes it easier to follow/see what options are being tackled where 🙏

PR #166, opened 2022-09-08
CompareTable (part 2)

pnavarrc:
Part 2 of implementing the `CompareTable`. I only added an example with state management because I'm not sure where that should happen yet. I also left out metrics (sorting by metric value requires access to metric values, etc) to keep the PR small

- Rename `SortableTable` → `CompareTable`
- Implement sub-components needed for the `CompareTable`
  - [`ColumnHeader`](https://act-now-packages--pr166-pablo-sortable-table-izav3d8y.web.app/storybook/index.html?path=/story/table-columnheader--sort-active)
  - [`SortControls`](https://act-now-packages--pr166-pablo-sortable-table-izav3d8y.web.app/storybook/index.html?path=/story/table-sortcontrols--active-asc)
- Implement an example of sorting logic (`CompareTable` → storybook)
- Update the style of the table cells and column headers to better match the [mocks](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System)

[sortable-compare-table.webm](https://user-images.githubusercontent.com/114084/189247517-8f4b0fde-8110-43d6-97ab-f47b21721226.webm)


chasulin (approved):
this all looks great!! one tiny nit- CAN's compare table toggles the sort direction when clicking the entire header cell. this isn't necessarily something we need to recreate, but just the arrow being clickable creates a pretty small click target. again, not a big deal for the time being. this looks 💯

pnavarrc:
Thanks @chasulin - I think that the click area is on the small size (but still reasonable), I will create a few follow up tickets to improve on that and also to take care of the accessibility of the sorting buttons. Thanks again!

PR #174, opened 2022-09-12
CsvDataProvider: allow non-timeseries data to be returned when `includeTimseries=true`

smcclure17:
`metricCatalog.useData()` sets `includeTimeseries` to `true` by default, which means that currently:

```typescript
import { MetricCatalog } from "@actnowcoalition/metrics";

const metricCatalog = new MetricCatalog(...)
// Where MetricId.NO_TS_METRIC is a metric that uses the CsvDataProvider class with no timeseries
const {data, error} = metricCatalog.useData(region, MetricId.NO_TS_METRIC)
```

will throw an error. 

Per discussion in https://github.com/covid-projections/act-now-packages/pull/108#discussion_r959013572, I think the desired behavior is to have `includeTimseries=true` return timeseries data if it exists, else it should just return the non-timeseries data.

### DIFF
--- a/packages/metrics/src/data_providers/CsvDataProvider.test.ts
+++ b/packages/metrics/src/data_providers/CsvDataProvider.test.ts
@@ -56,6 +56,15 @@ describe("CsvDataProvider", () => {
     expect(metricDataTs.timeseries.lastValue).toBe(150);
   });
 
+  test("fetchData() returns non-timeseries data if timeseries data is not available.", async () => {
+    const metricData = await testFetchingCsvData(
+      mockCsv,
+      /*includeTimeseries=*/ true
+    );
+    expect(metricData.currentValue).toBe(150);
+    expect(metricData.hasTimeseries()).toBe(false);
+  });
+
   test("populateCache() fails if csv does not have at least one row.", async () => {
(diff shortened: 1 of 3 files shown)


packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -56,18 +56,6 @@ describe("CsvDataProvider", () => {
...
-      regionColumn: "region",
-      csvText: mockCsv,
-    });
-    expect(async () => {
-      await provider.fetchData([newYork], [testMetric], true);
-    }).rejects.toThrow(
-      "includeTimeseries set to true but cached data has no timeseries."
-    );
-  });
-
```
pnavarrc, line 70: Consider keeping the test, but checking that calling `fetchData` does not throw when the metric data has no timeseries

pnavarrc (approved):
Only a suggestion, looks good!

packages/metrics/src/data_providers/CsvDataProvider.test.ts
smcclure17: Good idea--re-added in d0a9d24fb7f114bdcc4996e749639cbedd5a6137

PR #175, opened 2022-09-12
Implement `MetricSparklines`

### LINKED ISSUE #167: Implement MetricSparklines
It should take two metrics, one for the bar chart (raw data series) and one for the line chart (normally a rolling average). We should also pass a number of days (or a date range) to filter the timeseries by, that way we can make multiple series consistent with each other.

```tsx
// width: 100% height: 100%, 
<MetricSparklines
  region={region}
  metricLineChart={metricCasesRollingAvg}
  metricBarChart={metricCases}
  numDays={30}
/>
```

Image

smcclure17:
close #167 

view [stories](https://act-now-packages--pr175-smcclure17-sparkline-7xqehq5g.web.app/storybook/index.html?path=/story/metrics-metricsparklines--example-five-days)

### DIFF
--- a/packages/ui-components/src/components/MetricSparklines/MetricSparklines.style.ts
+++ b/packages/ui-components/src/components/MetricSparklines/MetricSparklines.style.ts
@@ -1,3 +1,0 @@
-import { styled } from "../../styles";
-
-export const Container = styled("div")``;
--- a/packages/ui-components/src/components/MetricSparklines/MetricSparklines.style.ts
+++ b/packages/ui-components/src/components/MetricSparklines/MetricSparklines.style.ts
@@ -1,0 +1,3 @@
+import { styled } from "../../styles";
+
+export const Container = styled("div")``;
--- a/packages/ui-components/src/components/MetricSparklines/MetricSparklines.tsx
+++ b/packages/ui-components/src/components/MetricSparklines/MetricSparklines.tsx
@@ -1,12 +1,13 @@
 import { Region } from "@actnowcoalition/regions";
 import React from "react";
-import { Container } from "./MetricSparklines.style";
 import { Metric } from "@actnowcoalition/metrics";
 import { SparkLine } from "../SparkLine";
 import { useMetricCatalog } from "../MetricCatalogContext";
 import { Skeleton } from "@mui/material";
+import { SparkLineProps } from "../SparkLine";
 
-export interface MetricSparklinesProps {
+export interface MetricSparklinesProps
+  extends Omit<SparkLineProps, "timeseriesBarChart" | "timeseriesLineChart"> {
   /** Region to generate sparkline for. */
   region: Region;
   /** Metric to use for line element of sparkline. */
@@ -15,19 +16,13 @@ export interface MetricSparklinesProps {
   metricBarChart: Metric | string;
   /** Number of days to show in sparkline, starting from the most recent date looking backwards. */
   numDays: number;
-  /** Width of the spark line component. */
-  width?: number;
-  /** Height of the spark line component. */
-  height?: number;
-  /** Width of each bar, in pixels. */
-  barWidth?: number;
 }
 
 export const MetricSparklines: React.FC<MetricSparklinesProps> = ({
   region,
   metricLineChart,
   metricBarChart,
-  numDays,
+  numDays = 30,
   ...optionalProps
 }) => {
   const metricCatalog = useMetricCatalog();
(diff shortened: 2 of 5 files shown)


packages/ui-components/src/components/MetricSparklines/MetricSparklines.tsx
```diff
@@ -0,0 +1,73 @@
...
+          {...optionalProps}
+        />
+      </Container>
+    );
+  } else {
+    return (
+      <Container>
+        <Skeleton
+          variant="rectangular"
+          width={optionalProps.width ?? 150}
```
smcclure17: These widths and heights are pulled from the defaults in the underlying `SparkLine` component, so it's possible they could become out of sync if either of them change, but I wasn't sure of an easy way to keep them aligned

packages/ui-components/src/components/MetricSparklines/MetricSparklines.style.ts
```diff
@@ -0,0 +1,3 @@
+import { styled } from "../../styles";
+
+export const Container = styled("div")``;
```
pnavarrc: This file can be removed, I don't think the container is necessary

packages/ui-components/src/components/MetricSparklines/MetricSparklines.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  if (data) {
+    const timeseriesLineChart = data
+      .metricData(metricLineChart)
+      ?.timeseries.assertFiniteNumbers()
+      .slice(-numDays);
+    const timeseriesBarChart = data
+      .metricData(metricBarChart)
+      ?.timeseries.assertFiniteNumbers()
+      .slice(-numDays);
```
pnavarrc, line 55: We should consider adding a method to get the most recent `n` days from a timeseries (we already have `filterToDateRange`) or maybe we can use `Timeseries.windowed` and get the last item?

packages/ui-components/src/components/MetricSparklines/MetricSparklines.tsx
pnavarrc: nit: Looks good, consider handling loading and error states before handling the case with data.

packages/ui-components/src/components/MetricSparklines/MetricSparklines.tsx
```diff
@@ -0,0 +1,73 @@
...
+  height?: number;
+  /** Width of each bar, in pixels. */
+  barWidth?: number;
+}
+
+export const MetricSparklines: React.FC<MetricSparklinesProps> = ({
+  region,
+  metricLineChart,
+  metricBarChart,
+  numDays,
```
pnavarrc: Consider passing a default value here and make the `numDays` prop optional (we have used `30` in the past`

packages/ui-components/src/components/MetricSparklines/MetricSparklines.tsx
```diff
@@ -0,0 +1,73 @@
...
+  metricBarChart: Metric | string;
+  /** Number of days to show in sparkline, starting from the most recent date looking backwards. */
+  numDays: number;
+  /** Width of the spark line component. */
+  width?: number;
+  /** Height of the spark line component. */
+  height?: number;
+  /** Width of each bar, in pixels. */
+  barWidth?: number;
+}
```
pnavarrc, line 19: Consider extending the props of `MetricSparklinesProps` (omitting the props we don't want) so they stay in sync

```tsx
export interface MetricSpartlinesProps extends Omit<SparklinesProps, "timeseriesBarChart" | "timeseriesLineChart"
> {
  region: Region;
  metricLineChart: Metric | string;
  metricBarChart: Metric | string;
  numDays?: number;
}


export const MetricSparklines: React.FC<> = ({
  region,
  metricLineChart,
  metricBarChart,
  numDays,
  ...otherSparklinesProps
}) => {

  // handle error and loading states

  return (
    <Sparklines
      timeseriesLineChart={timeseriesLineChart}
      timeseriesBarChart={timeseriesBarChart}
      {...otherSparklinesProps}
    />
  );
}

```

pnavarrc (approved):
Looks good, added a few notes.

packages/ui-components/src/components/MetricSparklines/MetricSparklines.tsx
smcclure17: Yeah, I agree that could be handy to have

If I understand correctly, I think we could add something like:

```typescript
getLastDays(days: number) {
  return this.slice(-days);
}
```
to the `Timeseries` class

PR #178, opened 2022-09-12
Implement MetricCompareTable

### LINKED ISSUE #59: Compare table

pnavarrc:
Close #59 - Implement metric-aware `CompareTable`

image

## Notes 
- I'm going to leave accessibility improvements for a follow up PR (#194) to keep this PR lean (I think that what's pending is to make the header a big button so it's easier to activate and toggle instead of targeting the small icons)
- The metric compare table is also missing links to region pages, I'm going to implement that on a follow-up PR so we can use the same solution for maps.


packages/ui-components/src/components/CompareTable/CompareTable.stories.tsx
```diff
@@ -30,56 +30,68 @@ const StatefulCompareTable: React.FC<{
...
+          onClickSort={(direction) => onClickSort(direction, column.columnId)}
           sx={{ minWidth: 200 }}
         />
       ),
       renderCell: ({ row }) => (
         <TableCell stickyColumn>{row.region.shortName}</TableCell>
       ),
-      compareFn: (rowA, rowB) =>
+      sorterAsc: (rowA, rowB) =>
         rowA.region.shortName < rowB.region.shortName ? -1 : 1,
```
mikelehen: It looks a bit wrong to not handle the `===` case, but I guess for a storybook, we don't care that much.

FWIW- In past projects I've ended up with `compare()` helpers that do this logic.  We may want something similar at some point.

```
export function compare<T>(a: T, b: T): number {
  if (a < b) { return -1; }
  else if (a > b) { return 1; }
  else { return 0;
}
```

That will work for `number`, `string`, `Date`, etc.

packages/ui-components/src/components/CompareTable/CompareTable.stories.tsx
```diff
@@ -30,56 +30,68 @@ const StatefulCompareTable: React.FC<{
...
       ),
       renderCell: ({ row }) => (
         <TableCell align="right">
           <Typography variant="dataTabular">{row.region.regionId}</Typography>
         </TableCell>
       ),
-      compareFn: (rowA, rowB) =>
-        rowA.region.regionId < rowB.region.regionId ? -1 : 1,
+      sorterAsc: (rowA, rowB) =>
+        rowA.region.regionId > rowB.region.regionId ? -1 : 1,
```
mikelehen: ditto here

packages/ui-components/src/components/CompareTable/CompareTable.stories.tsx
```diff
@@ -30,56 +30,68 @@ const StatefulCompareTable: React.FC<{
...
         <ColumnHeader
           stickyColumn
           stickyRow
-          label="Location"
+          label={column.name}
           align="left"
           sortDirection={sortDirection}
-          isSortActive={column.id === sortColumnId}
-          onClickSort={(direction) => onClickSort(direction, column.id)}
+          isSortActive={column.columnId === sortColumnId}
```
mikelehen: Unrelated to this PR, but is there a reason sortDirection and isSortActive are separate properties rather than having sortDirection be undefined when the sort is not active?  What does it mean for a column to have a sortDirection when sort is not active?

packages/ui-components/src/components/CompareTable/interfaces.ts
```diff
@@ -1,21 +1,45 @@
...
+ * */
 export enum SortDirection {
   ASC = "asc",
   DESC = "desc",
 }
 
 export interface ColumnDefinition<R> {
-  id: string;
-  rows: R[];
+  /** A unique ID that identifies the this column. */
```
mikelehen: the this => this

packages/ui-components/src/components/MetricCompareTable/MetricCompareTable.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { useDataForRegionsAndMetrics } from "../../common/hooks";
+import { Row, MetricCompareTableProps } from "./interfaces";
+import { createMetricColumn, createLocationColumn } from "./utils";
+
+export const MetricCompareTable: React.FC<MetricCompareTableProps> = ({
+  regions,
+  metrics: metricOrIds,
+  ...otherCompareTableProps
+}) => {
+  // TODO: It might be better to define and set a context to control the
```
mikelehen: Consider "TODO(pablo)" since you have extra context on what you mean by this that could be useful to somebody looking at this in the future.

packages/ui-components/src/components/MetricCompareTable/MetricCompareTable.style.ts
```diff
@@ -0,0 +1,3 @@
+import { styled } from "../../styles";
+
+export const Container = styled("div")``;
```
mikelehen: I think this file is unused?

packages/ui-components/src/components/MetricCompareTable/utils.tsx
```diff
@@ -0,0 +1,92 @@
...
+        rowB.multiMetricDataStore.metricData(metric);
+
+      if (isNumber(valueA) && isNumber(valueB)) {
+        return valueA - valueB;
+      }
+
+      if (!isNumber(valueA) && !isNumber(valueB)) {
+        return 0;
+      } else {
+        return isNumber(valueA) ? -1 : 1;
```
mikelehen, line 59: Perhaps add a comment "// Sort locations with a metric value before locations that are missing the value." (if I'm understanding the logic right?)

mikelehen (approved):
Couple minor nits, but this looks great!  Very clean and elegant.  Nicely done.

packages/ui-components/src/components/CompareTable/CompareTable.stories.tsx
pnavarrc: Good call, will simplify a bit!

PR #179, opened 2022-09-12
Multi progress bar

### LINKED ISSUE #74: ProgressBar (multiple values)
We need to implement a multi-level ProgressBar component to represent related values such as vaccination percent for multiple doses. I would recommend to keep the interface similar to ProgressBar, but receiving a list of items with accessor methods.

```tsx
<MultiProgressBar
  items={items}
  getItemColor={(item:T, itemIndex:number) => string}
  getItemLabel={(item:T, itemIndex:number) => string}
  getItemValue={(item:T, itemIndex:number) => number}
  {/* ... */}
/>
```

Image

[Act Now Design System (Figma)](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=95%3A4197)

faihegberg:
Closes https://github.com/covid-projections/act-now-packages/issues/74


packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
```diff
@@ -0,0 +1,58 @@
+import React from "react";
+
+import { ProgressBarContainer, StyledSvg } from "./MultiProgressBar.style";
+import { formatPercent } from "@actnowcoalition/number-format";
+
+export interface MultiProgressBarProps<T> {
+  items: T[];
+  bgColor: string;
+  getItemColor: (item: T, itemIndex: number) => string;
+  getItemLabel: (item: T, itemIndex: number) => string;
```
pnavarrc: Maybe we can do `(item: T, itemIndex?: number) => string` instead to have more flexibility

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
```diff
@@ -0,0 +1,58 @@
...
+  getItemLabel,
+  getItemValue,
+  bgColor,
+  width,
+}: MultiProgressBarProps<T>) => {
+  const height = 18;
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const color = bgColor;
+
+  items.sort(
```
pnavarrc: Consider creating a new variable instead of mutating the prop

```tsx
const sortedItems = sortBy([...items], getItemValue);
```

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
```diff
@@ -0,0 +1,58 @@
...
+    (a, b) =>
+      getItemValue(b, items.indexOf(b)) - getItemValue(a, items.indexOf(a))
+  );
+
+  return (
+    <ProgressBarContainer>
+      <StyledSvg width={width} height={height} role="img">
+        <g>
```
pnavarrc: I don't think you need the outer group, you can just use a fragment

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
mikelehen: Just drive-by feedback but I'm not entirely following this suggestion.  As-is, TypeScript would let you pass either `(item, itemIndex) => { ... }` or `(item) => { ... }` function as `getItemLabel` so it's already flexible in that respect (your function doesn't have to accept all of the parameters that are going to be passed).

But if you change it to `getItemLabel: (item: T, itemIndex?: number) => string` as you suggest then your function would have to deal with `itemIndex` being optional.  So you'd have to do ```getItemLabel={(item, itemIndex) => `Item ${itemIndex ?? 'unknown'}`}``` or something because `itemIndex` may be undefined.

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
mikelehen: FYI- I think `[...items]` could just be `items` since `sortBy` is creating a copy already.  We don't need two copies. 😁

faihegberg: 👍🏼

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+function getOffsetPercentage(decimal: number) {
+  return formatPercent(decimal, 1);
+}
+
+export const MultiProgressBar = <T,>({
+  items,
+  getItemColor,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  getItemLabel,
```
pnavarrc: We don't need to pass the `getItemLabel` prop if we are not using it (but we can keep it in the interface)

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+}: MultiProgressBarProps<T>) => {
+  // TODO (Fai): Incorporate aria-labels and/or accessibility attributes.
+
+  const theme = useTheme();
+
+  const sortedItems = sortBy(items, (item) => getItemValue(item) * -1);
+
+  return (
+    <ProgressBarContainer width={width}>
+      <StyledSvg width={width} height={height} role="img">
```
pnavarrc: A few notes

- Let's remove the `role` for now, since we need to figure out how we are going to make the component accessible
- I think that the styled components `ProgressBarContainer` and `StyledSvg` add little value, maybe we can just use the `svg` element and style it directly?

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+          <rect
+            fill={bgColor ?? theme.palette.border.default}
+            x={0}
+            width={width}
+            height={height}
+          />
+          {sortedItems.map((item, i) => {
+            return (
+              <rect
+                key={`item-${i}`}
```
pnavarrc: Consider using the item color as key (see https://github.com/covid-projections/act-now-packages/pull/182)

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+                fill={getItemColor(item)}
+                x={0}
```
pnavarrc: is not necessary to pass `x={0}`, that's the default value

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+  getItemValue,
+  width = 100,
+  height = 16,
+  bgColor,
+}: MultiProgressBarProps<T>) => {
+  // TODO (Fai): Incorporate aria-labels and/or accessibility attributes.
+
+  const theme = useTheme();
+
+  const sortedItems = sortBy(items, (item) => getItemValue(item) * -1);
```
pnavarrc, line 44: Maybe add a comment explaining why we are sorting (and that the sorting has to be descending)

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <ProgressBarContainer width={width}>
+      <StyledSvg width={width} height={height} role="img">
+        <Fragment>
```
pnavarrc: I think we don't really need this fragment?

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+  getItemLabel: (item: T) => string;
+  getItemValue: (item: T) => number;
+  width?: number;
+  height?: number;
+  bgColor?: string;
+}
+
+function getOffsetPercentage(decimal: number) {
+  return formatPercent(decimal, 1);
+}
```
pnavarrc, line 18: Consider implementing a similar interface/data model as the `ProgressBar` component, so we don't need to force the values to be percentages. In `ProgressBar`, we pass `minValue` and `maxValue` and create an internal scale to calculate the width of the components.

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+    />
+    <MultiProgressBar
+      items={sampleItems3}
+      getItemColor={getItemColor}
+      getItemLabel={getItemLabel}
+      getItemValue={getItemValue}
+      maxValue={1}
+    />
+  </Stack>
+);
```
pnavarrc, line 65: Consider splitting the stories to better show each case. Also, if you create variables for each item, you can use them in the stories more easily

```tsx
const Template: Story<MultiProgressBarProps<Item>> = (args) => (
  <MultiProgressBar
    getItemColor={getItemColor}
    getItemLabel={getItemLabel}
    getItemValue={getItemValue}
    maxValue={1}
    {...args} 
  />
);

export const SingleItem = Template.bind({});
SingleItem.args = { items={[item1]} };

export const TwoItems = Template.bind({});
TwoItems.args = { items={[item1, item2]} };
```

pnavarrc (approved):
Looks good, added a few suggestions.

PR #180, opened 2022-09-12
UI components: Region map

chasulin:
Adds `USStateMap` component + metric aware version, with the ability to render counties and/or bordering states.

Also reorganizes some files, prop interfaces, styled components, and fixes imports per [this ticket](https://github.com/covid-projections/act-now-packages/issues/181).

[Stories](https://act-now-packages--pr180-casulin-state-map-t9ag6vyr.web.app/storybook/index.html?path=/story/components-usstatemap--metric-aware-new-york)


packages/ui-components/src/common/utils/maps.ts
```diff
@@ -21,5 +21,14 @@ export default function getMetricMapFillColor(
   const { data } = metricCatalog.useData(region, metric);
   const currentValue = data?.currentValue;
 
-  return currentValue ? metric.getColor(currentValue) : "white";
+  return currentValue ? metric.getColor(currentValue) : "lightGray";
```
chasulin, line 13: changing the default fill to lightGray instead of white

packages/ui-components/src/common/utils/maps.ts
```diff
@@ -3,7 +3,7 @@
 import { Metric, MetricCatalog } from "@actnowcoalition/metrics";
 import { RegionDB, states, counties, metros } from "@actnowcoalition/regions";
 
-export default function getMetricMapFillColor(
+export function getMetricMapFillColor(
   metricCatalog: MetricCatalog,
   metricOrId: Metric | string,
   regionId: string
```
pnavarrc: I would recommend updating the input arguments of `getMetricMapFillColor`:

The parent component can easily access the regions that the map is using and pass the region as parameter instead of the `regionId`. For example, 

```tsx
<USNationalMap
  getFillColor={(regionId) => {
    const region = usRegions.findByRegionId(regionId);
    return region ? getMetricMapFillColor(metricCatalog, region, metric) : defaultColor;
  }}
  {...otherProps}
/>
```

Similarly, since the component using `getMetricMapFillColor` has access to the metric catalog, we can simplify the implementation of `getMetricMapFillColor` by passing the `Metric` instance instead of `Metric | string` (I'm okay either way with this one though)

```tsx
<USNationalMap
  getFillColor={(regionId) =>
    getMetricMapFillColor(metricCatalog, metricCatalog.getMetric(metric), regionId)
  }
  {...otherProps}
/>
```

That way the implementation of `getMetricMapFillColor` will be simpler and will be able to handle any region

```tsx
export function getMetricMapFillColor(
  metricCatalog: MetricCatalog,
  metric: Metric,
  region: Region
): string {
  const { data } = metricCatalog.useData(region, metric);
  const currentValue = data?.currentValue;
  return currentValue ? metric.getColor(currentValue) : "lightGray";
```

packages/ui-components/src/components/Maps/Maps.style.ts
```diff
@@ -17,15 +15,34 @@ export const PositionAbsolute = styled("div")`
...
+`;
+
+export const RegionShapeBase = styled("path")`
+  stroke: ${(props) => props.theme.palette.common.white};
+  stroke-width: 1;
+`;
+
+export const HighlightableShape = styled("path")<{
+  highlight: boolean;
+}>`
```
pnavarrc, line 46: Consider removing the `highlight` prop so it doesn't trigger the warning (`isValidProp`)

packages/ui-components/src/components/Maps/Maps.style.ts
```diff
@@ -17,15 +15,34 @@ export const PositionAbsolute = styled("div")`
...
+  cursor: pointer;
+  &:hover {
+    fill: ${(props) => props.theme.palette.grey[400]};
+  }
+`;
+
+export const RegionShapeBase = styled("path")`
+  stroke: ${(props) => props.theme.palette.common.white};
+  stroke-width: 1;
+`;
```
pnavarrc: nit: Maybe not worth it, but since this is the base, we could use it in other styled components to simplify the CSS a little

```tsx
export const BorderingRegion = styled(RegionShapeBase)`
  fill: ${(props) => props.theme.palette.grey[300]};
  cursor: pointer;
  &:hover {
    fill: ${(props) => props.theme.palette.grey[400]};
  }
`;
``

File: packages/ui-components/src/components/Maps/USNationalMap/MetricUSNationalMap.stories.tsx [hunk 5]
```diff
@@ -1,14 +1,13 @@
...
-} from "./MetricUSNationalMap";
+import { MetricUSNationalMap } from "./MetricUSNationalMap";
 import { states } from "@actnowcoalition/regions";
 import { defaultWidth } from "../../../common/geo-shapes";
 import { MetricId } from "../../../stories/mockMetricCatalog";
+import { MetricUSNationalMapProps } from "./interfaces";
 
 export default {
-  title: "Components/USNationalMap",
+  title: "Components/Maps/US National",
```
pnavarrc: How about moving maps to the top level?

packages/ui-components/src/components/Maps/USStateMap/USStateMap.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+  stateRegionId: "36",
+  renderTooltip: (regionId: string) => renderSimpleTooltip(regionId),
+  showBorderingStates: false,
+};
+
+export const NewYorkCountiesWithBorderingStates = Template.bind({});
+NewYorkCountiesWithBorderingStates.args = {
+  stateRegionId: "36",
+  renderTooltip: (regionId: string) => renderSimpleTooltip(regionId),
+};
```
pnavarrc, line 47: Thanks for all the examples! 🙌

packages/ui-components/src/common/utils/maps.ts
```diff
@@ -21,5 +21,14 @@ export default function getMetricMapFillColor(
...
+  return currentValue ? metric.getColor(currentValue) : "lightGray";
+}
+
+/** Checks if a county or congressional district belongs to a given state */
+
+export function belongsToState(
+  countyOrCongressionalDistrictFips: string,
+  stateFips: string
+) {
+  return countyOrCongressionalDistrictFips.substring(0, 2) === stateFips;
```
pnavarrc, line 21: We should probably revisit this at some point, now that we have `Region.parent` to represent hierarchy between regions.

pnavarrc (approved):
Looks good, added a few notes. I think we might want to leverage more on the `Region` type in the future (using more `Region` objects than `regionId`), but we can punt on that for now.

packages/ui-components/src/common/utils/maps.ts
chasulin: the region param edit makes sense to me, so that the util can be used for any region and not just US regions. im going to nix the other edit (metric param change) -- as id rather keep metric logic in the util

packages/ui-components/src/components/Maps/Maps.style.ts
chasulin: just wired up highlight and added isValidProp

Issue #177, opened 2022-09-12
Possible bug: Sorting CompareTable crashes when the cells use the useData hook


pnavarrc:
I'm working on implementing a metric-aware compare table, which renders fine initially, but it crashes if we attempt to sort the table (even if we don't sort by metric).

## Notes 

- Sorting works just fine for non-metric aware components (See storybook - https://act-now-packages--pr166-pablo-sortable-table-izav3d8y.web.app/storybook/index.html?path=/story/table-comparetable--example)
- I haven't figured out a good pattern to access MetricData from the `compareFn` of each column, so I set the sorting function to be by population of the region. One option could be to fetch metric data for all the regions and metrics up-front and include the data on the row data items, I tried that and the table still crashes. 

I suspect that this might be due to a race condition between the compare table internal state and the state of the mock data provider, but I'm not sure.

## How to Test

- Using Storybook https://act-now-packages--pr178-pablo-examples-compa-3fppwcmo.web.app/storybook/index.html?path=/story/metrics-metriccomparetable--example (it also crashes with metrics that use the mock data metric provider)

Using the hackathon repo:

- Use this branch https://github.com/covid-projections/act-now-packages/pull/166 and link `ui-components`
- On the hackathon repo, link `@actnowcoalition/ui-components`
- Add the MetricCompareTable to the homepage, adding all the states as regions and all the available metrics

```tsx
import { MetricCompareTable } from '@actnowcoalition/ui-components';

// ...
<MetricCompareTable
  regions={states.all}
  metrics={metricsCatalog.metrics}
/>
```

Note that the table will render correctly, but when we try to sort the table (even if we sort by location name), the table will crash.

mikelehen:
Fixed by #182

PR #182, opened 2022-09-12
Fix compare table sorting.

mikelehen:
Per https://reactjs.org/docs/lists-and-keys.html:

> We don’t recommend using indexes for keys if the order of items may change. This can negatively impact performance and **may cause issues with component state**. Check out Robin Pokorny’s article for an [in-depth explanation on the negative impacts of using an index as a key](https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/). If you choose not to assign an explicit key to list items then React will default to using indexes as keys.

I think what was happening is that after re-sorting the table, React didn't know that the order of elements had changed and so the internal `useEffect()` calls within each row were returning data for wrong rows.

Note I left it using the `columnIndex` in the keys, since we don't support re-ordering / removing columns, but we may want to revisit.  

Feel free to rework this in any way you like.  Another option might be to use the *initial* index of each row (before sorting) as the row's key.  That would _probably?_ be safe and means the consumer of the table doesn't have to provide explicit IDs.

### DIFF
--- a/packages/ui-components/src/components/CompareTable/CompareTable.stories.tsx
+++ b/packages/ui-components/src/components/CompareTable/CompareTable.stories.tsx
@@ -17,12 +17,13 @@ export default {
 } as ComponentMeta<typeof CompareTable>;
 
 interface RowItem {
+  rowId: string;
   region: Region;
 }
 
 const rows: RowItem[] = states.all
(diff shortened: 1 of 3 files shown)


pnavarrc (approved):
Excellent, thanks @mikelehen !

mikelehen:
Fixes #177

PR #183, opened 2022-09-13
Implement `RectClipGroup` component

pnavarrc:
Implement the `RectClipGroup` component to make it easier to implement rounded borders and to hide elements that would be outside the charting area

image

I also updated the `LegendThresholdHorizontal` and `ProgressBar` components to use this new building block.

### DIFF
--- a/packages/ui-components/src/components/RectClipGroup/RectClipGroup.stories.tsx
+++ b/packages/ui-components/src/components/RectClipGroup/RectClipGroup.stories.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { ComponentMeta } from "@storybook/react";
 import { RectClipGroup } from ".";
 
 export default {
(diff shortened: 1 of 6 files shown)


packages/ui-components/src/components/RectClipGroup/RectClipGroup.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+  component: RectClipGroup,
+} as ComponentMeta<typeof RectClipGroup>;
+
+const width = 300;
+const height = 300;
+
+const Template: ComponentStory<typeof RectClipGroup> = (args) => (
+  <svg width={width} height={height} style={{ border: "solid 1px #ddd" }}>
+    <RectClipGroup {...args} />
+    {/* The purple circle will show because it's outside the RectClipGroup */}
```
mikelehen: This comment doesn't make sense to me.  Should it say "inside" instead of "outside"?

packages/ui-components/src/components/RectClipGroup/RectClipGroup.tsx
```diff
@@ -0,0 +1,36 @@
...
+ *     <rect x={0} width={400} height={200} fill="red" />
+ *   </RectClipGroup>
+ * </svg>
+ * ```
+ */
+export const RectClipGroup: React.FC<RectClipGroupProps> = ({
+  children,
+  ...rectProps
+}) => {
+  const clipPathId = useId();
```
mikelehen, line 25: Ooh, nifty.  Didn't know about `useId()`.

mikelehen (approved):
Looks good!

packages/ui-components/src/components/RectClipGroup/RectClipGroup.tsx
pnavarrc: me neither, just found out about it today!

packages/ui-components/src/components/RectClipGroup/RectClipGroup.stories.tsx
pnavarrc: outside was correct (RectClipGroup will only clip its children, not components outside it), but I simplified the example to make things a bit more clear

PR #185, opened 2022-09-13
Move useData react hooks to ui-components package.

mikelehen:
Moves all of our metric data fetching react hooks from `metrics` package to `ui-components` and removes the react dependencies from the `metrics` package.

Note that I'm updating both packages in a single PR which normally is a bad idea, but since `ui-components` doesn't depend on the changes in `metrics`, it's safe.  

After this is merged, I'll:
* publish the packages
* bump ui-components to depend on the newly-published metrics package and publish ui-components again
* Update our template / hackathon repos to use the new packages and fix any `useData` hook calls.


packages/ui-components/src/common/hooks/metric-data.test.ts
```diff
@@ -0,0 +1,106 @@
+import { renderHook } from "@testing-library/react-hooks";
```
mikelehen, line 1: Everything in this file is copied over from MetricCatalog.test.ts.

packages/ui-components/src/common/hooks/metric-data.ts
```diff
@@ -0,0 +1,96 @@
+import { Region } from "@actnowcoalition/regions";
```
mikelehen, line 1: All of the code in this file is copied out of MetricCatalog.ts (other than changing the hooks to top-level functions and leveraging `useMetricCatalog()`.

packages/ui-components/src/common/hooks/metric-data.ts
```diff
@@ -0,0 +1,96 @@
...
+  const data = dataStore?.regionData(region);
+  return { data, error };
+}
+
+export function useDataForRegionsAndMetrics(
+  regions: Region[],
+  metrics: Array<string | Metric>,
+  includeTimeseries = true
+): DataOrError<MultiRegionMultiMetricDataStore> {
+  const catalog = useMetricCatalog();
```
mikelehen: Here's where we leverage the `useMetricCatalog()` hook.

packages/ui-components/src/common/hooks/useCachedArrayIfEqual.test.ts
```diff
@@ -0,0 +1,34 @@
+import { renderHook } from "@testing-library/react-hooks";
```
mikelehen, line 1: This test is copied from MetricCatalog.test.ts

packages/ui-components/src/common/hooks/useCachedArrayIfEqual.ts
```diff
@@ -0,0 +1,23 @@
+import isEqual from "lodash/isEqual";
+import { useRef } from "react";
```
mikelehen, line 2: This code is copied from MetricCatalog.ts.

packages/ui-components/src/common/utils/getMetricMapFillColor.ts
```diff
@@ -2,8 +2,9 @@
 
 import { Metric, MetricCatalog } from "@actnowcoalition/metrics";
 import { RegionDB, states, counties, metros } from "@actnowcoalition/regions";
+import { useData } from "../hooks";
 
-export default function getMetricMapFillColor(
+export default function useMetricMapFillColor(
```
mikelehen, line 7: I did a drive-by rename here.  Since this is using `useData()` and other hooks, it constitutes a "custom hook" and should be named with a `use` prefix (mostly so that React's static analysis (lint checks) will check it for violating hook rules, I think).

packages/ui-components/src/common/utils/getMetricMapFillColor.ts
```diff
@@ -18,7 +19,11 @@ export default function getMetricMapFillColor(
 
   const metric = metricCatalog.getMetric(metricOrId);
 
-  const { data } = metricCatalog.useData(region, metric);
+  // TODO(michael): This function should be renamed to useMetricMapFillColor()
+  // since it is using hooks, but that causes additional rules-of-hooks failures
+  // so leaving it as is for now.
+  // eslint-disable-next-line react-hooks/rules-of-hooks
```
mikelehen: I'll try to solve this in a subsequent PR but it seemed like it might get somewhat invasive.

pnavarrc (approved):
Looks good!

PR #186, opened 2022-09-13
Remove useData react hooks from metrics package.

mikelehen:


mikelehen:
Actually, I can merge this with #185 I think.

PR #190, opened 2022-09-14
Implement LineChartLevels

### LINKED ISSUE #188: Implement LevelLineChart
Ideally, we want the props to be somewhat consistent (or easy to map to) `Metric.thresholds` and `MetricLevel` so we can easily implement the metric-aware version of this chart.

The chart should assert that the number of levels is compatible with the number of thresholds (there should be one fewer threshold than level)

```tsx
const levels: MetricLevel[] = [
  { id: "low", color: "green" },
  { id: "medium", color: "yellow" },
  { id: "high", color: "red" },
];

const thresholds = [10, 50];

<LineChartLevels
  timeseries={timeseries}
  xScale={xScale}
  yScale={yScale}
  levels={levels}
  thresholds={thresholds}
/>
```

pnavarrc:
Close #188 - Implement the LineChartLevels component

This is a low-level chart that will have the same API as the `LineChart` component, except that the line is colored depending on the y-value of each line segment so we can use it for graded metrics. For example, if we define the following intervals

```tsx
const intervals: ChartInterval[] = [
  { lowerBound: 0.0, upperBound: 0.4, level: LOW }
  { lowerBound: 0.4, upperBound: 0.6, level: MEDIUM }
  { lowerBound: 0.6, upperBound: 1.0, level: HIGH }
];
```

it will define the colored regions below:

image

This is implemented by overlapping a full line of each color, but using the `RectClipGroup` component to hide the parts of the line that are outside the interval (showing them side by side here)

image

The result will look like:

image

The intervals can be calculated using a list of `MetricLevel` and their corresponding thresholds (plus the minimum and maximum values of the chart). We could calculate the intervals internally, but I think we will likely need them in the metric-aware version of this chart to add level annotations as we did for CAN (having the lower and upper bounds will help us putting the labels in the middle of each interval).

```tsx
export function generateChartIntervals(
  levels: MetricLevel[], 
  thresholds: number[], 
  minValue: number, 
  maxValue: number): ChartInterval[] {
  // check consistency between number of thresholds and levels
  // handle ascending or descending thresholds
  // calculate intervals and return
}
```

with this function, we should be able to implement a metric-aware function of this chart easily

```tsx
// ...
const { minValue, maxValue } = timeseries;

const yScale = scaleLinear({
  domain: [minValue, maxValue],
  range: [innerHeight, 0]
});

// The function returns consistent, non-overlapping, sorted intervals
// that cover the entire chart extent
const intervals = generateChartIntervals(metric.levels, metric.thresholds, minValue, maxValue)

return (
  <svg width={width} height={height}>
    <Group left={paddingLeft} top={paddingTop}>
      <AxisLeft scale={yScale} />
      <AxisBottom top={innerHeight} scale={xScale} />
      <LineChartLevels
        timeseries={timeseries}
        xScale={xScale}
        yScale={yScale}
        intervals={intervals}
        {...otherLineProps}
      />
     {/* We position the labels in the middle of each interval*/}
      {intervals.map((interval)=>(
        <StyledText x={20} y={getMidPoint(interval)}>
          {interval.name}
        </StyledText>
      ))}
      {/* current point and value, etc /*}
    </Group>
  </svg>
);
```

## Feedback requested

- Does it feel like a good idea to have `ChartInterval`? Any other ideas to explore?
- Better names for the intervals or the chart?
- The function `generateChartIntervals` will need to do a lot (what if the minValue is between two thresholds, handle reversed thresholds, etc), is that okay?
- Maybe having a class with more logic to represent the set of intervals would be better?


packages/ui-components/src/components/LineChartLevels/LineChartLevels.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+const width = 600;
+const height = 400;
+
+export default {
+  title: "Charts/LineChartLevels",
+  component: LineChartLevels,
+} as ComponentMeta<typeof LineChartLevels>;
+
+const timeseries = SolidLine.args?.timeseries;
```
mikelehen, line 17: This feels a tad hacky. 😛   We should maybe move the apple stock timeseries into a shared helper file somewhere.  Feel free to punt for now though.

packages/ui-components/src/components/LineChartLevels/LineChartLevels.tsx
```diff
@@ -0,0 +1,92 @@
+import React, { Fragment } from "react";
+import { MetricLevel } from "@actnowcoalition/metrics";
+import { assert } from "@actnowcoalition/assert";
+import { LineChart, LineChartProps } from "../LineChart";
+import { RectClipGroup } from "../RectClipGroup";
+
+/**
+ *
```
mikelehen, line 8: Guessing you're planning to fill in the comments later. 😁

packages/ui-components/src/components/LineChartLevels/LineChartLevels.tsx
```diff
@@ -0,0 +1,92 @@
...
+ */
+export interface ChartInterval {
+  /** */
+  lowerBound: number;
+  /** */
+  upperBound: number;
+  /**  */
+  level: MetricLevel;
```
mikelehen, line 16: It feels a bit weird to use `MetricLevel` in a non-metric-aware UI component.  It looks like the only thing we actually need is a color.  So could we replace this with `color: Color` ?

At that point, it probably doesn't make sense to put "Level" in the component name...  Maybe "LineChartMultiColor" or "LineChartColorIntervals" or something?

packages/ui-components/src/components/LineChartLevels/LineChartLevels.tsx
```diff
@@ -0,0 +1,92 @@
...
+              xScale={xScale}
+              yScale={yScale}
+              stroke={interval.level.color}
+            />
+          </RectClipGroup>
+        );
+      })}
+    </Fragment>
+  );
+};
```
mikelehen, line 83: FWIW- This component is simple enough that it's a little tempting not to have a non-metric-aware version of it.  We could just inline this logic into the MetricThresholdsLineChart or whatever.  But I don't feel strongly at all.

mikelehen (commented):
Generally this seems totally reasonable / understandable.  I left some possible alternatives / tweaks, but I don't feel very strongly.

FWIW- Instead of having a `generateChartIntervals` function, perhaps we could build some sort of helper like that into the `Metric` class itself.  The `thresholds` / `levels` are defined the way they are on `Metric` to make it easy (and mistake-resistant) to *define* the levels / thresholds for a metric, but they are not optimal for building charts / visuals, etc. that need to be threshold-aware.  I think we could have a `metric.metricLevelIntervals` getter that returns an array of `MetricLevelInterval` that's basically identical to your `ChartInterval`.  And that would perhaps be useful for other components as well. 🤷

packages/ui-components/src/components/LineChartLevels/LineChartLevels.stories.tsx
pnavarrc: Good idea, makes sense!

packages/ui-components/src/components/LineChartLevels/LineChartLevels.tsx
pnavarrc: That's a good point - I think that building the logic on the metric makes more sense, I'm going to wait for a bit more feedback and then potentially close this PR (we can always look at the code in it once we implement the metric-aware version of it).

pnavarrc:
Closing this PR (the component is probably too simple to be worth merging) - we can reference the code one we start working on the metric-aware version.

PR #193, opened 2022-09-14
Add social meta tags

### LINKED ISSUE #169: Implement SocialMetaTags component
Let's use a placeholder for now. The social copy, title, etc will ultimately be written in the CMS, but we can implement the component and add it to the template pages

faihegberg:
Closes https://github.com/covid-projections/act-now-packages/issues/169


packages/ui-components/src/components/SocialMetaTags/PageMetaTags.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const PageMetaTags: React.FC<PageMetaTagsProps> = ({
+  siteName,
+  url,
+  title,
+  description,
+}) => {
+  return (
+    <Fragment>
+      <title data-react-helmet="true">{title}</title>
+      <link data-react-helmet="true" rel="canonical" href={url} />
```
mikelehen: I don't think we need data-react-helmet in any of these.  react-helmet is something we used for CAN but we shouldn't need it for next.js apps.  I think!

packages/ui-components/src/components/SocialMetaTags/PageMetaTags.tsx
```diff
@@ -0,0 +1,30 @@
...
+      <meta data-react-helmet="true" name="description" content={description} />
+      <meta property="og:type" content="website" />
+      <meta property="og:site_name" content={siteName} />
+      <meta property="og:url" content={url} />
+      <meta property="og:title" content={title} />
+      <meta property="og:description" content={description} />
+      <meta name="twitter:title" content={title} />
+      <meta name="twitter:description" content={description} />
```
mikelehen, line 43: Just confirming that we are intentionally including these here even though they're twitter-related?

packages/ui-components/src/components/SocialMetaTags/TwitterMetaTags.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { Fragment } from "react";
+
+export interface TwitterMetaTagsProps {
+  twitterCard?: string;
```
mikelehen: It looks like this is the type of card and it can be "summary", "summary_large_image", "player", or "app" so can we make the type be `"summary" | "summary_large_image" | "player" | "app"` instead of just `string`? to make it easier to use?

packages/ui-components/src/index.ts
```diff
@@ -92,11 +92,12 @@ export * from "./components/MetricDot";
 export * from "./components/MetricLegendCategorical";
 export * from "./components/MetricLegendThreshold";
 export * from "./components/MetricOverview";
+export * from "./components/MetricScoreOverview";
```
mikelehen, line 95: This seems unrelated to your changes.  Is it intentional?

packages/ui-components/src/components/SocialMetaTags/FacebookMetaTags.tsx
```diff
@@ -0,0 +1,31 @@
...
+}) => {
+  return (
+    <Fragment>
+      {fbAppId && <meta property="fb:app_id" content={fbAppId} />}
+      {fbImg && fbImgWidth && fbImgHeight && (
+        <>
+          <meta property="og:image:url" content={fbImg} />
+          <meta property="og:image:width" content={fbImgWidth} />
+          <meta property="og:image:height" content={fbImgHeight} />
+          <meta property="og:image:type" content={fbImgType} />
```
mikelehen: FWIW- These are all og (Open Graph) tags that aren't really specific to Facebook.  I think they're responsible for making preview images show up correctly in e.g. Slack or LinkedIn, etc. as well.  fb:app_id is the only one that's Facebook-specific.

So I am kinda' tempted to combine all of these components into a single PageMetaTags component or something.  But I don't feel very strongly.

mikelehen (approved):
Looks good to me with a couple nits. And I'm also a bit tempted to merge all 3 components into one. But I don't feel strongly.  So feel free to leave it as-is or run it by Pablo or Chelsi or something.

Also, can you remove "Closes #169" since I think we need to also update the template repo to use these components before we close that issue (or you could create a separate issue for using it in the template repo if you prefer).  Thanks!

faihegberg:
> Looks good to me with a couple nits. And I'm also a bit tempted to merge all 3 components into one. But I don't feel strongly. So feel free to leave it as-is or run it by Pablo or Chelsi or something.
> 
> Also, can you remove "Closes #169" since I think we need to also update the template repo to use these components before we close that issue (or you could create a separate issue for using it in the template repo if you prefer). Thanks!

Thank you! I discussed merging these 3 components into one with Pablo but we figured it will be cleaner and more readable separated.

As for updating the template repo, I'll create a separate issue.

packages/ui-components/src/components/SocialMetaTags/PageMetaTags.tsx
faihegberg: Yea, I think it's a bit odd too 😅 but this is because it's using the `title` and `description` props, which are general page props. If we are to include these in `TwitterMetaTags` component instead, we will need to pass `title` and `description` to that component, which may be a little bit redundant for pages where we're using both `PageMetaTags` and `TwitterMetaTags` component.

packages/ui-components/src/components/SocialMetaTags/TwitterMetaTags.tsx
faihegberg: Yes, good call!

packages/ui-components/src/index.ts
faihegberg: Yes, there was a merge conflict and I resolved it. But after I resolved it this line was not placed alphabetically. So I just moved it from the bottom to its appropriate place 😄

PR #197, opened 2022-09-15
Fix coloring functions for metric maps

chasulin:
Problem: the counties view of `MetricUSNationalMap` was not wired up, because getting the map fill color was dependent on the `useData` hook, and the counties map uses canvas, which is dependent on `useEffect` (and you can't call a hook inside of a `useEffect`).

As a possible solution, we now call `useDataForRegionsAndMetrics` at the top level (in MetricUSNationalMap.tsx and MetricUSStateMap.tsx) and then pass the resulting data to the map's `getFillColor` prop


.eslintrc.json
```diff
@@ -21,7 +21,8 @@
   "plugins": ["react", "@typescript-eslint"],
   "rules": {
     "@typescript-eslint/no-explicit-any": "off",
-    "prefer-const": "off"
+    "prefer-const": "off",
+    "react-hooks/rules-of-hooks": "off"
```
chasulin, line 25: temporary solution so i could commit my changes

packages/ui-components/src/common/hooks/useMetricMapFillColor.ts
```diff
@@ -0,0 +1,15 @@
+/** Returns color associated with a metric's current value for a region. */
```
chasulin, line 1: just moved over into `hooks` [from its previous home in `utils`](https://github.com/covid-projections/act-now-packages/pull/197/files#diff-ff544a559ac553ce8a03afcbb2d8f74cb1f2b5c5c874c05b226cd03ac3776715L6-L19) -(.. a move which is causing its own problems now)

packages/ui-components/src/common/utils/maps.ts
```diff
@@ -1,27 +1,47 @@
...
+}
+
+export function generateFillColorMap(
+  regions: Region[],
+  metric: string | Metric,
+  data: MultiRegionMultiMetricDataStore
+): FillColorForRegionId {
+  const regionIdAndFillColor = regions.map((region) => {
+    const currentValue = data.metricData(region, metric)?.currentValue;
+    const metricFromData = data.metricData(region, metric).metric;
```
mikelehen: Not sure why we need this.  It should be guaranteed to be the same as `metric`, I think?

And actually, I think this `regionIdAndFillColor` could be simplified to:

```
  const regionIdAndFillColor = regions.map((region) => (
    {
      regionId: region.regionId,
      fillColor: data.metricData(region, metric).getColor()
    }));
```

packages/ui-components/src/components/Maps/USNationalMap/MetricUSNationalMap.tsx
```diff
@@ -1,24 +1,42 @@
...
+  const [mapRegions, setMapRegions] = useState(states.all);
+
+  useEffect(() => {
+    if (showCounties) {
+      setMapRegions(statesAndCounties.all);
+    }
+    if (!showCounties) {
+      setMapRegions(states.all);
+    }
+  }, [showCounties]);
```
mikelehen: I'm not sure why we need `useEffect` and `useState` here.  Could we just do:

```
const mapRegions = showCounties ? statesAndCounties.all : states.all;
```
?

packages/ui-components/src/components/Maps/USNationalMap/MetricUSNationalMap.tsx
```diff
@@ -1,24 +1,42 @@
...
     return null;
   }
 
+  const regionIdToFillColorMap = generateFillColorMap(mapRegions, metric, data);
+
   return (
     <USNationalMap
       getFillColor={(region) =>
-        getMetricMapFillColor(metricCatalog, metric, region)
+        regionIdToFillColorMap[region.regionId].fillColor
```
mikelehen: Not a big deal, but I think we could get rid of `regionIdToFillColorMap` and `generateFillColorMap` and just do:

```
getFillColor={(region) => data.metricData(region, metric).getColor()}
```

packages/ui-components/src/components/Maps/USStateMap/MetricUSStateMap.tsx
```diff
@@ -1,24 +1,36 @@
...
     return null;
   }
 
+  const regionIdToFillColorMap = generateFillColorMap(mapRegions, metric, data);
+
   return (
     <USStateMap
       getFillColor={(region) =>
-        getMetricMapFillColor(metricCatalog, metric, region)
+        regionIdToFillColorMap[region.regionId].fillColor
```
mikelehen: Again, I think this could be:
```
getFillColor={(region) => data.metricData(region, metric).getColor()}
```

mikelehen (approved):
Looks good, but I think there are a couple opportunities for cleanup / simplification!

mikelehen (commented):
Oh, also, this should ideally have a changeset added (patch level change for ui-components with a summary of "Fix coloring for metric maps" or whatever).

packages/ui-components/src/components/Maps/USNationalMap/MetricUSNationalMap.tsx
chasulin: ah yes, i always overuse `useEffect`+`useState`

packages/ui-components/src/components/Maps/USStateMap/MetricUSStateMap.tsx
chasulin: ha. oh. so much simpler.

PR #204, opened 2022-09-20
Upgrade Storybook to 6.5.12

pnavarrc:


.changeset/slimy-hornets-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@actnowcoalition/ui-components": patch
+---
+
+Upgrade Storybook to 6.5.12
```
mikelehen: FWIW- We probably don't need to create changesets for things like this that have zero impact on our published packages (since storybook is just a dev dependency).  So you could delete this file if you wanted.  It doesn't really matter though. Up to you!

mikelehen (approved):

.changeset/slimy-hornets-suffer.md
pnavarrc: I was on the fence myself, going to remove the changeset file before merging 👍

PR #206, opened 2022-09-20
Implement share button

### LINKED ISSUE #170: Implement SocialSharing buttons
Let's implement them based on one of our more recent iterations in CAN, adapting to better match the MUI system.

Image

faihegberg:
Closes https://github.com/covid-projections/act-now-packages/issues/170


packages/ui-components/src/components/ShareButton/ShareButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+  quote,
+  hashtags = [],
+  onCopyLink,
+  onShareTwitter,
+  onShareFacebook,
+  justifyButton = "flex-start",
+}) => {
+  const [visibleSocialButtons, setVisibleSocialButtons] = useState(false);
+
+  // TODO (Fai): Implement useEscToClose in addition to ClickAwayListener.
```
mikelehen: Is this something we definitely intend to do?  If so then I'd either do it in this PR or open a github issue for it (and then you can change this to TODO(#123) or whatever).

packages/ui-components/src/components/ShareButton/FacebookShareButton.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import * as ReactShare from "react-share";
+import FacebookIcon from "@mui/icons-material/Facebook";
+import { Button } from "@mui/material";
+
+export const FacebookShareButton: React.FC<{
+  url: string;
+  quote: string;
+  onClick: () => void;
```
mikelehen, line 9: Is there a way to make this accept all of the same props as `ReactShare.FacebookShareButton`, including e.g. `hashtag` without having to duplicate the props manually (like we're doing with url and quote)?  I don't remember the syntax offhand but I feel like I've seen @pnavarrc do it.

packages/ui-components/src/components/ShareButton/TwitterShareButton.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import * as ReactShare from "react-share";
+import TwitterIcon from "@mui/icons-material/Twitter";
+import { Button } from "@mui/material";
+
+export const TwitterShareButton: React.FC<{
+  url: string;
+  quote: string;
+  hashtags?: string[];
```
mikelehen: Same feedback here.  It looks like `ReactShare.TwitterShareButton` accepts `title`, `via`, `related`, etc.  Is there a way to accept and pass-thru all of those automatically?

packages/ui-components/src/components/ShareButton/ShareButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const timer = setTimeout(() => setVisibleSocialButtons(false), 1000);
+    return () => clearTimeout(timer);
+  };
+
+  const toggleSocialButtons = () => {
+    setVisibleSocialButtons(!visibleSocialButtons);
+  };
+
+  return (
+    <ClickAwayListener onClickAway={() => hideSocialButtons()}>
```
mikelehen: The ClickAwayListener doesn't seem to be working for me, at least when I test in Storybook.  When I click "Share" the menu comes up, but it doesn't go away when I click outside of the share button / menu.

packages/ui-components/src/components/ShareButton/ShareButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+  onShareTwitter,
+  onShareFacebook,
+  justifyButton = "flex-start",
+}) => {
+  const [visibleSocialButtons, setVisibleSocialButtons] = useState(false);
+
+  // TODO (Fai): Implement useEscToClose in addition to ClickAwayListener.
+  const hideSocialButtons = () => {
+    const timer = setTimeout(() => setVisibleSocialButtons(false), 1000);
+    return () => clearTimeout(timer);
```
mikelehen: We don't seem to use the return value of `hideSocialButtons` anywhere, so I don't know if there's a reason to return the clearTimeout callback? 🤔

packages/ui-components/src/components/ShareButton/ShareButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+  };
+
+  const toggleSocialButtons = () => {
+    setVisibleSocialButtons(!visibleSocialButtons);
+  };
+
+  return (
+    <ClickAwayListener onClickAway={() => hideSocialButtons()}>
+      <Box display="flex" flexDirection="column">
```
mikelehen: FWIW- I feel like there should be an easier way to implement this than using flexbox, since it's just a single column of items.  Like maybe we could use a `Stack` with `alignItems` or `justifyContent`?  But not sure. My CSS is weak. 😬

packages/ui-components/src/components/ShareButton/ShareButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { FacebookShareButton } from "./FacebookShareButton";
+
+export interface ShareButtonProps {
+  url: string;
+  quote: string;
+  hashtags?: string[];
+  onCopyLink: () => void;
+  onShareTwitter: () => void;
+  onShareFacebook: () => void;
+  justifyButton?: "flex-start" | "flex-end" | "center";
```
mikelehen: I guess this is relating to the facebook/twitter/copy buttons, rather than the share button, so perhaps rename it to "justifySocialButtons" or something?

I also feel like using using the terms "justify" and "flex-start" / "flex-end" / "center" is kind of leaking implementation details about how this component is implemented (using flexbox).  I'd be tempted to rename to "alignSocialButtons" with options "left", "center", "right" but maybe I'm missing something.

packages/ui-components/src/components/ShareButton/ShareButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { TwitterShareButton } from "./TwitterShareButton";
+import { FacebookShareButton } from "./FacebookShareButton";
+
+export interface ShareButtonProps {
+  url: string;
+  quote: string;
+  hashtags?: string[];
+  onCopyLink: () => void;
+  onShareTwitter: () => void;
+  onShareFacebook: () => void;
```
mikelehen, line 15: I'm mildly curious if we have a use case in mind for these onCopyLink / onShareXyz callbacks.  It doesn't hurt to have them, but I wouldn't be surprised if we never use them?  Feel free to leave them though.

mikelehen (changes requested):
Generally looks good with a few nits / questions and an issue with the ClickAwayListener not seeming to work for me. 🤷

packages/ui-components/src/components/ShareButton/ShareButton.tsx
faihegberg: Pablo and I discussed this this afternoon and we figured to try experimenting with [MUI Menu](https://mui.com/material-ui/react-menu/) 😄 Partially because it can automatically handle closing the menu as the user clicks away.

packages/ui-components/src/components/ShareButton/ShareButton.tsx
faihegberg: Refer to comment above about the MUI Menu.

packages/ui-components/src/components/ShareButton/ShareButton.tsx
faihegberg: I'm a bit puzzled by this too. Simply calling `clearTimeout(timer)` won't actually close the menu. I see in our [CAN repo](https://github.com/covid-projections/covid-projections/blob/e77ae9904bbac4c81715adee9485fee7565f8abf/src/components/ShareButtons/ShareButtonGroup.tsx#L37-L40) that it also returns a callback but I'm not exactly sure why 🤔 Having it there appears to close the menu though.

packages/ui-components/src/components/ShareButton/ShareButton.tsx
faihegberg: [MUI menu](https://mui.com/material-ui/react-menu/#customization) to the rescue 😄

packages/ui-components/src/components/ShareButton/ShareButton.tsx
faihegberg: This applies even if we use MUI menu, except it has built in `anchorOrigin` and `transformOrigin` props which actually accept "left", "center", or "right" 😄 So I'll rename this prop to `menuOrigin`. Let me know if you think it's still unclear though.

packages/ui-components/src/components/ShareButton/ShareButton.tsx
faihegberg: Let's keep them for now. If after some implementation we don't end up using them at all, then we can remove them.

packages/ui-components/src/components/ShareButton/FacebookShareButton.tsx
faihegberg: As discussed on slack, I'm implementing your suggestion of using
```ts
type BaseProps = React.ComponentProps<typeof ReactShare.FacebookShareButton>;

export interface FacebookShareButtonProps extends BaseProps {
  onClick: () => void;
}

export const FacebookShareButton: React.FC<FacebookShareButtonProps> = ({
  onClick,
  ...otherProps
}) => ( ... );
```

Thanks again 😄

packages/ui-components/src/components/ShareButton/ShareButton.stories.tsx
```diff
@@ -1,14 +1,17 @@
...
+import { Box } from "@mui/material";
 
 export default {
   title: "Components/ShareButton",
   component: ShareButton,
 } as ComponentMeta<typeof ShareButton>;
 
 const Template: ComponentStory<typeof ShareButton> = (args) => (
-  <ShareButton {...args} />
+  <Box display="flex" justifyContent="center">
```
mikelehen, line 12: Out of curiosity, why the wrapper box?

packages/ui-components/src/components/ShareButton/ShareButton.style.tsx
```diff
@@ -0,0 +1,24 @@
+import { styled } from "../../styles";
+import {
+  Menu as MuiMenu,
+  MenuItem as MuiMenuItem,
+  Button as MuiButton,
+} from "@mui/material";
+
+export const Menu = styled(MuiMenu)`
+  .MuiMenu-paper {
+    margin-top: 8px;
```
mikelehen: Curious if there's a reason to use 8px instead of spacing(1) or whatever?

packages/ui-components/src/components/ShareButton/ShareButton.style.tsx
```diff
@@ -0,0 +1,24 @@
...
+    min-width: ${(props) => props.theme.spacing(18)};
```
mikelehen: This feels like an example where we might *not* want to use spacing.  I assume this min-width is based on the textual lengths of the button text, so that they're all the same length or something?  It's not obvious to me that it should be tied to the theme spacing. 🤷

packages/ui-components/src/components/ShareButton/ShareButton.tsx
```diff
@@ -23,63 +23,69 @@ export const ShareButton: React.FC<ShareButtonProps> = ({
...
+  const [anchorButton, setAnchorButton] = useState<null | HTMLElement>(null);
+  const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    setAnchorButton(event.currentTarget);
   };
-
-  const toggleSocialButtons = () => {
-    setVisibleSocialButtons(!visibleSocialButtons);
+  const handleClose = () => {
+    const timer = setTimeout(() => setAnchorButton(null), 1000);
+    return () => clearTimeout(timer);
```
mikelehen: I think you can safely remove this `return` since we're not using the return value of `handleClose()` anywhere.

mikelehen (approved):
Looks good with a couple nits!

packages/ui-components/src/components/ShareButton/ShareButton.stories.tsx
faihegberg: This is just to place the button at the center in storybook and better illustrate the menu being left, center, or right justified :)

packages/ui-components/src/components/ShareButton/ShareButton.style.tsx
faihegberg: Good catch!

PR #210, opened 2022-09-21
Stylized search item

chasulin:
[Search item with colored indicator](https://act-now-packages--pr210-casulin-search-item-0aunveuo.web.app/storybook/index.html?path=/story/components-searchitem--new-york-with-indicator)
[Search item without colored indicator](https://act-now-packages--pr210-casulin-search-item-0aunveuo.web.app/storybook/index.html?path=/story/components-searchitem--new-york)
[States search](https://act-now-packages--pr210-casulin-search-item-0aunveuo.web.app/storybook/index.html?path=/story/components-regionsearch--states-only)

### DIFF
--- a/packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
+++ b/packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
@@ -25,6 +25,10 @@ export type CustomAutocompleteProps = AutocompleteProps<
 
 export interface RegionSearchProps
   extends Omit<CustomAutocompleteProps, "renderInput"> {
+  LinkComponent: React.FC<{
+    children: React.ReactElement;
+    targetUrl: string;
+  }>;
   /** Placeholder text to show in the inner text field  */
   inputLabel?: string;
   /** Optional renderInput function. See https://mui.com/material-ui/api/autocomplete/ */
@@ -35,6 +39,7 @@ export const RegionSearch: React.FC<RegionSearchProps> = ({
   options,
   inputLabel = "City, county, state, or district",
   renderInput: customRenderInput,
+  LinkComponent,
   ...otherAutocompleteProps
 }) => {
   const defaultRenderInput: CustomAutocompleteProps["renderInput"] = (
@@ -62,12 +67,13 @@ export const RegionSearch: React.FC<RegionSearchProps> = ({
           option: Region
         ) => {
           return (
-            <SearchItem
-              itemLabel={option.shortName}
-              itemSublabel={`${option.population} population`}
-              itemUrl={option.relativeUrl}
-              {...props}
-            />
+            <LinkComponent targetUrl={option.relativeUrl}>
+              <SearchItem
+                itemLabel={option.shortName}
+                itemSublabel={`${option.population} population`}
+                {...props}
+              />
+            </LinkComponent>
           );
         }}
         getOptionLabel={stringifyOption}
--- a/packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
+++ b/packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
@@ -1,9 +1,9 @@
-import React from "react";
-import { Container } from "./RegionSearch.style";
+import React, { HTMLAttributes } from "react";
 import { Region } from "@actnowcoalition/regions";
 import { Autocomplete, AutocompleteProps, TextField } from "@mui/material";
 import SearchIcon from "@mui/icons-material/Search";
 import { createFilterOptions } from "@mui/material/useAutocomplete";
+import { SearchItem } from "./SearchItem/SearchItem";
 
 function stringifyOption(region: Region) {
   return region.fullName;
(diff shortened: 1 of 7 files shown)


packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
```diff
@@ -51,19 +51,32 @@ export const RegionSearch: React.FC<RegionSearchProps> = ({
...
         clearIcon={<></>}
         renderInput={customRenderInput ?? defaultRenderInput}
+        renderOption={(
+          props: HTMLAttributes<HTMLLIElement>,
+          option: Region
+        ) => {
+          return (
+            <SearchItem
+              itemLabel={option.shortName}
+              itemSublabel={`${option.population} population`}
```
pnavarrc: Consider formatting the population to have thousands separators (and maybe rounding the number?)

packages/ui-components/src/components/RegionSearch/SearchItem/SearchItem.style.ts
```diff
@@ -0,0 +1,36 @@
...
+`;
+
+export const Container = styled(Box)`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  padding: ${({ theme }) => theme.spacing(1, 1.5)};
+
+  &:hover {
+    background-color: ${({ theme }) => theme.palette.grey[100]};
```
pnavarrc: Consider using `theme.action.hover` instead (similar for the active state)

packages/ui-components/src/components/RegionSearch/SearchItem/SearchItem.style.ts
```diff
@@ -0,0 +1,36 @@
...
+  margin-top: ${({ theme }) => theme.spacing(0.75)};
+  margin-right: ${({ theme }) => theme.spacing(1)};
+`;
+
+export const ArrowIcon = styled(ArrowForwardIosIcon)`
+  font-size: 14px;
+  color: ${({ theme }) => theme.palette.common.black};
+`;
+
+export const Container = styled(Box)`
```
pnavarrc, line 22: Do we need the container? Maybe we can just style the `Stack` element instead.

packages/ui-components/src/components/RegionSearch/SearchItem/SearchItem.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <Container>
+        <Stack direction="row">
+          {iconColor && <CircleIcon iconColor={iconColor} />}
+          <Stack spacing={0.5}>
+            <Typography variant="labelLarge">{itemLabel}</Typography>
+            <Typography variant="paragraphSmall">{itemSublabel}</Typography>
+          </Stack>
+        </Stack>
+        <ArrowIcon />
+      </Container>
```
pnavarrc: Consider removing the container and using `Stack` elements instead

packages/ui-components/src/components/RegionSearch/SearchItem/SearchItem.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export const SearchItem: React.FC<SearchItemProps> = ({
+  itemLabel,
+  itemSublabel,
+  itemUrl,
+  iconColor,
+}) => {
+  return (
+    <WrappingLink href={itemUrl}>
```
pnavarrc: The `RegionSearch` component will need the ability to receive a `LinkComponent` (or some other solution) so it can use whichever `Link` component the parent app uses:

```tsx
import Link from "next/link"
// or alternatively, for CAN:
// import { Link } from "react-router";

<RegionSearch 
  regions={regions}
  LinkComponent={Link}
/>
```

So maybe it would be better to not have the link in the `SearchItem` component (and remove the `itemUrl` prop) and make this a concern of the `RegionSearch` component:

```tsx
// ...
<Autocomplete
  options={options}
  onChange={(e, item: Region | null) => onChange(item)}
  clearIcon={<></>}
  renderInput={customRenderInput ?? defaultRenderInput}
  renderOption={(
    props: HTMLAttributes<HTMLLIElement>,
    option: Region
    ) => (
      <LinkComponent href={region.relativeUrl}>
        <SearchItem
          itemLabel={option.shortName}
          itemSublabel={`${option.population} population`}
        />
      </LinkComponent>
    }
// ...
```

that will allow us to not pass `LinkComponent` and `itemUrl` to the `SearchItem` (instead of passing both). 

Figuring out how to pass the app `Link` component to the RegionSearch, Map and CompareTable components is out of the scope of this PR, but maybe moving the link out can simplify things.

packages/ui-components/src/components/RegionSearch/SearchItem/SearchItem.style.ts
```diff
@@ -0,0 +1,36 @@
...
+  font-size: 12px;
+  color: ${({ iconColor }) => iconColor};
+  margin-top: ${({ theme }) => theme.spacing(0.75)};
+  margin-right: ${({ theme }) => theme.spacing(1)};
+`;
+
+export const ArrowIcon = styled(ArrowForwardIosIcon)`
+  font-size: 14px;
+  color: ${({ theme }) => theme.palette.common.black};
+`;
```
pnavarrc, line 20: Maybe we can create or export the (admitedly extremely simple) `Dot` component from https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/components/MetricDot/MetricDot.style.ts#L3 and reuse it here (or reuse this one and use it there)?

pnavarrc (approved):
Added some suggestions, looks good!

packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
chasulin: yeah our numbers utils don't seem to have rounding functionality, so i was gonna add that there and then implement here

packages/ui-components/src/components/RegionSearch/SearchItem/SearchItem.style.ts
chasulin: hm not much difference between making this `Box` vs `Stack` i don't think? its in the style file because we need to declare hover state styles

packages/ui-components/src/components/RegionSearch/SearchItem/SearchItem.tsx
chasulin: [ditto here](https://github.com/covid-projections/act-now-packages/pull/210#discussion_r978000072)

PR #213, opened 2022-09-21
Fix legend threshold labels

### LINKED ISSUE #207: Fix the alignment of start and end labels on MetricLegendThreshold
When using `MetricLegendThreshold` in the hackathon repo (https://github.com/covid-projections/hackathon-august-2022/pull/6), the start and end labels are not aligned with the bars by default.

Image

Note: This doesn't happen in Storybook 🤷 

Update: The component is not showing the labels, I introduced that bug 😅

pnavarrc:
Close #207  - The issue was actually that by default the labels are shown, but they were not visible because of a bug I introduced when I refactored the component to use `RectClipGroup`. The fix is to move the labels (and tick marks) outside `RectClipGroup`

**Before - No labels** 😢

**After**

image

### DIFF
--- a/.changeset/calm-cobras-cry.md
+++ b/.changeset/calm-cobras-cry.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/ui-components": patch
+---
+
+Fix a a bug that prevented the labels to show on LegendThreshold (horizontal)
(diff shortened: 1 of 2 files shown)


mikelehen (approved):
✅

PR #218, opened 2022-09-22
Style Chip component

### LINKED ISSUE #212: Style the MUI Chip component
We need to style the MUI Chip component to better match the mocks - [ActNow Design System - Chip](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=72%3A2447)

Image

Instead of implementing a new component, I think it makes sense to customize the Chip component in [`styles/theme/components.ts`](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/styles/theme/components.ts#L43) (see the customizations for `MuiToggleGroup` and `MuiButton` in that file) and to create a story for it under `stories/` (see [Button.stories.tsx](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/stories/Buttons.stories.tsx#L1))

- https://mui.com/material-ui/react-chip/

pnavarrc:
Close #212 - Style the MUI Chip component

- [ActNow Design System - Chip (Tag)](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=72%3A2447)
- [Storybook → Chip → Clickable & Deletable](https://act-now-packages--pr218-pablo-chip-component-vi2ogvue.web.app/storybook/index.html?path=/story/design-system-chip--clickable-and-deletable)

image

I discovered a way to use all the theme variables when styling a component. Instead of importing the bare `typography` and `palette` objects, we can pass a callback to the styles of each component. 

Before this change, only the `palette` variables that we customized were available, not the ones inherited from MUI (for example, `theme.palette.grey[100]` or `theme.palette.text.disabled` were not available). Using the callback syntax makes all the colors available (similar for the typography).

I also added the hex codes to the stories for Color, so we can more easily match the mocks when they reference colors with a different name.

Note that we only have designs for the outlined version, so I only styled that one.

### DIFF
--- a/packages/ui-components/src/styles/theme/components.ts
+++ b/packages/ui-components/src/styles/theme/components.ts
@@ -81,6 +81,25 @@ const components: ThemeOptions["components"] = {
       },
     },
   },
+  MuiChip: {
+    styleOverrides: {
+      root: {
+        border: `solid 1px ${palette.border.default}`,
+      },
+      label: {
+        fontFamily: typography.labelSmall.fontFamily,
+        fontSize: typography.labelSmall.fontSize,
+        fontWeight: typography.labelSmall.fontWeight,
+        lineHeight: typography.labelSmall.lineHeight,
+        color: palette.secondary.dark,
+      },
+      deleteIcon: {
+        ":hover": {
+          color: palette.secondary.dark,
+        },
+      },
+    },
+  },
 };
 
 export default components;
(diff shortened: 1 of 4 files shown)


packages/ui-components/src/styles/theme/components.ts
```diff
@@ -1,71 +1,69 @@
...
-import typography, { typographyConstants } from "./typography";
-import palette from "./palette";
 
 const referenceTheme = createTheme();
 
 const components: ThemeOptions["components"] = {
   MuiButton: {
     styleOverrides: {
-      root: {
+      root: ({ theme }) => ({
```
chasulin, line 9: this is excelllent!

chasulin (approved):
✨

PR #221, opened 2022-09-26
Implement tabs

### LINKED ISSUE #96: Implement Tab component
I would recommend to implement a non-metric aware version first with the styles of the tab (width, border-bottom, styles for the active tab, etc) and then implement the metric-aware version, probably by customizing the MuiTab component in the [`src/styles/theme/components.ts`](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/styles/theme/components.ts) and creating a story file under `src/stories` (similar to `ToggleButton`, `Button`, etc)

Image

- [ActNow Design System](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=720%3A22518)

faihegberg:
Closes https://github.com/covid-projections/act-now-packages/issues/96

Notes:
- For "tabs" that are not clickable, we decided to not implement them using MUI tabs. This PR only focuses on clickable tabs.
- The tab component is not supposed to be aware of its content. So we're leaving the dot inside the `MetricValue` component as its original color instead of greyed out when one selects a different tab.

### DIFF
--- a/packages/ui-components/src/styles/theme/components.ts
+++ b/packages/ui-components/src/styles/theme/components.ts
@@ -4,6 +4,38 @@ import { ThemeOptions, createTheme } from "@mui/material";
 const referenceTheme = createTheme();
 
 const components: ThemeOptions["components"] = {
+  MuiTabs: {
+    styleOverrides: {
+      indicator: {
+        height: "3px",
+        backgroundColor: "black",
+      },
+      flexContainer: ({ theme }) => ({
+        borderBottom: `1px solid ${theme.palette.border.default}`,
+        width: "fit-content",
+      }),
+    },
+  },
+  MuiTab: {
+    styleOverrides: {
+      root: ({ theme }) => ({
+        textTransform: "none",
+        padding: "0px 0px 16px 0px",
+        margin: "0px 13px",
+        ":first-of-type": {
+          marginLeft: "0px",
+        },
+        ":last-of-type": {
+          marginRight: "0px",
+        },
+        ":not(&.Mui-selected)": {
+          span: {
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/styles/theme/components.ts
```diff
@@ -4,6 +4,38 @@ import { ThemeOptions, createTheme } from "@mui/material";
...
+    },
+  },
+  MuiTab: {
+    styleOverrides: {
+      root: ({ theme }) => ({
+        textTransform: "none",
+        padding: `0px 0px ${theme.spacing(2)} 0px`,
+        margin: `0px ${theme.spacing(2)}`,
+        ":first-of-type": {
+          marginLeft: "0px",
```
pnavarrc: nit: `0` is equivalent to `"0px"` (zero is zero in any unit), consider changing to `0`

packages/ui-components/src/styles/theme/components.ts
```diff
@@ -4,6 +4,38 @@ import { ThemeOptions, createTheme } from "@mui/material";
 const referenceTheme = createTheme();
 
 const components: ThemeOptions["components"] = {
+  MuiTabs: {
+    styleOverrides: {
+      indicator: ({ theme }) => ({
+        height: theme.spacing(0.375),
```
pnavarrc: nit: I think that hardcoding the pixel value in this one case is better than using spacing units (kind of hard to make the conversion)

pnavarrc (approved):
Looks good, thanks for including metric-based examples!

Issue #220, opened 2022-09-26
Update the map components to use app regions


pnavarrc:
The maps are currently not using the app regions (the regions are imported directly), which is causing the URLs not to match (since they are overwritten at the app level).

Michael proposed some ideas:

- We could create a context to allow components to access the `RegionDB` object at the app level, for this we will need to create a context provider and a hook to access the regions DB.
- The RegionDB class could take a function to configure the `relativeUrl`, the current solution we have feels a bit hacky

```tsx
import { RegionDB, Region, states, counties } from '@actnowcoalition/regions';

const regions = new RegionDB([...states.all, ...counties.all], { 
  regionUrl: (region: Region):string => `/us/${region.slug}`
});
```

PR #222, opened 2022-09-26
regions: Update RegionDB to take an optional relativeUrl function

pnavarrc:
Part of #220 - Updating the map components to take app regions. This will require updating a few different parts, so this PR is just to make the `relativeUrl` customizable when initializing the `RegionDB`.

**Example:**

```tsx
// src/utils/regions.ts

const regions = new RegionsDB([...states.all], {
  relativeUrl: (region:Region) => `/us/states/${region.slug}`
});

// ...

const washington = regions.findByRegionIdStrict('53');
washington.relativeUrl  // "/us/states/washington-wa"
```

### DIFF
--- a/.changeset/old-lizards-turn.md
+++ b/.changeset/old-lizards-turn.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/regions": minor
+---
+
+RegionDB receives options to customize the relativeUrl region
--- a/packages/regions/src/Region.ts
+++ b/packages/regions/src/Region.ts
@@ -22,7 +22,7 @@ export class Region {
     public readonly slug: string,
     public readonly relativeUrl: string,
     public readonly parent: Region | null,
-    // TODO: transfrom population and potentially other attributes to Metric.
+    // TODO: transform population and potentially other attributes to Metric.
     // This will allow us to leverage on more advanced Metric functionality,
     // such as display sources, number formatting options, etc.
     public readonly population: number
@@ -38,7 +38,7 @@ export class Region {
 
   /**
    * Generates a slug from the input string by replacing accented
-   * characters with their basic Latin equivalents, replacing whitespaces
+   * characters with their basic Latin equivalents, replacing white spaces
    * with underscores and splitting words when capitalization changes.
    *
    * @example
--- a/packages/regions/src/RegionDB.ts
+++ b/packages/regions/src/RegionDB.ts
@@ -2,11 +2,30 @@ import keyBy from "lodash/keyBy";
 import { assert } from "@actnowcoalition/assert";
 import { Region } from "./Region";
 
+export interface RegionDBOptions {
+  /** Customize the relativeUrl property of each region in this RegionDB */
+  relativeUrl: (region: Region) => string;
+}
+
+const defaultOptions: RegionDBOptions = {
+  relativeUrl: () => "",
+};
+
 export class RegionDB {
   private regionsById: { [regionId: string]: Region };
 
-  constructor(private regions: Region[]) {
-    this.regionsById = keyBy(regions, (region: Region) => region.regionId);
+  constructor(private regions: Region[], private options?: RegionDBOptions) {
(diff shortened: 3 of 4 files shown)


packages/regions/src/Region.ts
```diff
@@ -22,7 +22,7 @@ export class Region {
     public readonly slug: string,
     public readonly relativeUrl: string,
     public readonly parent: Region | null,
-    // TODO: transfrom population and potentially other attributes to Metric.
+    // TODO: transform population and potentially other attributes to Metric.
```
pnavarrc, line 25: Installed a spell checker in VSCode :D

.changeset/old-lizards-turn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@actnowcoalition/regions": minor
+---
+
+RegionDB receives options to customize the relativeUrl region
```
mikelehen, line 5: nit: "... to customize the relativeUrls of regions." or something?

packages/regions/src/RegionDB.ts
```diff
@@ -2,11 +2,30 @@ import keyBy from "lodash/keyBy";
 import { assert } from "@actnowcoalition/assert";
 import { Region } from "./Region";
 
+export interface RegionDBOptions {
+  /** Customize the relativeUrl property of each region in this RegionDB */
+  relativeUrl: (region: Region) => string;
```
mikelehen, line 7: I feel like this should be `getRelativeUrl` or `relativeUrlGetter` or something to indicate that it's a function that returns relative URLs?  If you disagree / know of a convention this is following, feel free to push back.

packages/regions/src/RegionDB.ts
```diff
@@ -2,11 +2,30 @@ import keyBy from "lodash/keyBy";
...
+}
+
+const defaultOptions: RegionDBOptions = {
+  relativeUrl: () => "",
```
mikelehen, line 11: I almost wonder if this should `fail()` instead.  Is there a case where returning an empty string is helpful?

packages/regions/src/RegionDB.ts
```diff
@@ -2,11 +2,30 @@ import keyBy from "lodash/keyBy";
...
+
 export class RegionDB {
   private regionsById: { [regionId: string]: Region };
 
-  constructor(private regions: Region[]) {
-    this.regionsById = keyBy(regions, (region: Region) => region.regionId);
+  constructor(private regions: Region[], private options?: RegionDBOptions) {
+    const { relativeUrl } = { ...defaultOptions, ...options };
+
+    // Reconstruct the region array using the provided options
```
mikelehen, line 20: Hrm.  This isn't quite what I was imagining, though it may be another alternative that works.  But it seems like this ends up being equivalent to our existing code where we end up with a RegionDB with "mutated" regions that are different than the regions imported by the map.  So I'm curious what you are imagining for how we actually fix the map.  Would it no longer import states / counties directly somehow?

What I was imagining (I think) was that RegionDB would expose the `relativeUrl()` function as a member, and the map would do something like `regionDb.relativeUrl(region)` to get the URL for a given region.

That way we are never re-creating `Region` instances and the global `RegionDB` used in the app handles "routing" of regions to URLs.  

That said, it does mean we'd need to pass in `regionDB` to the map and any other components that need to generate URLs, which could be cumbersome (unless we put it in a react context?).

Anyway, let me know what you are imagining.  I could likely be convinced this will work out.

mikelehen (changes requested):
Not 100% sure I'm on the same page regarding the ultimate plan here.  Feel free to ping me on slack or whatever if that's easier.

mikelehen (approved):
Actually, based on your slack comments I think you *are* planning to make it so the map is driven by RegionDB and doesn't import states / counties directly.  So if we think we can make that happen, then I think this change is probably fine.

pnavarrc:
I'm going to close this PR and implement the other solution, thanks for reviewing @mikelehen !

Issue #224, opened 2022-09-27
Find a solution so the map components access the app-level RegionDB


pnavarrc:
work in progress https://www.dropbox.com/scl/fi/cv1o8tqqh3m5yk8gb9it4/Refactoring-Maps.paper?dl=0&rlkey=dscnn11usno213fi8n0dtml6z

pnavarrc:
Done

PR #226, opened 2022-09-27
Update maps to receive regionDB instead of importing regions directly

pnavarrc:
Part of https://github.com/covid-projections/act-now-packages/issues/224 - Update the maps so they don't need to import regions directly. There are two main changes here:

- `getFillColor` now takes `regionId: string` instead of `Region`
- metric-aware maps receive `regionDB` as a prop instead of importing the regions directly

I also removed the links in the maps so we can get the URL from the app-level `RegionDB`, once that's implemented. Updated stories accordingly too.

### DIFF
--- a/packages/ui-components/src/common/utils/maps.ts
+++ b/packages/ui-components/src/common/utils/maps.ts
@@ -1,4 +1,4 @@
-import { counties, Region } from "@actnowcoalition/regions";
+import { RegionDB, Region } from "@actnowcoalition/regions";
 
 /** Checks if a county or congressional district belongs to a given state */
 export function belongsToState(
@@ -11,8 +11,11 @@ export function belongsToState(
 }
 
 /** Returns an array of all counties of a given state */
-export function getCountiesOfState(stateRegionId: string): Region[] {
-  return counties.all.filter(
+export function getCountiesOfState(
+  regionDB: RegionDB,
+  stateRegionId: string
+): Region[] {
+  return regionDB.all.filter(
     (county) => county.parent?.regionId === stateRegionId
   );
 }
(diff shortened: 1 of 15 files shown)


packages/ui-components/src/components/Maps/USNationalMap/MetricUSNationalMap.stories.tsx
```diff
@@ -31,4 +32,5 @@ MetricAwareCounties.args = {
   renderTooltip: (regionId: string) => renderSimpleTooltip(regionId),
   metric: MetricId.MOCK_CASES,
   showCounties: true,
+  regionDB: counties,
```
chasulin: this should be counties and states, no?

packages/ui-components/src/common/utils/maps.ts
```diff
@@ -11,8 +11,11 @@ export function belongsToState(
 }
 
 /** Returns an array of all counties of a given state */
-export function getCountiesOfState(stateRegionId: string): Region[] {
-  return counties.all.filter(
+export function getCountiesOfState(
+  regionDB: RegionDB,
+  stateRegionId: string
+): Region[] {
+  return regionDB.all.filter(
```
chasulin, line 18: hm the point of this utility function is to return all of the counties that belong to a given state. the filter shouldn't depend on which regions are passed in as regionDB. it should always return the same thing, given a stateRegionId

chasulin:
im a bit confused about this pr overall. if the point is to use the app-specific regions, maybe we should just be grabbing them from `utils/regions` within the app instead of the regions package directly?

chasulin (approved):

pnavarrc:
As we discussed on our call, maybe a better idea is to pass states and counties as RegionDB instances separately to the US map components so the type of region DB is more explicit. I will work on that when I try to separate the county and state maps. Thanks for the feedback @chasulin !

PR #230, opened 2022-09-27
Remove meta tags

faihegberg:
Moving social meta tags from this packages repo to act now template repo instead. These meta tags rely heavily on the `Head` component of `next/head`. So they are better suited for the template repo, which would have nextJS implemented. Keeping them here would prevent some meta tags (e.g. `og:title`) from rendering under the `Head` component properly.

More info at the bottom of this doc: https://nextjs.org/docs/api-reference/next/head


pnavarrc (approved):
Looks good!

PR #231, opened 2022-09-28
Ensure we don't have multiple fetches for the same CSV going at once.

mikelehen:
I noticed that when you go to https://hackathon-august-2022.vercel.app/ there are dozens of requests for monkeypox.csv in the network tab of chrome dev tools.  While we were caching the response already, there was a window of time where we are waiting for the response and more metric fetches could being triggered.

What we need to do is cache the *promise* so as long as the CSV fetch has been initiated we won't start a new one.

### DIFF
--- a/packages/metrics/src/data_providers/CsvDataProvider.ts
+++ b/packages/metrics/src/data_providers/CsvDataProvider.ts
@@ -41,6 +41,7 @@ export class CsvDataProvider extends CachingMetricDataProviderBase {
   private readonly dateColumn?: string;
   private readonly url?: string;
   private readonly csvText?: string;
+  private fetchedTextPromise: Promise<string> | undefined;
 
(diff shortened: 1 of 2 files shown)


packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -115,6 +119,7 @@ export class CsvDataProvider extends CachingMetricDataProviderBase {
    */
   private async fetchCsvText(): Promise<string> {
     assert(this.url, "URL must be specified in order to use fetchCsvText()");
+    console.log("Fetching", this.url);
```
mikelehen, line 122: Intentionally leaving this in since it could be handy for debugging, though we should maybe revisit in the future.  Ideally we might have a debug logging flag that we can turn on/off to enable/disable this kind of logging.

smcclure17 (approved):
ah, hmm 😭😕 thanks for this fix!!

PR #233, opened 2022-09-28
demo: RegionLink component

### LINKED ISSUE #208: Pass link component to components that need it
Some of the components need to render links to locations (`USNationalMap`, `MetricCompareTable`, `RegionSearch`?), but the type of link component to be used depends on the type of link that the app is using. For example, for Next applications, we want to use Link from `next/link`, whereas for CAN we would use `Link` from `react-router`.

One solution would be to add a prop to the components so we can pass the link component to use (and have the native `a` element being the default)

```tsx
import Link from "next/link";
// ...

<MetricCompareTable
  regions={regionList}
  metrics={[...]}
  linkComponent={Link}
/>
```

Ideally, we want all the components that need a link to have a consistent interface to receive the component.

pnavarrc:
Close #208  - In some components (maps, metric-aware compare table), we need to add links to location pages on some of the inner elements of the components. For the map, for example, each shape should link to location pages.

The _link_ element to be used depends on how routing works for that particular application. For CAN, for example, we use the `Link` component from `react-router`, on the template, we use `Link` from `next/link`, and on other apps we might want to use the native anchor element `a`. One option to solve this is to pass a component (with a given interface) as a prop, so each app can pass the type of link that they want (and attach additional props too). The [MUI Accordion](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Accordion/Accordion.js) component uses a similar pattern to allow users to customize the transition.

This demo adds a `RegionLinkComponent` prop that can be used to wrap the linking element to the corresponding region page, without tying it to a specific type of link component. For example, if we wanted to use `MetricCompareTable` using the Link from `react-router`, we could:

```tsx
import { Link } from "react-router";

<MetricCompareTable
  regions={regions}
  metrics={metrics}
  RegionLinkComponent={({ region, children }) => (
    <Link to={regionDB.getRegionUrl(region)} aria-label={region.shortName}>
      {children}
    </Link>
  )}
/>
```

Looking for feedback on the pattern for now. On the implementation side, we will need to style the table cell and the link a bit to make sure that we have good hover states (for the row) and to ensure that the link covers the entire cell.

### DIFF
--- a/packages/ui-components/src/components/MetricCompareTable/MetricCompareTable.stories.tsx
+++ b/packages/ui-components/src/components/MetricCompareTable/MetricCompareTable.stories.tsx
@@ -1,9 +1,9 @@
 import React from "react";
 import { ComponentStory, ComponentMeta } from "@storybook/react";
 import { Paper } from "@mui/material";
-import { states } from "@actnowcoalition/regions";
+import { states, Region } from "@actnowcoalition/regions";
 import { MetricId } from "../../stories/mockMetricCatalog";
-import { MetricCompareTable } from ".";
+import { MetricCompareTable, RegionLinkComponentProp } from ".";
 
 export default {
   title: "Metrics/MetricCompareTable",
@@ -16,8 +16,31 @@ const Template: ComponentStory<typeof MetricCompareTable> = (args) => (
   </Paper>
 );
 
+const RegionLinkComponent: RegionLinkComponentProp = ({
+  children,
+  region,
+}: {
+  children: React.ReactNode;
+  region: Region;
+}) => (
+  <a
+    href={`/states/${region.slug}`}
+    aria-label={region.shortName}
+    style={{ display: "block", textDecoration: "none", color: "initial" }}
+  >
+    {children}
+  </a>
+);
+
 export const Example = Template.bind({});
 Example.args = {
+  regions: states.all,
+  metrics: [MetricId.PI, MetricId.MOCK_CASES, MetricId.PASS_FAIL],
+  RegionLinkComponent,
+};
+
+export const WithoutRegionLink = Template.bind({});
+WithoutRegionLink.args = {
   regions: states.all,
   metrics: [MetricId.PI, MetricId.MOCK_CASES, MetricId.PASS_FAIL],
 };
--- a/packages/ui-components/src/components/MetricCompareTable/MetricCompareTable.tsx
+++ b/packages/ui-components/src/components/MetricCompareTable/MetricCompareTable.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useState, Fragment } from "react";
 import {
   ColumnDefinition,
   CompareTable,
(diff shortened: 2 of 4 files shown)


packages/ui-components/src/components/MetricCompareTable/MetricCompareTable.stories.tsx
```diff
@@ -16,8 +16,31 @@ const Template: ComponentStory<typeof MetricCompareTable> = (args) => (
...
+  >
+    {children}
+  </a>
+);
+
 export const Example = Template.bind({});
 Example.args = {
   regions: states.all,
   metrics: [MetricId.PI, MetricId.MOCK_CASES, MetricId.PASS_FAIL],
+  RegionLinkComponent,
```
mikelehen, line 39: Naively it seems like it would be simpler to just make this a `LinkComponent` (that takes an href, not a region) and have the compare table pass in the correct href and aria-label correctly based on the region itself.  Then to use the compare table you could hopefully just do:

```
<MetricCompareTable metrics={...} regions={...} LinkComponent={Link} />
```

or something like that.  And that pattern would work for any component, even if it's going to be linking to things besides regions.

Maybe I'm missing some of the rationale though?

pnavarrc: That could work, but we will need to define the interface of the `LinkComponent` (in a similar way as the MUI Accordion component defines the API of acceptable `TransitionProps` https://mui.com/material-ui/transitions/#transitioncomponent-prop components) and we will need to pass the `RegionDB` instance (or at least the `RegionDB.getRegionUrl` method) to the `MetricCompareTable`.

On one hand, we could define the `LinkComponent` interface  to include `href`, `children` and maybe `aria-label`, but if later on we discover that we need additional properties (`onClick`, `className`, etc), we will need to update the interface and implementation of the compare table.

On the other hand, defining the `LinkComponent` interface will be beneficial because the implementation of a client will be much simpler (in most cases we could just pass the next `Link` component) - for users using a non-compliant `Link` component (react-router) they could easily create a component that adapts the props to be compliant.

Now that I'm writing this I think that passing the `RegionDB` instance and defining the `LinkComponent` interface is simpler and also more consistent across components (we will need to pass the `RegionDB` to the map components anyways, see https://github.com/covid-projections/act-now-packages/pull/226). 

I'm going to wait for more feedback, but I'm leaning towards the suggestion. Thanks @mikelehen !

mikelehen: Actually, does https://mui.com/material-ui/guides/routing/#global-theme-link solve this for us?  It looks like maybe we can configure the link component through the MUI theme?

If that doesn't work, I had one other related thought to share.  I keep being tempted to leverage ReactContext more given we are running into more "common things that components might need."  So we could consider wrapping everything up like:

```
<ActNowContext
  metricCatalog={...}
  regionDb={...}
  LinkComponent={...}>
  
   ... now you can use any of our components
   
</ActNowContext>
```

And FWIW- `LinkComponent` could default to a dummy version that uses `<a>` tags directly. 🤷

pnavarrc:
I think we found a better solution (customizing MuiLink at the theme level), closing!

PR #234, opened 2022-09-28
Fix MetricCompareTable, MetricDot, and MetricValue to not fetch timeseries data.

mikelehen:
Also change the default for includeTimeseries to false when fetching data.


smcclure17 (approved):
lgtm! 🙏

PR #237, opened 2022-09-28
AxesTimeseries component

chasulin:
Composed component of AxisBottom and AxisLeft. It's specific to timeseries, since the x axis explicitly uses a date scale.

[Storybook](https://act-now-packages--pr237-casulin-axes-fhx3yiox.web.app/storybook/index.html?path=/story/charts-axes--default-settings)


packages/ui-components/src/common/utils/charts.ts
```diff
@@ -0,0 +1,30 @@
...
+    return formatDateTime(date, DateFormat.MMM_D);
+  } else {
+    // Shows the year if the tick is in January (0) or December (11)
+    const dateFormat =
+      date.getMonth() === 0 || date.getMonth() === 11
+        ? DateFormat.MMM_YY
+        : DateFormat.MMM;
+    return formatDateTime(date, dateFormat).replace(" ", "'");
+  }
+};
```
chasulin: grabbed these utils from CAN

pnavarrc:
I would recommend using the default `tickValues` and `tickFormat` until we implement better versions of both functions (we did it in a rush at the time, when we had more than a year of data). Currently, the component doesn't render well if we have shorter or longer time spans (try starting on 2020-01-01 or 2022-06-01).

One option for now is to set `numTicks` to 8 or so, and to remove the custom `tickValues` and `tickFormat` so we can write and test better implementations of those functions.

pnavarrc (approved):
Looks good!

PR #238, opened 2022-09-28
Implement `ChartOverlayX` component

pnavarrc:
This component is necessary to render tooltips that depend on the horizontal position of the cursor only. Mostly taken from CAN, there is an example on how it can be used to add tooltips/tracking lines.

image

### DIFF
--- a/packages/ui-components/src/components/ChartOverlayX/ChartOverlayX.stories.tsx
+++ b/packages/ui-components/src/components/ChartOverlayX/ChartOverlayX.stories.tsx
@@ -1,0 +1,61 @@
+import React, { useState } from "react";
+import { scaleUtc } from "@visx/scale";
+import { Group } from "@visx/group";
+import { AxisBottom } from "../Axis";
+import { formatUTCDateTime, DateFormat } from "@actnowcoalition/time-utils";
+import { ComponentMeta } from "@storybook/react";
+import { ChartOverlayX } from ".";
+
+export default {
+  title: "Charts/ChartOverlayX",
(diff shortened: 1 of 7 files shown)


packages/ui-components/src/components/ChartOverlayX/ChartOverlayX.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+              stroke="#ccc"
+              strokeWidth="1"
+              pointerEvents="none"
+            />
+            <text y={height / 2} textAnchor="middle">
+              {formatUTCDateTime(date, DateFormat.YYYY_MM_DD)}
+            </text>
+          </Group>
+        )}
+        <ChartOverlayX
```
mikelehen, line 50: This name doesn't really convey what this is for to me.  It also sort of suggests that we should / might have a `ChartOverlayY` as well, which I'm guessing we won't (if we wanted that we should probably just combine it into this right now).  So maybe `HoverOverlay` or `HoverHandlerOverlay` or something?

mikelehen (approved):
Looks good to me!

packages/ui-components/src/components/ChartOverlayX/ChartOverlayX.stories.tsx
pnavarrc: there will be an XY equivalent for Trends. In that case, you need to track both the x and y cursor coordinates to determine which line/point to highlight and attach the tooltip to. I'm going to ship for now, but I agree that we might want to rename to something better.

PR #239, opened 2022-09-28
Metric timeseries chart component

chasulin:
Adds 2 components:
- `TimeseriesChart`
- `MetricTimeseriesChart` (wrapper component that passes `Metric` and `Region` down to inner component, `TimeseriesChart`) -- [storybook, New York mock case data](https://act-now-packages--pr239-casulin-metric-chart-b3daqs5z.web.app/storybook/index.html?path=/story/charts-metrictimeserieschart--new-york-mock-cases)


packages/ui-components/src/components/MetricTimeseriesChart/MetricTimeseriesChart.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  if (!timeseries.hasData()) {
+    return null;
+  }
+
+  return (
+    <TimeseriesChart
+      width={width}
+      height={height}
+      timeseries={timeseries as NonEmptyTimeseries<number>}
```
chasulin: cc @mikelehen - explicitly casting the type here with `<number>` was necessary because [maxValue in the yScale's domain](https://github.com/covid-projections/act-now-packages/pull/239/files#diff-9be9c8c845ce512b778dff10876d82e5da96eb4573a664db097752dbbd9b4b76R27-R30) needs to be composed of numbers. does the casting feel too hacky/fragile

packages/ui-components/src/components/MetricTimeseriesChart/MetricTimeseriesChart.tsx
```diff
@@ -0,0 +1,32 @@
...
+    />
+  );
```
pnavarrc: Maybe rename to `MetricLineChart`? All the metric charts will be timeseries charts (but maybe using colored lines, or bars)

packages/ui-components/src/components/MetricTimeseriesChart/interfaces.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Region } from "@actnowcoalition/regions";
+
+interface BaseChartProps {
+  width: number;
+  height: number;
+  marginTop?: number;
+  marginBottom?: number;
+  marginLeft?: number;
+  marginRight?: number;
+}
```
pnavarrc: This base interface is great, we can reuse that for a bunch of high level charts.

pnavarrc (approved):
Looks good!

faihegberg (approved):

packages/ui-components/src/components/MetricTimeseriesChart/MetricTimeseriesChart.tsx
mikelehen: Thanks for the ping!

I would probably change 

```
const { timeseries } = data;
```
to:
```
const timeseries = data.timeseries.assertFiniteNumbers();
```

above, and that should fix it.

Ideally we could just do `timeseries.assertFiniteNumbers()` here, but we end up with `Timeseries<number>` instead of `NonEmptyTimeseries<number>` so we have to do it before the `hasData()` check.  A bit clunky.  Maybe we can find a way to make it smoother.

PR #240, opened 2022-09-28
Update map components to have configurable region URLs

### LINKED ISSUE #220: Update the map components to use app regions

pnavarrc:
Close #220 

Updated the base maps to take an optional `getRegionUrl` function, and the metric-aware maps to use the `getRegionUrl` from the `RegionDB`. Updated the stories accordingly and simplified them a bit.

### DIFF
--- a/packages/ui-components/src/components/Maps/USNationalMap/MetricUSNationalMap.tsx
+++ b/packages/ui-components/src/components/Maps/USNationalMap/MetricUSNationalMap.tsx
@@ -21,6 +21,11 @@ export const MetricUSNationalMap: React.FC<MetricUSNationalMapProps> = ({
         const region = regionDB.findByRegionId(regionId);
         return region ? data.metricData(region, metric).getColor() : "#eee";
       }}
+      getRegionUrl={(regionId: string) => {
+        const region = regionDB.findByRegionId(regionId);
+        const url = region ? regionDB.getRegionUrl(region) : undefined;
+        return url;
+      }}
       showCounties={showCounties}
       {...otherProps}
     />
--- a/packages/ui-components/src/components/Maps/USNationalMap/StatesMap.tsx
+++ b/packages/ui-components/src/components/Maps/USNationalMap/StatesMap.tsx
@@ -19,7 +19,7 @@ const StatesMap: React.FC<{
   renderTooltip,
   showCounties,
   getFillColor,
-  getRegionUrl,
+  getRegionUrl = () => undefined,
 }) => {
   return (
     <svg width={width} height={height}>
@@ -27,7 +27,7 @@ const StatesMap: React.FC<{
         const stateFips = `${geo.id}`;
         return (
           <Tooltip key={stateFips} title={renderTooltip(stateFips)}>
-            <a href={getRegionUrl && getRegionUrl(stateFips)}>
+            <a href={getRegionUrl(stateFips)}>
               <g>
                 {!showCounties && (
                   <RegionShapeBase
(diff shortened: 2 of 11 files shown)


packages/ui-components/src/components/Maps/USNationalMap/StatesMap.tsx
```diff
@@ -11,30 +11,33 @@ const StatesMap: React.FC<{
...
-            <g>
-              {!showCounties && (
-                <RegionShapeBase
-                  d={geoPath(geo) ?? ""}
-                  fill={getFillColor(stateFips)}
-                />
-              )}
-              <RegionOverlay d={geoPath(geo) ?? ""} />
-            </g>
+            <a href={getRegionUrl && getRegionUrl(stateFips)}>
```
mikelehen: Same here, you could do `getRegionUrl?.(stateFips)` if you want.

packages/ui-components/src/components/Maps/USNationalMap/MetricUSNationalMap.tsx
```diff
@@ -21,6 +21,11 @@ export const MetricUSNationalMap: React.FC<MetricUSNationalMapProps> = ({
         const region = regionDB.findByRegionId(regionId);
         return region ? data.metricData(region, metric).getColor() : "#eee";
       }}
+      getRegionUrl={(regionId: string) => {
+        const region = regionDB.findByRegionId(regionId);
+        const url = region ? regionDB.getRegionUrl(region) : undefined;
```
mikelehen, line 26: FWIW- If you wanted I think you could actually do `region?.getRegionUrl?.(region)`.  I go back/forth on whether I actually like that syntax though, so 100% up to you.

Another thought: We could make getRegionUrl() accept `string | Region` and do the lookup internally.

packages/ui-components/src/components/Maps/USStateMap/USStateMap.tsx
```diff
@@ -25,6 +25,7 @@ const USStateMapInner: React.FC<USStateMapProps> = ({
   currentRegion,
   showCounties = true,
   showBorderingStates = true,
+  getRegionUrl,
```
mikelehen: Consider `getRegionUrl = () => undefined` or something so we don't need need to do `getRegionUrl && ...` below?

packages/ui-components/src/components/Maps/USNationalMap/StatesMap.tsx
```diff
@@ -11,30 +11,33 @@ const StatesMap: React.FC<{
...
   getFillColor: (regionId: string) => string;
+  getRegionUrl?: (regionId: string) => string | undefined;
 }> = ({
   width,
   height,
   geoPath,
   renderTooltip,
   showCounties,
   getFillColor,
+  getRegionUrl,
```
mikelehen: Consider `getRegionUrl = () => undefined` or something so we don't need need to do `getRegionUrl && ...` below?

mikelehen (approved):
Some minor optional nits, but looks good!

PR #243, opened 2022-09-29
Remove nils from timeseries fetched from CAN or CSV.

mikelehen:
I ran into an issue with CAN data breaking the sparklines because it had nulls in it.  We could make the sparklines remove / deal with the null data, but I think we probably don't want it in the first place.  So I'm just removing it up front in the data provider.

I actually think we should maybe change `Timeseries` so that it doesn't allow `null`s in the Timeseries at all.  I added a topic to standup for tomorrow.

### DIFF
--- a/.changeset/funny-apricots-think.md
+++ b/.changeset/funny-apricots-think.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/metrics": patch
+---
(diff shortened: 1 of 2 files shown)


pnavarrc (approved):
Sweet! 💯

PR #244, opened 2022-09-29
Implement select component

### LINKED ISSUE #219: Style MUI Select
The [Trends](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=100%3A5329) chart and the [MiniMap](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=118%3A4529) use a dropdown control that looks like the following:

Image

We need to style the Select component form MUI so it looks like the mocks. Note that the mocks don't have an example of the open/hovered states, so let's use MUI defaults for now

- https://mui.com/material-ui/react-select/
- [Figma - ActNow Design System → Input Selector](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=41%3A1555)

### Examples

Image

Image

faihegberg:
Closes https://github.com/covid-projections/act-now-packages/issues/219

### DIFF
--- a/packages/ui-components/src/stories/Select.stories.tsx
+++ b/packages/ui-components/src/stories/Select.stories.tsx
@@ -30,7 +30,7 @@ const sampleOptions = [
 const longOptions = sampleOptions.concat({
   value: "option 4",
   label:
-    "Option 4. I am a very long option. I may extend pass the width of the component. Some might even mysteriously decide to make me a long paragraph...",
+    "Option 4. I am a very long option. I may extend past the width of the component. Some might even mysteriously decide to make me a long paragraph...",
 });
 
 const defaultProps: TextFieldProps = {
--- a/packages/ui-components/src/stories/Select.stories.tsx
+++ b/packages/ui-components/src/stories/Select.stories.tsx
@@ -1,12 +1,18 @@
 import React, { useState } from "react";
-import { Box, MenuItem, TextField, Typography } from "@mui/material";
+import {
+  Box,
+  MenuItem,
+  TextField,
+  TextFieldProps,
+  Typography,
+} from "@mui/material";
 import KeyboardArrowDownIcon from "@mui/icons-material/KeyboardArrowDown";
 
 export default {
   title: "Design System/Select",
 };
 
-const options = [
+const sampleOptions = [
   {
     value: "option 1",
     label: "Option 1",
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/stories/Select.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  return (
+    <Box component="form" width={320}>
+      <TextField
+        select
+        label={<Typography fontSize={14}>Select</Typography>}
+        value={option}
+        onChange={handleChange}
+        variant="filled"
```
faihegberg: Using "filled" variant as its label positioning matches our mocks the most. Using other variants may call for hacky css styling.

packages/ui-components/src/stories/Select.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const [option, setOption] = useState("option 1");
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setOption(event.target.value);
+  };
+
+  return (
+    <Box component="form" width={320}>
+      <TextField
+        select
```
faihegberg: Using `Textfield` component where we can specify it as a `Select` component instead of pure `Select` component. This allows for more flexibility in terms of styling ([more info](https://mui.com/material-ui/react-text-field/#select)).

pnavarrc:
## Feedback

The "Select" label should be `paragraphSmall` according to the mocks https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=41%3A1555

image

The border should take the primary color when focused

image

packages/ui-components/src/styles/theme/components.ts
```diff
@@ -4,6 +4,24 @@ import { ThemeOptions, createTheme } from "@mui/material";
 const referenceTheme = createTheme();
 
 const components: ThemeOptions["components"] = {
+  MuiTextField: {
+    styleOverrides: {
+      root: ({ theme }) => ({
+        backgroundColor: theme.palette.common.white,
+        margin: `${theme.spacing(1)} ${theme.spacing(2)}`,
```
pnavarrc: I _think_ you can do `margin: theme.spacing(1, 2)` if you want

packages/ui-components/src/styles/theme/components.ts
```diff
@@ -4,6 +4,24 @@ import { ThemeOptions, createTheme } from "@mui/material";
 const referenceTheme = createTheme();
 
 const components: ThemeOptions["components"] = {
+  MuiTextField: {
+    styleOverrides: {
```
pnavarrc, line 15: You can customize the `defaultProps` to have the correct icon and options by default:

```tsx
  MuiTextField: {
    defaultProps: {
      SelectProps: {
        disableUnderline: true,
        IconComponent: KeyboardArrowDown,
      },
    },
    styleOverrides: {
      // ...
    }
// ...
```

packages/ui-components/src/stories/Select.stories.tsx
pnavarrc: Maybe make it a default prop (see below)

packages/ui-components/src/stories/Select.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+  {
+    value: "option 3",
+    label: "Option 3",
+  },
+];
+
+const longOptions = sampleOptions.concat({
+  value: "option 4",
+  label:
+    "Option 4. I am a very long option. I may extend pass the width of the component. Some might even mysteriously decide to make me a long paragraph...",
```
mikelehen: nit: "pass" => "past"

packages/ui-components/src/stories/Select.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+};
+
+export const Select = () => {
+  const [option, setOption] = useState("option 1");
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setOption(event.target.value);
+  };
+  return (
+    <Box component="form" width={320}>
+      <TextField {...defaultProps} value={option} onChange={handleChange}>
```
mikelehen, line 50: Out of curiosity (no need to change it), is there a reason we're using `<TextField select ...>` instead of `<Select ...>` ?  I'm guessing there is, but it was confusing to me at first glance.  I didn't know you could make a TextField turn into a Select.

mikelehen (approved):
Looks good!

packages/ui-components/src/stories/Select.stories.tsx
faihegberg: Yes, this is intentional because the textfield component gives us more flexibility in terms of styling. If one just uses the select component, styling the label would much trickier (for example, there isn't a default option for positioning the label in a way that matches our mocks). In contrast, the textfield component offers an option to create a select component that matches our mocks very closely ([more info](https://mui.com/material-ui/react-text-field/#select)).

mikelehen: Cool, thanks for explaining!

PR #245, opened 2022-09-29
Add tooltip and marker dot to `MetricLineChart`

pnavarrc:
This PR implements the `MetricTooltip` component which renders a tooltip for a given `metric`, `region` and `point` with the date and value to show in the tooltip. The position of the tooltip is controlled by the children of the tooltip, and the orientation of the tooltip adapts when necessary.

**Storybook**

 - [MetricLineChart](https://act-now-packages--pr245-pablo-chart-tooltip-g3t5mrhe.web.app/storybook/index.html?path=/story/charts-metriclinechart--new-york-mock-cases)
- [MetricTooltip](https://act-now-packages--pr245-pablo-chart-tooltip-g3t5mrhe.web.app/storybook/index.html?path=/story/components-metrictooltip--example)

image

I had to change the implementation of `MetricLineChart` (but not the interface) to contain most of the logic that used to be in `TimeseriesLineChart` (I left that component untouched). I discussed this briefly with @chasulin and we decided to punt on it, but it was actually necessary to add the tooltip functionality.

### DIFF
--- a/packages/ui-components/src/components/MetricTooltip/MetricTooltip.stories.tsx
+++ b/packages/ui-components/src/components/MetricTooltip/MetricTooltip.stories.tsx
@@ -1,0 +1,41 @@
+import React from "react";
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { states } from "@actnowcoalition/regions";
+import { metricCatalog, MetricId } from "../../stories/mockMetricCatalog";
+import { MetricTooltip, MetricTooltipContent } from ".";
+
+export default {
+  title: "Components/MetricTooltip",
+  component: MetricTooltip,
+} as ComponentMeta<typeof MetricTooltip>;
+
+const date = new Date("2022-03-15");
+const value = 12.54;
+const metric = metricCatalog.getMetric(MetricId.MOCK_CASES);
+const region = states.findByRegionIdStrict("53");
+
+const Template: ComponentStory<typeof MetricTooltip> = (args) => (
+  <svg width={600} height={400} style={{ border: "solid 1px #ddd" }}>
+    <MetricTooltip {...args}>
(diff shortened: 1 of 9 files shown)


packages/ui-components/src/components/MetricTooltip/MetricTooltip.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+  <svg
+    width={width}
+    height={height}
+    style={{ backgroundColor: colors.blue[50] }}
+  >
+    <MetricTooltip {...args}>
+      <Group left={width / 2} top={height / 2}>
+        <circle r={60} fill={colors.purple[900]} />
+        <text fill="white" textAnchor="middle" dominantBaseline="middle">
+          Hover me
```
mikelehen, line 33: Hovering doesn't work for me. Is it supposed to?

mikelehen (approved):
Nice! LGTM except not sure the MetricTooltip storybook is working, and there's a bug with the chart where if your mouse is located inside where the tooltip would display then it flickers and disappears.  We could punt on that if it's not an easy fix, but please log a github issue then.

packages/ui-components/src/components/MetricTooltip/MetricTooltip.stories.tsx
pnavarrc: it absolutely should, debugging!

pnavarrc: So using `Group` doesn't work because the children element needs to be able to hold a ref, so I changed it in the example to use the native `g` instead.

In the case of the chart, we needed to disable the hover listener of the tooltip itself, since we don't want the tooltip to do anything when we hover it (and we control its open state directly)

PR #250, opened 2022-09-29
Add parallelization and retries to CovidActNowDataProvider.

mikelehen:


packages/metrics/src/data_providers/CovidActNowDataProvider.ts
```diff
@@ -40,22 +42,41 @@ export class CovidActNowDataProvider extends CachingMetricDataProviderBase {
...
+        chunk.map(async (region) => {
+          const cacheKey = `${region.regionId}-${includeTimeseries}`;
+          // If timeseries data exists in the cache, skip the fetch no matter what,
+          // as the timeseries endpoints also contain all of the non-timeseries data,
+          // and we will use this data if necessary via the getCachedData method.
+          if (!this.getCachedData(region, includeTimeseries)) {
+            const url = this.buildFetchUrl(region, includeTimeseries);
+            for (let attempt = 0; attempt < 3; attempt++) {
+              try {
+                console.log("Fetching", url);
```
mikelehen, line 60: Intentionally keeping this for now since it's helpful to see when we're fetching data.  We should move this behind some sort of debug logging flag in the future.

packages/metrics/src/data_providers/CovidActNowDataProvider.ts
```diff
@@ -67,10 +88,12 @@ export class CovidActNowDataProvider extends CachingMetricDataProviderBase {
...
-    assert(
-      cachedData,
-      `No data found in cache for ${region.regionId} with includeTimeseries=${includeTimeseries}`
-    );
+    if (!cachedData) {
+      console.warn(
+        `No data found in cache for ${region.regionId} with includeTimeseries=${includeTimeseries}. Perhaps the fetch failed?`
+      );
+      return new MetricData(metric, region, /*currentValue=*/ null);
+    }
```
mikelehen, line 96: This change isn't related, but I ran into an issue fetching data for FIPS 69085.  It exists in the region DB, but there's no CAN endpoint for it, so the fetch would fail and we'd end up here with it not in the cache.  This may not be the best fix, but it works for now.

yarn.lock
```diff
@@ -3889,6 +3889,14 @@
...
     d3-shape "^1.0.6"
 
+"@visx/event@^2.6.0":
+  version "2.6.0"
+  resolved "https://registry.yarnpkg.com/@visx/event/-/event-2.6.0.tgz#0718eb1efabd5305cf659a153779c94ba4038996"
+  integrity sha512-WGp91g82s727g3NAnENF1ppC3ZAlvWg+Y+GG0WFg34NmmOZbvPI/PTOqTqZE3x6B8EUn8NJiMxRjxIMbi+IvRw==
+  dependencies:
+    "@types/react" "*"
+    "@visx/point" "2.6.0"
+
```
mikelehen, line 3899: This is unrelated to my changes. Guessing it got left out of somebody else's PR.

pnavarrc (approved):
Looks good!

PR #252, opened 2022-09-29
Update URL routing for RegionSearch

smcclure17:
Based on changes in https://github.com/covid-projections/act-now-packages/pull/240

RegionSearch component is not navigating to location pages due to the routing not being updated yet.

Example use: https://github.com/covid-projections/hackathon-august-2022/pull/13/files


pnavarrc (approved):
Looks good!

PR #258, opened 2022-10-01
Rework hooks to fix bug found in hackathon.

mikelehen:
This fixes a bug we found in the September hackathon where switching tabs on the chart group would cause a crash when we went to fetch the data for the new tab.

Basically, if your component called `useData(region, metric1)` the first time it rendered and `useData(region, metric2)` the second time, it would cause a crash.

The core problem is that the useData hooks (and any hook that uses useEffect) update asynchronously.  When you pass a different argument to the hook (e.g. changing the selected metric), the hook will continue to return the "old" result until the asynchronous operations complete.

That wouldn't be a big problem (in fact it's kind of convenient, since it avoids the chart disappearing momentarily while the hook populates), but the way I implemented each hook on top of the others ended up problematic.  In particular:

```
export function useData(
  region: Region,
  metric: string | Metric,
  includeTimeseries = false
): DataOrError<MetricData> {
  const { data: dataStore, error } = useDataForMetrics(
    region,
    [metric],
    includeTimeseries
  );
>>>  const data = dataStore?.metricData(metric); <<<
  return { data, error };
}
```

This code assumes that `useDataForMetrics()` either returns no data or it returns data that contains the metric we asked for, but when you switch to a new metric, the `useDataForMetrics()` call is going to temporarily return the data for the old metric, and so trying to call `.metricData(metric)` is going to fail.

I fixed this by making it so that each hook is implemented on its own with its own state variables and doesn't make any assumptions about what metric we're returning data for.  This ends up with some repetitive code, but was the cleanest solution I came up with.


packages/ui-components/src/common/hooks/metric-data.ts
```diff
@@ -32,13 +32,13 @@ export function useData(
...
-  );
-  const data = dataStore?.metricData(metric);
-  return { data, error };
+  const catalog = useMetricCatalog();
+
+  const fetchData = useCallback(() => {
+    return catalog.fetchData(region, metric, includeTimeseries);
+  }, [catalog, region, metric, includeTimeseries]);
+
+  return useFetchedData(fetchData);
```
mikelehen, line 40: I created this `useFetchedData()` helper to try to share code between the hooks. It helps, but the other 2 still end up kinda' ugly / repetitive because of the `useCachedArrayIfEqual` stuff we have to do.

In general I'm not very happy with the way these hooks have ended up, but this is the cleanest solution I came up with for now.

packages/metrics/src/data/MultiMetricDataStore.ts
```diff
@@ -57,8 +57,9 @@ export class MultiMetricDataStore<T = unknown> {
    * @param metric The metric to check for data for.
    * @returns Whether this data store contains data for the given metric.
    */
-  hasMetricData(metric: Metric): boolean {
-    return this.data[metric.id] !== undefined;
+  hasMetricData(metric: Metric | string): boolean {
+    const metricId = metric instanceof Metric ? metric.id : metric;
+    return this.data[metricId] !== undefined;
```
mikelehen, line 62: This was an oversight where hasMetricData() didn't let you pass in a string metric.  I ran into it while updating tests, so I updated it.

packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
```diff
@@ -132,7 +132,7 @@ export class MultiRegionMultiMetricDataStore<T = unknown> {
    * @param metric The metric to check for data for.
    * @returns Whether the data store contains data for the given region and metric.
    */
-  hasMetricData(region: Region, metric: Metric): boolean {
+  hasMetricData(region: Region, metric: Metric | string): boolean {
```
mikelehen, line 135: same

packages/ui-components/src/common/hooks/metric-data.test.tsx
```diff
@@ -0,0 +1,207 @@
+import { renderHook } from "@testing-library/react-hooks";
```
mikelehen, line 1: I reworked the tests to wrap the hook in a `<MetricCatalogProvider>` which caused me to have to convert them to .tsx.

But I also refactored them and added a test for each of `useDataForRegionsAndMetrics()`, `useDataForMetrics()`, and `useData()`, so the file changed a lot and github doesn't recognize it as a rename anymore.  Sorry!

packages/ui-components/src/common/hooks/metric-data.ts
```diff
@@ -69,44 +77,49 @@ export function useDataForMetrics(
  * @param regions The regions to get data for.
  * @param metrics The metrics to get data for as either strings or `Metric` objects.
  * @param includeTimeseries Whether to fetch timeseries data or not (default false).
- * @param optCatalog Optional metric catalog to use. If not provided, the
```
mikelehen, line 72: Note that this used to accept `optCatalog` for the sake of testing, but I figured out how to wrap the hook in the tests in a `MetricCatalogProvider` so it's no longer necessary.

pnavarrc (approved):
Looks good!

PR #260, opened 2022-10-03
Fix JSON fetching when not all fields are present for all rows

smcclure17:
Fixes error when fetching vaccine data from the CAN API, which fails because not all timeseries records contain the vaccine data fields. 

```Error fetching metric data: Error: INTERNAL ASSERTION FAILED: Metric key vaccinationsInitiatedRatio missing```.


mikelehen (approved):
Looks good! Thanks for jumping on this.

Issue #259, opened 2022-10-03
Try to unify levelSet and categories for COMMUNITY_RISK_LEVEL metric.


mikelehen:
See https://github.com/covid-projections/hackathon-september-2022/pull/3/files#diff-ce3a93efd046e7a95d61484997047a9987f83e8947ca7abfdd6f3669cfed455eR97

Basically to set up the COMMUNITY_RISK_LEVEL metric in the hackathon, I had to specify a levelSet so that we could render a thermometer for it, and also categories (so that the 0,1,2 values could be rendered as Low,Medium,High and colored correctly).

This feels kinda' redundant / hacky.  I'm wondering if we could make the thermometer fall back to rendering from the categories or something.  Or maybe categories should be tied to levels somehow (you specify a value and a level instead of value,label,color)?  Not sure.

pnavarrc:
I have been thinking about this recently, I think we could consolidate the notion of level for metrics with categories and thresholds

- Levels have a color, a name and a description or extended name
- There is a notion of order in the levels. For example: _high risk_ to _low risk_, or _high_, _medium high_, _medium low_, and _low income_ for categorical, and levels sorted by thresholds in the case of threshold metrics
- In both cases we have a "default" or "missing data" level that will likely be treated differently in the UI (it shouldn't be part of the thermometer, for example)
- Both types of metrics should have a way to resolve a level given metric data

We could update `MetricCategory` to be defined in terms of `MetricLevel` objects, and separate the default level from the ordered sequence of levels for the metric.

mikelehen:
👍 to all of that. I think it's worth trying and we can see how it looks!

PR #263, opened 2022-10-03
Update `MetricLegendThreshold` to have thresholds between levels

### LINKED ISSUE #254: The `MetricThresholdLegend` component should show thresholds instead of labels
At least in the horizontal version, the labels of the ticks between the levels should be the threshold values, not the level name.

Example from https://hackathon-september-2022-4dve.vercel.app/

Image

[ActNow Design System - Legend](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=13%3A948)

pnavarrc:
Close #254  - MetricLegendThreshold component should show thresholds between labels

image

I added a utility function to format the items that we pass to the `LegendThreshold` component so we have all we need to show thresholds, level name, color and description. It works as expected for both horizontal and vertical orientations.

### DIFF
--- a/packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
+++ b/packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
@@ -1,13 +1,25 @@
 import React from "react";
-import { LegendThreshold } from "../LegendThreshold";
-import { Metric, MetricLevel } from "@actnowcoalition/metrics";
-import { useMetricCatalog } from "../MetricCatalogContext";
-import { assert } from "@actnowcoalition/assert";
 import { Stack, Typography } from "@mui/material";
+import { Metric } from "@actnowcoalition/metrics";
+import { assert } from "@actnowcoalition/assert";
+import { LegendThreshold } from "../LegendThreshold";
+import { useMetricCatalog } from "../MetricCatalogContext";
 
-const getItemColor = (item: MetricLevel) => item.color;
-const getItemLabel = (item: MetricLevel) => item.name ?? item.id;
-const getItemSublabel = (item: MetricLevel) => item.description ?? "";
+interface LevelItem {
+  /** Level name (e.g. "High") */
+  name: string;
+  /** Level color */
+  color: string;
+  /** Description of the level */
+  description: string | undefined;
+  /** Formatted value of the threshold at the end of the current level */
+  endThreshold: string;
+}
+
+const getItemColor = (item: LevelItem) => item.color;
+const getItemLabelHorizontal = (item: LevelItem) => item.endThreshold;
+const getItemLabelVertical = (item: LevelItem) => item.name;
+const getItemSublabel = (item: LevelItem) => item.description ?? "";
 
 interface CommonMetricLegendThresholdProps {
   /** Metric to display thresholds for. */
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
```diff
@@ -109,13 +114,37 @@ export const MetricLegendThreshold: React.FC<MetricLegendThresholdProps> = ({
...
+  assert(
+    metricLevels,
+    "Metric must have levels and thresholds to use MetricLegendThreshold." +
+      `No levels found for metric ${metric}.`
+  );
+  assert(
+    metricThresholds,
+    "Metric must have levels and thresholds to use MetricLegendThreshold." +
+      `No thresholds found for metric ${metric}.`
+  );
```
mikelehen: I think this is going to break the "Community Risk Level" MetricScoreOverview on the hackathon location pages based on the way I [hooked it up](https://github.com/covid-projections/hackathon-september-2022/pull/3) this morning:

image

The metric uses `categories` instead of `thresholds`, but I did also gave it a `levelSetId` just so that the `MetricLegendThreshold` would work. This felt a bit hacky so I opened https://github.com/covid-projections/act-now-packages/issues/259

But for now the metric has `levels` but not `thresholds` and so your change will break it.  Perhaps make `endThreshold` optional and omit it if the metric doesn't have `thresholds` or something?

packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
```diff
@@ -92,9 +97,9 @@ export const MetricLegendThreshold: React.FC<MetricLegendThresholdProps> = ({
         </Stack>
         <Stack direction="row" spacing={1} alignItems="center">
           {startLabel && startLabel}
```
mikelehen: Unrelated to your change, but can't we just do `{startLabel}`?  

And there are 3 more of these below to update if you change this one.

packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
```diff
@@ -68,22 +80,15 @@ export const MetricLegendThreshold: React.FC<MetricLegendThresholdProps> = ({
...
-  const derivedProps = isHorizontal
-    ? { items, getItemColor, getItemLabel }
-    : { items, getItemColor, getItemLabel, getItemSublabel };
+
+  const items = getLevelItems(metric);
+  const commonProps = { items, getItemColor, ...legendThresholdProps };
 
   if (!includeOverview) {
-    return <LegendThreshold {...derivedProps} {...legendThresholdProps} />;
+    return <LegendThreshold {...commonProps} {...legendThresholdProps} />;
```
mikelehen: `legendThresholdProps` is included in `commonProps` right?  Maybe I'm missing something.

packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
```diff
@@ -68,22 +80,15 @@ export const MetricLegendThreshold: React.FC<MetricLegendThresholdProps> = ({
...
-    "Metric must have thresholds to use MetricLegendThreshold." +
-      `No thresholds found for metric ${metric}.`
-  );
-  const isHorizontal = legendThresholdProps.orientation === "horizontal";
-  const derivedProps = isHorizontal
-    ? { items, getItemColor, getItemLabel }
-    : { items, getItemColor, getItemLabel, getItemSublabel };
+
+  const items = getLevelItems(metric);
+  const commonProps = { items, getItemColor, ...legendThresholdProps };
```
mikelehen, line 87: Perhaps add a comment, "// common props regardless of horizontal / vertical orientation" or something?

mikelehen (approved):
LGTM with a couple nits and one issue with requiring metrics to have thresholds.

packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
pnavarrc: Good point, maybe a good solution for now is to only assert the thresholds when we need them (when `orientation = 'horizontal'` and `showLabels = true`)

PR #265, opened 2022-10-03
`AxesTimeseries` now supports custom axis options

### LINKED ISSUE #262: Metric chart axes should use Metric.formatValue()
In the hackathon, the vaccination charts have axes labels like 0.1 instead of 10%, etc.

Image

pnavarrc:
Close https://github.com/covid-projections/act-now-packages/issues/262 - Metric chart axes should use `Metric.formatValue`

The component now can take `axisLeftProps` and `axisBottomProps` so the parent component can further customize the axes. This component might be too simple to keep, but updating as is for now to have the y-axis for `MetricLineChart` using the same formatter function as the metric.

image

### DIFF
--- a/.changeset/pink-mugs-notice.md
+++ b/.changeset/pink-mugs-notice.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/ui-components": minor
+---
+
+AxisTimeseries now allows custom options for the axis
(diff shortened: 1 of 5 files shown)


mikelehen (approved):
Nice, looks good!

PR #266, opened 2022-10-04
Share button improvements

### LINKED ISSUE #256: Improve `ShareButton` component
- Make `onShareX` and `onCopyLink` handlers optional
- The `ShareButton` should take the same options as `Button` and `ButtonGroup` (`size` in particular)

chasulin:
Fixes #256 

Updates:
- Fixes facebook+twitter share buttons. They previously were each a button nested within another button (a MUI button nested inside of a ReactShare button). This PR adds component="div" to the MUI button, so the dom doesn't register it as a second button
- Reverses the nesting of the ReactShare buttons (puts MUI button inside ReactShare button) so that the whole button is clickable
- Adds MUI `ButtonProps` to the top level `ShareButtonProps` and applies them to the anchor share button (so, for example, we could make the share button large or small using the native MUI `size` prop)
- Makes additional onClick props (onCopyLink, onShareTwitter, onShareFacebook) all optional and default to an empty function, so we don't manually need to pass empty functions if those props are unused

### DIFF
--- a/packages/ui-components/src/components/ShareButton/FacebookShareButton.tsx
+++ b/packages/ui-components/src/components/ShareButton/FacebookShareButton.tsx
@@ -1,8 +1,7 @@
 import React from "react";
 import { FacebookShareButton as ReactShareFacebookShareButton } from "react-share";
 import FacebookIcon from "@mui/icons-material/Facebook";
-import { StyledReactShareFacebookButton } from "./ShareButton.style";
-import { Typography, Stack } from "@mui/material";
+import { Button } from "@mui/material";
 
 type BaseProps = React.ComponentProps<typeof ReactShareFacebookShareButton>;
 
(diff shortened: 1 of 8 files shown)


packages/ui-components/src/components/ShareButton/FacebookShareButton.tsx
```diff
@@ -13,7 +14,16 @@ export const FacebookShareButton: React.FC<FacebookShareButtonProps> = ({
...
+    {...otherProps}
+    onClick={onClick}
+    resetButtonStyle={false} /** allows us to customize styles */
+  >
+    <Stack spacing={0.75} direction="row" justifyContent="center">
+      <Typography variant="labelLarge" color="primary">
+        Facebook
+      </Typography>
+      <FacebookIcon color="primary" />
+    </Stack>
```
pnavarrc: How about using a styled Button with `component="div"` - that way we can have a similar layout and typography settings without it being an actual `<button>` element

```tsx
<StyledFacebookShareButton {/* ... */}>
  <Button component="div" endIcon={<FacebookIcon />} >
    Facebook
  </Button>
</StyledFacebookShareButton>
```

we could also potentially propagate the button props from the main button to the children (size, etc) if we wanted to do that

pnavarrc (approved):
Added a few suggestions, but looks good!

PR #267, opened 2022-10-04
Remove supporting text in favor of the extended metric name.

mikelehen:


packages/ui-components/src/components/MetricOverview/MetricOverview.tsx
```diff
@@ -38,9 +35,7 @@ export const MetricOverview: React.FC<MetricOverviewProps> = ({
         <LabelIcon>{metric.name}</LabelIcon>
         {metricChart}
         <MetricValue region={region} metric={metric} />
-        {supportingText && (
-          <Typography variant="paragraphSmall">{supportingText}</Typography>
-        )}
+        <Typography variant="paragraphSmall">{metric.extendedName}</Typography>
```
mikelehen, line 38: I think the `supportingText &&` guard is unnecessary since we just end up with an empty `<span>` which is fine, so I removed it for simplicity (and consistency with other places where we don't do it).  I can add it back if we think it's important though.

pnavarrc (approved):
Looks good!

PR #274, opened 2022-10-05
Updates to metric compare table (extended names, population)

mikelehen:
* Add metric extended names to metric columns.
* Add population to location column.

To be honest, I am not sure it looks great (especially with the long metric names we have in storybook), but I think this roughly matches the mocks:

image

### DIFF
--- a/packages/ui-components/src/common/utils/index.ts
+++ b/packages/ui-components/src/common/utils/index.ts
@@ -1,5 +1,5 @@
 export { default as importJson } from "./importJson";
-import { formatDecimal } from "@actnowcoalition/number-format";
+import { formatInteger } from "@actnowcoalition/number-format";
 
 /**
  * Format the population with thousands separator and keeping 3 significant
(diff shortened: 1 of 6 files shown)


packages/ui-components/src/common/utils/index.ts
```diff
@@ -1 +1,18 @@
...
+ * ```ts
+ * formatPopulation(107766) // 108,000
+ * ```
+ */
+export function formatPopulation(population: number) {
+  return formatDecimal(population, {
+    maximumSignificantDigits: 3,
+    maximumFractionDigits: 0,
+  });
+}
```
mikelehen, line 17: I moved `formatPopulation()` from `RegionSearch` to here.  I'm not sure I'm wild about putting it in the `index.ts` file though.  Maybe I should create a `misc.ts` file or something?

I could also move this into the number-format package, but I assume that's for general-purpose number-formatting, not specific kinds of numbers (like populations).

pnavarrc: Maybe we could have a second `formatInteger` function in `number-format` that takes options, so we can use currency symbols, thousands separator, etc

pnavarrc (approved):
Looks good!

packages/ui-components/src/common/utils/index.ts
mikelehen: I went ahead and did this, but I also kept `formatPopulation()` to ensure that RegionSearch and CompareTable use the same logic. 🤷

PR #279, opened 2022-10-05
Style select multiple component

faihegberg:
This PR styles the MUI autocomplete component to match the multi-select component in our [mocks](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=41%3A1555).

image

### DIFF
--- a/packages/ui-components/src/stories/SelectMultiple.stories.tsx
+++ b/packages/ui-components/src/stories/SelectMultiple.stories.tsx
@@ -1,0 +1,91 @@
+import React from "react";
+import { Box, TextField, TextFieldProps, Typography } from "@mui/material";
+import KeyboardArrowDownIcon from "@mui/icons-material/KeyboardArrowDown";
+import Chip from "@mui/material/Chip";
+import Autocomplete from "@mui/material/Autocomplete";
+
+export default {
+  title: "Design System/SelectMultiple",
+};
+
+const cities = [
+  { name: "Beijing", country: "China" },
+  { name: "Tokyo", country: "Japan" },
+  { name: "Kinshasa", country: "DR Congo" },
+  { name: "Moscow", country: "Russia" },
+  { name: "Jakarta", country: "Indonesia" },
+  { name: "Cairo", country: "Egypt" },
+  { name: "Seoul", country: "South Korea" },
+  { name: "Mexico City", country: "Mexico" },
+  { name: "London", country: "United Kingdom" },
+  { name: "Dhaka", country: "Bangladesh" },
+];
+
+const textFieldProps: TextFieldProps = {
+  variant: "filled",
+  fullWidth: true,
+  label: <Typography variant="paragraphSmall">Locations</Typography>,
+  InputLabelProps: { shrink: true },
+};
+
+export const ShortMultiSelect = () => {
+  return (
+    <Box width={320}>
+      <Autocomplete
+        multiple
+        popupIcon={<KeyboardArrowDownIcon />}
+        options={cities.map((option) => option.name)}
+        defaultValue={[cities[0].name]}
+        renderTags={(location, getTagProps) =>
+          location.map((locationName, index) => (
+            <Chip
+              {...getTagProps({ index })}
+              key={locationName}
+              variant="outlined"
+              label={locationName}
+            />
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/stories/SelectMultiple.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+  { name: "London", country: "United Kingdom" },
+  { name: "Dhaka", country: "Bangladesh" },
+];
+
+const textFieldProps: TextFieldProps = {
+  variant: "filled",
+  fullWidth: true,
+  label: <Typography variant="paragraphSmall">Locations</Typography>,
+  InputLabelProps: { shrink: true },
+};
```
faihegberg, line 34: Would be nice if we can incorporate some of these as default props in the style override file. Currently, incorporating these there would cause certain props to be ignored. I wonder whether it's to do with `params` being passed in `renderInput`. Any ideas to work around this?

packages/ui-components/src/stories/SelectMultiple.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+export const ShortMultiSelect = () => {
+  return (
+    <Box width={320}>
+      <Autocomplete
+        multiple
+        popupIcon={<KeyboardArrowDownIcon />}
```
faihegberg, line 41: Was also hoping to incorporate this icon as a default prop, but incorporating it there currently throws a typescript error since it's expecting a react node instead of svg.

packages/ui-components/src/stories/SelectMultiple.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+              variant="outlined"
+              label={locationName}
+            />
+          ))
+        }
+        renderInput={(params) => (
+          <TextField
+            {...params}
+            {...textFieldProps}
+            InputProps={{ ...params.InputProps, disableUnderline: true }}
```
faihegberg, line 58: Somewhat related to the note about `params` above, leaving `...params.InputProps` out and only including `disableUnderline: true` will cause the component to no longer function as a select dropdown.

chasulin: yeah you want to keep `...params.InputProps` in there, otherwise the InputProps obj gets emptied and then only contains `disableUnderline: true`

packages/ui-components/src/styles/theme/components.ts
```diff
@@ -5,6 +5,21 @@ import KeyboardArrowDownIcon from "@mui/icons-material/KeyboardArrowDown";
 const referenceTheme = createTheme();
 
 const components: ThemeOptions["components"] = {
+  MuiAutocomplete: {
```
chasulin, line 8: region search is also a MuiAutocomplete component so be careful when styling at the theme level // make sure to sanity check region search 🔍

packages/ui-components/src/stories/SelectMultiple.stories.tsx
mikelehen: To be clear, you tried specifying these as `defaultProps` on `MuiTextField`?  Or on `MuiAutocomplete`?  It looks like they are `TextField` props, so they'd need to be on `MuiTextField` (but we'd need to be careful to make sure we want them applied to all text fields in our apps).

packages/ui-components/src/stories/SelectMultiple.stories.tsx
mikelehen: I think you can, but you'd need to rename `components.ts` to `components.tsx` and do:

```
      popupIcon: (<KeyboardArrowDownIcon />),
```

packages/ui-components/src/stories/SelectMultiple.stories.tsx
faihegberg: Oh right 😅 I was trying to incorporate them under `MuiAutocomplete`. On second thought, I'm not going to incorporate them as style overrides for `MuiTextField` though. The styling is pretty specific to this component and we wouldn't want them spilling over to other text field components in our app.

PR #285, opened 2022-10-06
Clip the map projection to avoid rendering outside the viewBox

pnavarrc:
Adding a small improvement to the state maps. The map will normally project render the entire shape being passed, which can be wasteful if part of the map falls outside the viewBox.

In the state map case, we show parts of neighbouring states, but the map was actually rendering the entire state outside the visible area. Adding the `.clipExtent([[0, 0], [width, heigh]])` option "cuts" the projection using the provided dimensions, which avoids unnecessary rendering.

This change also fixes another bug - the tooltip for the state is centered around the center of the whole state, not around the visible area. Now that the SVG shape is restricted to the visible section, the tooltip is correctly positioned.

**Before**

image

Note: ☝️  This tooltip is from the "Before" image

**After**

image

👌  Now the tooltip is positioned around the part of Ohio that's visible on the screen.

### DIFF
--- a/.changeset/olive-buttons-grin.md
+++ b/.changeset/olive-buttons-grin.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/ui-components": patch
+---
+
+Clip the map extent to fix the tooltip positioning (and more efficient rendering)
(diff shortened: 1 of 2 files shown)


mikelehen (approved):
Looks good! Thanks for the very clear PR description!

PR #286, opened 2022-10-06
Implement ChartOverlayXY component

### LINKED ISSUE #287: Implement ChartOverlayXY

pnavarrc:
Close https://github.com/covid-projections/act-now-packages/issues/287

Complex charts such as the Trends chart might have multiple time series, each one with multiple points. Normally we could attach a tooltip to each point in the series, but it will be difficult for the user to hover exactly the point they are interested in (the polygons are visible here, but normally they wouldn't be).

image

One technique to solve this problem is to generate a Voronoi partition, which, given a set of points, generates a set of polygons; one for each point in the original set; such that all the points inside the polygon are closer to its corresponding point than to any other point in the set, so if we add hover events on the polygon instead of the point, the user will have a larger area to target.

image

The Trends chart has additional complexities that can also be solved with this component. For example, when hovering a point, we want to highlight the point and the line being hovered, as well as the location label. 

image

To achieve that, the `onMouseMove` handler receives the point being hovered, the index of the time series and the index of the point in the time series, which allows us to highlight the line and the point, add labels, date markers, tooltips, etc.

image

### DIFF
--- a/packages/ui-components/src/components/ChartOverlayXY/ChartOverlayXY.tsx
+++ b/packages/ui-components/src/components/ChartOverlayXY/ChartOverlayXY.tsx
@@ -41,6 +41,10 @@ export interface PointInfo {
   pointIndex: number;
 }
 
+const noop = () => {
+  // Do nothing
+};
+
 /**
  * This component is an overlay that helps capture the information of the
  * point that is being hovered by the user. Given a list of timeseries,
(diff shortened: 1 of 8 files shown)


packages/ui-components/src/components/ChartOverlayXY/ChartOverlayXY.tsx
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * See `ChartOverlayXY.stories.tsx` for usage examples.
+ */
+export const ChartOverlayXY = ({
+  width,
+  height,
+  xScale,
+  yScale,
+  timeseriesList,
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
```
chasulin: [this noop function](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/components/ShareButton/ShareButton.tsx#L10-L12) seems to not make typescript complain

chasulin (approved):
💫💫💫

PR #288, opened 2022-10-10
Legend threshold improvements 

### LINKED ISSUE #255: `LegendThreshold` component should have better ui defaults
_description in progress_

The `height`, `barHeight` and labels interact with each other in ways that make the default not look nice if `showLabels=false` - we want the component to always look nice

chasulin:
Fixes #255 

Having `barHeight` and `height` props was making it difficult to have sane default UI on all variations (with labels, without labels). This streamlines the horizontal legend's height by getting rid of the `barHeight` prop and just using `height`, adding an extra 20px of height to the horizontal legend when labels render (labels rendering below the horizontal legend shouldn't be longer than short numeric value, a percentage, etc. so 20px should suffice).

This PR also consolidates the props for `LegendThreshold` and `MetricLegendThreshold`. The props overlap enough that we don't need a `common` version, a `vertical` version, and a `horizontal` version for legend props, especially now that we've gotten rid of the `barHeight` prop, and all legends now just take an optional `height` and `width` (with sane defaults)

### DIFF
--- a/packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.stories.tsx
+++ b/packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.stories.tsx
@@ -25,7 +25,7 @@ export const HorizontalDefault = Template.bind({});
 HorizontalDefault.args = {
   orientation: "horizontal",
   width: horizontalWidth,
-  barHeight: horizontalBarHeight,
+  height: horizontalBarHeight,
   metric: MetricId.MOCK_CASES,
 };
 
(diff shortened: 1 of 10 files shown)


packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.stories.tsx
```diff
@@ -46,23 +40,23 @@ HorizontalOnlySideLabels.args = {
...
 };
 
-export const HorizontalSquared = Template.bind({});
-HorizontalSquared.args = {
+export const HorizontalSquaredNoLabels = Template.bind({});
+HorizontalSquaredNoLabels.args = {
   ...HorizontalDefault.args,
   borderRadius: 0,
-  showLabels: true,
+  showLabels: false,
```
chasulin, line 59: making these settings match the rounded variation's story above

smcclure17 (approved):
Do we need to add a changeset to this? But, looks great to me! Thanks for this cleanup

PR #290, opened 2022-10-11
Metric-aware MultiProgressBar component

chasulin:


packages/ui-components/src/components/MetricMultiProgressBar/MetricMultiProgressBar.tsx
```diff
@@ -0,0 +1,62 @@
+import React from "react";
+import { maxBy } from "lodash";
```
pnavarrc: remember to import only the function we need 

```tsx
import maxBy from "lodash/maxBy";
```

packages/ui-components/src/components/MetricMultiProgressBar/MetricMultiProgressBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+  /* Region for which to chart */
+  region: Region;
+  /* Metrics of which the progress bar will chart current values */
+  metrics: (Metric | string)[];
+  /* All props below - MultiProgressBar props to pass down */
+  width?: number;
+  height?: number;
+  bgColor?: string;
+  borderRadius?: number;
+}
```
pnavarrc, line 20: You can do

```tsx
export interface MetricMultiProgressBarProps extends MultiProgressBarProps {
  /* Region for which to chart */
  region: Region;
  /* Metrics of which the progress bar will chart current values */
  metrics: (Metric | string)[];
}
```

which should include the props from `MultiProgressBar` (and ensure that they are in sync, if we ever change the props for `MultiProgressBar`)

packages/ui-components/src/components/MetricMultiProgressBar/MetricMultiProgressBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+  color: string;
+  label: string;
+}
+
+export const MetricMultiProgressBar: React.FC<MetricMultiProgressBarProps> = ({
+  region,
+  metrics,
+  ...otherProgressBarProps
+}) => {
+  const { data } = useDataForRegionsAndMetrics([region], metrics, false);
```
pnavarrc: Maybe use `useDataForMetrics` instead?

packages/ui-components/src/components/MetricMultiProgressBar/MetricMultiProgressBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+        color: metricData.getColor(),
+        currentValue: metricData.currentValue as number, // improve this
+        label: metricData.formatValue(),
+      };
+    }
+  );
+
+  const maxValue =
+    maxBy(progressBarItems, (item: MetricProgressBarItem) => item.currentValue)
+      ?.currentValue ?? 1000; // fix this + how to treat min?
```
pnavarrc: I think we should pass the `maxValue` as a prop, as we do for `MultiProgressBar`. The max value for this type of chart is probably a reference value for the metrics themselves (a constant). For example, in the case of vaccination levels (1st and 2nd dose), the maximum value is 100%, not necessarily the max of both metrics.

For the minimum value, I think that starting from zero is reasonable (as it is for most bar charts). Starting from any other number would be misleading and will require adding axis or other annotations.

pnavarrc (commented):
Added a few notes, looks good in general, but we probably want to add patterns to the metrics so we can tell them apart from the main metric, even if they have the same color palette.

packages/ui-components/src/components/MetricMultiProgressBar/MetricMultiProgressBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const progressBarItems: MetricProgressBarItem[] = metrics.map(
+    (metric: Metric | string) => {
+      const metricData = data.metricData(region, metric);
+      return {
+        color: metricData.getColor(),
+        currentValue: metricData.currentValue as number, // improve this
+        label: metricData.formatValue(),
+      };
+    }
+  );
```
pnavarrc, line 48: Consider sorting the items descending in value to ensure that the smaller rectangles are always on top (the MultiProgressBar component does that already, so not really critical, but it could be if we add patterns/hatched lines in a specific order at this level).

pnavarrc (approved):
Approving, looks good minus a few recommendations.

packages/ui-components/src/components/MetricMultiProgressBar/MetricMultiProgressBar.tsx
chasulin: we don't want to include all `MultiProgressBarProps`, which is why i didn't do the above

chasulin: `MultiProgressBarProps` includes `items`, `getItemValue`, `getItemColor`, etc., but those aren't passed into this component-- they're declared in this component and then passed to `MultiProgressBar`

PR #291, opened 2022-10-11
Install and configure `eslint-plugin-lodash`

pnavarrc:
Importing individual lodash functions help keeping the bundle size small. This PR adds the eslint-plugin-lodash with some config to throw a warning when we attempt to use the second syntax - I updated the code in a few places to match the new rule.

```tsx
// this only imports the groupBy function 👍 
import groupBy from "lodash/groupBy";

// This imports all the lodash methods 👎 
import { groupBy } from "lodash";
```

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -18,10 +18,11 @@
     "ecmaVersion": "latest",
     "sourceType": "module"
   },
-  "plugins": ["react", "@typescript-eslint"],
+  "plugins": ["react", "@typescript-eslint", "lodash"],
   "rules": {
     "@typescript-eslint/no-explicit-any": "off",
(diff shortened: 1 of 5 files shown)


chasulin (commented):
confused, should i be seeing a warning when i type what's below? (im in your branch, ran yarn, etc.)

Screen Shot 2022-10-12 at 11 24 22 AM

pnavarrc:
It wasn't working for me either, I had to install the [ESLint VSCode Extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) and enable it, then the squiggly line appears

image

It does show the error during the pre-commit checks too

chasulin (approved):

PR #294, opened 2022-10-12
Implement `AutoWidth` component

pnavarrc:
Implements a component that measures the width of its container and passes the width prop down to its children. This can be used to make charts adapt to the width of their containers more easily.

```tsx
<div style={{ width: 450, border: "solid 1px black" }}>
  <AutoWidth>
    <Chart color="#fff9c4" />
  </AutoWidth>
</div>
```

image

### DIFF
--- a/packages/ui-components/src/components/AutoWidth/AutoWidth.stories.tsx
+++ b/packages/ui-components/src/components/AutoWidth/AutoWidth.stories.tsx
@@ -1,0 +1,94 @@
+import React from "react";
+import { Grid } from "@mui/material";
+import { states } from "@actnowcoalition/regions";
+import { ComponentMeta } from "@storybook/react";
+import { MetricId } from "../../stories/mockMetricCatalog";
+import { MetricSparklines } from "../MetricSparklines";
+import { AutoWidth } from ".";
+
+export default {
+  title: "Components/AutoWidth",
+  component: AutoWidth,
+} as ComponentMeta<typeof AutoWidth>;
+
+const Chart = ({ width, color }: { width?: number; color: string }) => (
+  <div
+    style={{
+      width,
+      height: 60,
+      backgroundColor: color,
+      display: "grid",
+      placeItems: "center",
+    }}
+  >
+    <div>{`width = ${width}px`}</div>
+  </div>
+);
+
+export const DefaultExample = () => (
+  <div style={{ width: 450, border: "solid 1px black" }}>
+    <AutoWidth>
+      <Chart color="#fff9c4" />
+    </AutoWidth>
+  </div>
+);
+
+export const ExampleSized = () => (
+  <div style={{ width: 450, border: "solid 1px black" }}>
+    <AutoWidth>
+      <Chart color="#fff9c4" width={120} />
+    </AutoWidth>
+  </div>
+);
+
+export const ExampleGrid = () => (
+  <Grid container spacing={2}>
+    <Grid item xs={4}></Grid>
+    <Grid item xs={4}></Grid>
+    <Grid item xs>
+      <AutoWidth>
+        <Chart color="#ffecb3" />
+      </AutoWidth>
+    </Grid>
+  </Grid>
+);
+
+const region = states.findByRegionIdStrict("12");
+
+export const ExampleSparklines = () => (
+  <Grid container spacing={2}>
+    <Grid item xs={4}>
+      <AutoWidth>
+        <MetricSparklines
+          height={90}
+          region={region}
+          metricBarChart={MetricId.MOCK_CASES}
+          metricLineChart={MetricId.MOCK_CASES}
+          numDays={30}
+        />
+      </AutoWidth>
+    </Grid>
+    <Grid item xs={4}>
+      <AutoWidth>
+        <MetricSparklines
+          height={90}
+          region={region}
+          metricBarChart={MetricId.MOCK_CASES}
+          metricLineChart={MetricId.MOCK_CASES}
+          numDays={30}
+        />
+      </AutoWidth>
+    </Grid>
+    <Grid item xs>
+      <AutoWidth>
+        <MetricSparklines
+          height={90}
+          region={region}
+          metricBarChart={MetricId.MOCK_CASES}
+          metricLineChart={MetricId.MOCK_CASES}
+          numDays={30}
+        />
+      </AutoWidth>
+    </Grid>
+  </Grid>
+);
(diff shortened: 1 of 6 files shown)


packages/ui-components/src/components/AutoWidth/AutoWidth.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+      backgroundColor: color,
+      display: "grid",
+      placeItems: "center",
+    }}
+  >
+    <div>{`width = ${width}px`}</div>
+  </div>
+);
+
+export const DefaultExample = () => (
```
chasulin: maybe calling this one `StaticWidth` to state that it shouldn't resize. overall, we wouldn't use this component for something that wouldn't need to autosize, so idk if a static width example is even needed

packages/ui-components/src/components/AutoWidth/AutoWidth.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+  <div style={{ width: 450, border: "solid 1px black" }}>
+    <AutoWidth>
+      <Chart color="#fff9c4" />
+    </AutoWidth>
+  </div>
+);
+
+export const ExampleSized = () => (
+  <div style={{ width: 450, border: "solid 1px black" }}>
```
chasulin: ditto here [re: examples with static width](https://github.com/covid-projections/act-now-packages/pull/294/files#r993583914)

packages/ui-components/src/components/AutoWidth/AutoWidth.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+    <AutoWidth>
+      <Chart color="#fff9c4" width={120} />
+    </AutoWidth>
+  </div>
+);
+
+export const ExampleGrid = () => (
```
chasulin: i dont totally understand the intention of this example. there are 2 empty grid items and one (the autosize one) that resizes in a kind of confusing way (see below)

Screen Shot 2022-10-12 at 11 05 28 AM

packages/ui-components/src/components/AutoWidth/AutoWidth.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+      <AutoWidth>
+        <Chart color="#ffecb3" />
+      </AutoWidth>
+    </Grid>
+  </Grid>
+);
+
+const region = states.findByRegionIdStrict("12");
+
+export const ExampleSparklines = () => (
```
chasulin: again, i think just showing a single sparkline resizing within a parent (maybe with a border to show the parent) would be a clearer storybook example. this one starts wrapping in confusing ways

Screen Shot 2022-10-12 at 11 07 25 AM

packages/ui-components/src/components/AutoWidth/AutoWidth.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <MetricSparklines
+          height={90}
+          region={region}
+          metricBarChart={MetricId.MOCK_CASES}
+          metricLineChart={MetricId.MOCK_CASES}
+          numDays={30}
+        />
+      </AutoWidth>
+    </Grid>
+    <Grid item xs>
```
chasulin: i think we're missing the `={4}` after `xs`

packages/ui-components/src/components/AutoWidth/AutoWidth.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { ParentSize } from "@visx/responsive";
+
+export interface AutoWidthProps {
+  /** A single child element that receives an optional `width` prop. */
+  children: React.ReactElement<{ width?: number }>;
```
chasulin, line 10: the point of this component is that its used when things need to resize automatically with their parent (so, `ParentSize` without having to write out the syntax). that being said, i feel like there shouldn't be a `width` prop, because it should always be grabbing the parent's width (and, in a case where the child isn't meant to use the parent's width, we shouldn't be using this component)

chasulin (commented):
looking good! left some comments. i think we could simplify the implementation of this component (and the storybook examples)

packages/ui-components/src/components/AutoWidth/AutoWidth.stories.tsx
pnavarrc: I think you are right, that was causing the issues we were seeing!

packages/ui-components/src/components/AutoWidth/AutoWidth.stories.tsx
pnavarrc: Keeping it just to make it more evident that it takes the width of the parent (which can be seen in the code)

packages/ui-components/src/components/AutoWidth/AutoWidth.stories.tsx
pnavarrc: Adding a note - this is just to demonstrate the behaviour of `AutoWidth` when the child already has a width property (it shouldn't overwrite it)

packages/ui-components/src/components/AutoWidth/AutoWidth.stories.tsx
pnavarrc: my bad, I misunderstood what happens when we have `xs` by itself on a Grid item (I wanted it to take the remaining space, but apparently that's not how it works 🤷

packages/ui-components/src/components/AutoWidth/AutoWidth.stories.tsx
pnavarrc: Updated, the example should be better now 😅

packages/ui-components/src/components/AutoWidth/AutoWidth.tsx
pnavarrc: This is an annotation for the type of the children. The children should receive an optional `width` and maybe other props. Defining this type for the children means that if we try to pass a component that doesn't take a `width` (in which case, we shouldn't use `AutoWidth`), TS should give us a warning saying so

pnavarrc: **Update**: I tried passing a component that doesn't take a `width` and TS didn't say anything, will look more into it and update it on a follow up PR. Added a TODO for now.

chasulin (approved):

PR #295, opened 2022-10-12
Style bottom axis

### LINKED ISSUE #292: Improve default tick handling of AxisBottom
The `AxisBottom` component automatically determines the date format and which tick marks to use, which is not always optimal (not too bad in this case actually):

Image

The main problem here is that if the chart is too narrow, the tick labels will overlap with each other. We can prevent this by changing:

- The format used to render the tick labels
- How many tick labels we include (and where)

Of course, we want to make sure that the labels don't overlap, and that they are informative and useful. Based on our previous work, we probably need to support charts as narrow as 280px and wider than 600px, and timeframes between a few days to 10+ years. I would recommend writing stories and figuring out a heuristic that gives reasonable results for a set of timespans and the aforementioned widths.

- https://airbnb.io/visx/docs/axis

faihegberg:
Closes https://github.com/covid-projections/act-now-packages/issues/292

Before:
image

After:
image
image
image

Once approved, I will adjust other time series stories as a follow-up PR.


packages/ui-components/src/components/Axis/utils.tsx
```diff
@@ -0,0 +1,18 @@
...
+    return 7;
+  }
+  return 10;
+}
+
+export function getDateFormat(dateDomain: Date[]): DateFormat {
+  const [minDate, maxDate] = dateDomain;
+  const dateRange = getTimeDiff(maxDate, minDate, TimeUnit.DAYS);
+  return dateRange < 365 ? DateFormat.MMM_D : DateFormat.MMM_YY;
+}
```
faihegberg: Although this formatting is short (which helps accommodate variable chart widths), the format can be confusing. For example, "Jan 21" can mean Jan 21st or Jan 2021, depending on whether the date range is more than 365 days. I propose we add a date format with a `'` in front of the year. So Jan 2021 would be `Jan'21`. What do you think?

pnavarrc:
For this one, maybe the month abbreviation is enough (we could remove the "1" since it doesn't really add info). 

image

packages/ui-components/src/components/Axis/utils.tsx
pnavarrc: The other option (a bit more complex) is to use a combination

```
-----|------|------|------|------|------|------|------|------|----->
  Jan'21   Apr    Jul    Oct  Jan'22   Apr    Jul    Oct   Jan'23 ...
```

note that in this case, we will need to return a formatting function instead of a date format.

packages/ui-components/src/components/Axis/utils.tsx
```diff
@@ -0,0 +1,22 @@
...
+    return 3;
+  } else if (width <= 600) {
+    return 5;
+  } else if (width <= 900) {
+    return 7;
+  }
+  return 10;
+}
+
+export function getDateFormat(dateDomain: Date[]): DateFormat {
```
pnavarrc: Consider changing the function arguments to either be a tuple or two dates

```tsx
export function getDateFormat(dateDomain: [Date, Date]): DateFormat {}
export function getDateFormat(startDate: Date, endDate: Date): DateFormat {}
```

faihegberg:
> For this one, maybe the month abbreviation is enough (we could remove the "1" since it doesn't really add info).
> 
> image

I think this is fine as long as it's under one year? I don't think it would necessarily crowd out the chart if that's your concern.

packages/ui-components/src/components/Axis/AxisLengths.stories.tsx
```diff
@@ -0,0 +1,117 @@
+import React from "react";
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { scaleTime } from "@visx/scale";
+import { AxisBottom, AxisBottomProps } from ".";
+import { isOverOneYear, getNumTicks, formatDateTick } from "./utils";
+
+export default {
+  title: "Charts/Axis Lengths",
```
chasulin: these should probably be stories for `Axis Bottom` (and maybe we could separate the `Axis Left` stories out into their own folder)

chasulin:
> > For this one, maybe the month abbreviation is enough (we could remove the "1" since it doesn't really add info).
> > image
> 
> I think this is fine as long as it's under one year? I don't think it would necessarily crowd out the chart if that's your concern.

the `1` feels redundant/unnecessary, would be cleaner without it i think

packages/ui-components/src/components/Axis/AxisLengths.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+} as ComponentMeta<typeof AxisBottom>;
+
+const wideWidth = 900;
+const mediumWidth = 600;
+const smallWidth = 280;
```
chasulin: i dont totally understand hard-coding the widths for each story. rather, i feel like the stories should be responsive/interactive. so, for instance, a "two year" story that encompasses all screen sizes by being responsive (and ditto for other time ranges)

faihegberg:
> > For this one, maybe the month abbreviation is enough (we could remove the "1" since it doesn't really add info).
> > image
> 
> I think this is fine as long as it's under one year? I don't think it would necessarily crowd out the chart if that's your concern.

Ok, in that case I'll only display the dates if the time span is less than 2 months.

chasulin (approved):
Looks good to merge! We might need to tweak things when implementing, if for instance the Units version has digits much larger than 1-10 that don't fit. But definitely good to merge

PR #296, opened 2022-10-13
Improvements to MultiProgressBar (and metric-aware version)

### LINKED ISSUE #257: Implement metric-aware `MultiProgressBar` component
_description in progress_

The metrics needs to share the same min and max values (for vaccinations, we use percentages). Since the thresholds will also need to be shared, we might need to include some texture or else 2 values will look the same.

Image

This might need some design help actually (or we could use CAN as base).

chasulin:
[Default colors](https://act-now-packages--pr296-casulin-metric-multi-219p9pan.web.app/storybook/index.html?path=/story/charts-multiprogressbar--default-colors)
[Custom colors](https://act-now-packages--pr296-casulin-metric-multi-219p9pan.web.app/storybook/index.html?path=/story/charts-multiprogressbar--custom-colors)

- Simplifies `MultiProgressBar` to take only 2 items (`firstItem` and `secondItem`) instead of an array of any amount of items. This way, we could render one bar as solid and one bar as patterned/hatched.
- Adds a `barColor` prop that dictates the color of the progress bar's solid and hatched `rect`s (the coloring shouldn't be tied to the color associated with the metric/value/threshold)

Fixes #257

### DIFF
--- a/packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
+++ b/packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
@@ -6,8 +6,7 @@ import { scaleLinear } from "@visx/scale";
 import { MultiProgressBarProps } from "./interfaces";
 
 export const MultiProgressBar = <T,>({
-  firstItem,
-  secondItem,
+  items,
   getItemValue,
   getItemLabel,
   maxValue,
@@ -32,7 +31,7 @@ export const MultiProgressBar = <T,>({
    * in the correct order in the progress bar.
    * */
   const [sortedFirstItem, sortedSecondItem] = sortBy(
-    [firstItem, secondItem],
+    items,
     (item) => getItemValue(item) * -1
   );
 
--- a/packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
+++ b/packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useId } from "react";
 import sortBy from "lodash/sortBy";
 import { useTheme } from "@mui/material";
 import { RectClipGroup } from "../RectClipGroup";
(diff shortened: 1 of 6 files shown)


packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
```diff
@@ -1,23 +1,26 @@
...
 import { useTheme } from "@mui/material";
 import { RectClipGroup } from "../RectClipGroup";
 import { scaleLinear } from "@visx/scale";
 import { MultiProgressBarProps } from "./interfaces";
 
 export const MultiProgressBar = <T,>({
-  items,
-  getItemColor,
+  firstItem,
+  secondItem,
```
pnavarrc: Maybe you can use the same pattern as `MetricMultiProgressBar` and have

```tsx
export const MultiProgressBar = <T,>({
  items: [T, T]
}) ...
```

instead of first and second items?

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
```diff
@@ -28,35 +31,60 @@ export const MultiProgressBar = <T,>({
...
+          <line
+            x1="1"
+            y1="0"
+            x2="1"
+            y2="1"
+            stroke={barColor}
+            strokeWidth={1.5}
+          />
+        </pattern>
+      </defs>
```
pnavarrc, line 57: Alternatively we could use https://airbnb.io/visx/docs/pattern to have other options available, but we can punt on that until we need other patterns.

pnavarrc (approved):
Looks great! :shipit:

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
chasulin: i originally did this (typed `items` as `[T, T]` and typescript kept complaining

mikelehen: I took a peek at this because I was curious and TypeScript seemed to behave okay? https://github.com/covid-projections/act-now-packages/pull/302 Not a big deal either way though.

chasulin: ah yes i revisited this eod on thursday and it ended up working fine. im not sure what the original issue that i saw was 🤷🏻‍♀️. pushing up shortly!

PR #301, opened 2022-10-13
Allow ripple in tabs

### LINKED ISSUE #298: Enable "focus" on tabs component
Currently when we focus on a tab without selecting it, there's no visible indication that we're focusing on that tab. This can be confusing for users navigating the tabs without a mouse, since there's no way to know which tab they're on.

The tab receiving focus should have a clear and visible indication that it is currently receiving focus.

faihegberg:
Disabling ripple made it challenging for users to tell which tab they're focusing on, particularly if they're not navigating with a mouse. This PR re-enables ripple, which highlights the tab the user focuses on.

Closes https://github.com/covid-projections/act-now-packages/issues/298

### DIFF
--- a/.changeset/quick-cherries-clean.md
+++ b/.changeset/quick-cherries-clean.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/ui-components": patch
+---
+
+Allow ripple in tabs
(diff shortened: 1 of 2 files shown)


pnavarrc (approved):
Looks good! Maybe remove the changeset since this PR doesn't change the implementation of the tabs, only storybook (which is internal)

PR #307, opened 2022-10-18
Add resize observer hook

smcclure17:
Adds a hook to track resize events for a referenced element

### DIFF
--- a/packages/ui-components/src/common/hooks/useResizeObserver.ts
+++ b/packages/ui-components/src/common/hooks/useResizeObserver.ts
@@ -1,4 +1,3 @@
-/** Helpers and React Hooks relating to share images and share links. */
 import { useEffect, useState, useCallback, useRef } from "react";
 
 /** Hook to track element resizing using the ResizeObserver API
--- a/packages/ui-components/src/common/hooks/useResizeObserver.ts
+++ b/packages/ui-components/src/common/hooks/useResizeObserver.ts
@@ -1,0 +1,28 @@
+/** Helpers and React Hooks relating to share images and share links. */
+import { useEffect, useState, useCallback, useRef } from "react";
+
+/** Hook to track element resizing using the ResizeObserver API
+ *
+ * Interpolated from: https://tobbelindstrom.com/blog/resize-observer-hook/
+ */
+export const useResizeObserver = () => {
+  const [observerEntry, setObserverEntry] =
+    useState<ResizeObserverEntry | null>(null);
+  const [node, setNode] = useState<HTMLElement | null>(null);
+  const observer = useRef<ResizeObserver | null>(null);
+  const disconnect = useCallback(() => observer.current?.disconnect(), []);
+
+  const observe = useCallback(() => {
+    observer.current = new ResizeObserver(([entry]) => setObserverEntry(entry));
+    if (node) {
+      observer.current.observe(node);
+    }
+  }, [node]);
+
+  useEffect(() => {
+    observe();
+    return () => disconnect();
+  }, [disconnect, observe]);
+
+  return { ref: setNode, observerEntry: observerEntry };
+};
--- a/packages/ui-components/src/common/hooks/useResizeObserver.ts
+++ b/packages/ui-components/src/common/hooks/useResizeObserver.ts
@@ -1,7 +1,7 @@
 import { useEffect, useState, useCallback, useRef } from "react";
 
-interface useResizeObserverResult {
-  setNodeState: React.Dispatch<React.SetStateAction<HTMLElement | null>>;
+interface UseResizeObserverResult {
+  setObservedNode: React.Dispatch<React.SetStateAction<HTMLElement | null>>;
   observerEntry: ResizeObserverEntry | null;
 }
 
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/common/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,27 @@
...
+/** Hook to track element resizing using the ResizeObserver API
+ *
+ * Interpolated from: https://tobbelindstrom.com/blog/resize-observer-hook/
+ */
+export const useResizeObserver = () => {
+  const [observerEntry, setObserverEntry] =
+    useState<ResizeObserverEntry | null>(null);
+  const [node, setNode] = useState<HTMLElement | null>(null);
+  const observer = useRef<ResizeObserver | null>(null);
+  const disconnect = useCallback(() => observer.current?.disconnect(), []);
```
mikelehen: Sorry I didn't flag this before, but this looks suspicious.  `observer.current?.disconnect` could be undefined (since we're suing `?`), so it seems like it should be `observer.current?.disconnect?.()` or `observer.current && observer.current.disconnect()` or similar since `undefined()` would crash.

packages/ui-components/src/common/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect, useState, useCallback, useRef } from "react";
+
+/** Hook to track element resizing using the ResizeObserver API
+ *
+ * Interpolated from: https://tobbelindstrom.com/blog/resize-observer-hook/
+ */
+export const useResizeObserver = () => {
```
mikelehen: nit: Can you rewrite this as a `export function useResizeObserver() { ... `. I think it's slightly more idiomatic for top-level functions to be written as `function`.  Also, can you declare the return type and add a `@returns` doc comment?

packages/ui-components/src/common/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,27 @@
...
+      observer.current.observe(node);
+    }
+  }, [node]);
+
+  useEffect(() => {
+    observe();
+    return () => disconnect();
+  }, [disconnect, observe]);
+
+  return { ref: setNode, observerEntry: observerEntry };
```
mikelehen: `ref: setNode` doesn't look right...  Should it be `ref: observer` or something?  I guess I'm not actually sure what the ref is for.  But `setNode` isn't a ref.

mikelehen (approved):
Flagged a couple things (sorry I didn't in the original review), but overall looks good.

packages/ui-components/src/common/hooks/useResizeObserver.ts
smcclure17: We use `ref` here to set/indicate which element we want to track with something like:

```tsx
const { ref, observerEntry } = useResizeObserver()
...
<div ref={ref}>to be tracked</div>
```

To be entirely honest I don't fully understand exactly how it works, but I think we use a state object instead of a ref object because [`useRef()`](https://reactjs.org/docs/hooks-reference.html#useref) doesn't trigger updates/re-renders on changes. From the docs: 

> ...useRef doesn’t notify you when its content changes. Mutating the .current property doesn’t cause a re-render...

I've updated the name here to `setNodeState` in 16fc6edca033a7049bfea3686215df42770a9e36, but LMK if this doesn't feel sufficient

packages/ui-components/src/common/hooks/useResizeObserver.ts
smcclure17: Good catch--I've updated this to `observer.current?.disconnect?.()` in https://github.com/covid-projections/act-now-packages/commit/16fc6edca033a7049bfea3686215df42770a9e36

packages/ui-components/src/common/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,34 @@
+import { useEffect, useState, useCallback, useRef } from "react";
+
+interface useResizeObserverResult {
```
mikelehen: interfaces / classes should always start with a capital letter.  Also, I didn't realize how long this return type would be when I asked you to add it. Sorry! 😆

packages/ui-components/src/common/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,34 @@
...
+      observer.current.observe(node);
+    }
+  }, [node]);
+
+  useEffect(() => {
+    observe();
+    return () => disconnect();
+  }, [disconnect, observe]);
+
+  return { setNodeState: setNode, observerEntry: observerEntry };
```
mikelehen: I think `setNode` would be sufficient, or maybe `setObservedNode`?

packages/ui-components/src/common/hooks/useResizeObserver.ts
mikelehen: Thanks! Yeah, I think you are right.  I just didn't realize you could pass a setter function to `<Box ref=...>` in React.  I can't find docs that say you can, but it must be true or this wouldn't work. 🤷

mikelehen (commented):
Couple nits, but otherwise looks good!

mikelehen (approved):

packages/ui-components/src/common/hooks/useResizeObserver.ts
smcclure17: oops thanks 😅 and np! I think it being such a long return type probably makes it even more worthwhile to be explicit

PR #310, opened 2022-10-19
Implement `MetricLineThresholdChart` and minor fixes

### LINKED ISSUE #311: Implement MetricLineThresholdChart component

pnavarrc:
Close https://github.com/covid-projections/act-now-packages/issues/311

## Changes

- Implement the `useHoveredDate` hook to be used in charts where we use `ChartOverlayX`
- Update `MetricLineChart` to use the `useHoveredDate` hook
- Implement the `MetricLineThresholdChart` component

image

### DIFF
--- a/packages/ui-components/src/components/ChartOverlayX/useHoveredDate.tsx
+++ b/packages/ui-components/src/components/ChartOverlayX/useHoveredDate.tsx
@@ -1,0 +1,54 @@
+import { useState } from "react";
+import { Timeseries, TimeseriesPoint } from "@actnowcoalition/metrics";
+import { ChartOverlayXProps } from "./ChartOverlayX";
+
+/**
+ * React hook that keeps track of the point being hovered in a chart. This
+ * hook is normally used with ChartOverlayX. Given a timeseries, it returns
+ * the handlers onMouseMove, onMouseLeave and hoveredPoint, the point in
+ * the timeseries closest to the date being hovered.
+ *
+ * @example
+ * ```tsx
+ * const { hoveredPoint, onMouseMove, onMouseLeave } = useHoveredDate(timeseries);
+ *
+ * return (
+ *   <svg width={width} height={height}>
+ *     <LineChart {...lineChartProps} />
+ *     {hoveredPoint && (
+ *       <Tooltip point={hoveredPoint}>
+ *         <CircleMarker />
+ *       </Tooltip>
+ *     )}
+ *     <ChartOverlayX
+ *       onMouseMove={onMouseMove}
+ *       onMouseLeave={onMouseLeave}
+ *       {...otherProps}
+ *     />
+ *   </svg>
+ * );
+ * ```
+ *
+ * @param timeseries The timeseries with the points being hovered.
+ * @returns The hovered point and the onMouseMove and onMouseLeave handlers.
+ */
+export function useHoveredDate<T>(timeseries: Timeseries<T> | undefined) {
+  const [hoveredPoint, setHoveredPoint] = useState<TimeseriesPoint<T> | null>(
+    null
+  );
+
+  const onMouseLeave = () => setHoveredPoint(null);
+
+  const onMouseMove: ChartOverlayXProps["onMouseMove"] = ({ date }) => {
+    if (!timeseries) {
+      return;
+    }
+
+    const point = timeseries.findNearestDate(date);
+    if (point) {
+      setHoveredPoint(point);
+    }
+  };
+
+  return { hoveredPoint, onMouseMove, onMouseLeave };
+}
--- a/packages/ui-components/src/components/MetricLineChart/index.ts
+++ b/packages/ui-components/src/components/MetricLineChart/index.ts
@@ -1,2 +1,3 @@
 export * from "./interfaces";
 export * from "./MetricLineChart";
+export * from "./MetricLineChart.styles";
--- a/packages/ui-components/src/components/MetricLineChart/index.ts
+++ b/packages/ui-components/src/components/MetricLineChart/index.ts
@@ -1,3 +1,2 @@
 export * from "./interfaces";
 export * from "./MetricLineChart";
-export * from "./MetricLineChart.styles";
--- a/packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
+++ b/packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
@@ -1,0 +1,109 @@
+import React, { useState } from "react";
+import { scaleLinear, scaleTime } from "@visx/scale";
+import { Group } from "@visx/group";
+import { Metric, TimeseriesPoint } from "@actnowcoalition/metrics";
+import { Region } from "@actnowcoalition/regions";
+import { useData } from "../../common/hooks";
+import { AxesTimeseries } from "../Axes";
+import { ChartOverlayX, ChartOverlayXProps } from "../ChartOverlayX";
+import { LineChart } from "../LineChart";
+import { useMetricCatalog } from "../MetricCatalogContext";
+import { MetricTooltip } from "../MetricTooltip";
+import { BaseChartProps, CircleMarker } from "../MetricLineChart";
+
+export interface MetricLineThresholdChartProps extends BaseChartProps {
+  metric: Metric | string;
+  region: Region;
+}
+
+export const MetricLineThresholdChart = ({
+  metric: metricOrId,
+  region,
+  width,
+  height,
+  marginTop = 10,
+  marginBottom = 30,
+  marginLeft = 70,
+  marginRight = 20,
+}: MetricLineThresholdChartProps) => {
+  const metricCatalog = useMetricCatalog();
+  const metric = metricCatalog.getMetric(metricOrId);
+  const { data } = useData(region, metric, true);
+
+  const [hoveredPoint, setHoveredPoint] =
+    useState<TimeseriesPoint<number> | null>(null);
+
+  const onMouseLeave = () => {
+    setHoveredPoint(null);
+  };
+  const onMouseMove: ChartOverlayXProps["onMouseMove"] = ({ date }) => {
+    const point = timeseries.findNearestDate(date);
+    if (point) {
+      setHoveredPoint(point);
+    }
+  };
+
+  if (!data) {
+    return null;
+  }
+
+  const timeseries = data.timeseries.assertFiniteNumbers();
+
+  if (!timeseries.hasData()) {
+    return null;
+  }
+
+  const chartHeight = height - marginTop - marginBottom;
+  const chartWidth = width - marginLeft - marginRight;
+
+  const { minDate, maxDate, maxValue } = timeseries;
+
+  const dateScale = scaleTime({
+    domain: [minDate, maxDate],
+    range: [0, chartWidth],
+  });
+
+  const yScale = scaleLinear({
+    domain: [0, maxValue],
+    range: [chartHeight, 0],
+  });
+
+  return (
+    <svg width={width} height={height}>
+      <Group left={marginLeft} top={marginTop}>
+        <AxesTimeseries
+          height={chartHeight}
+          dateScale={dateScale}
+          yScale={yScale}
+          axisLeftProps={{ tickFormat: (value) => metric.formatValue(value) }}
+        />
+        <LineChart timeseries={timeseries} xScale={dateScale} yScale={yScale} />
+        {hoveredPoint && (
+          <MetricTooltip
+            metric={metric}
+            region={region}
+            point={hoveredPoint}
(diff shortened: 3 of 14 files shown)


packages/ui-components/src/components/MetricLineChart/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from "./interfaces";
 export * from "./MetricLineChart";
+export * from "./MetricLineChart.styles";
```
pnavarrc: We need `CircleMarker` from here, we should probably make that its own component (although it barely has any logic in it)

packages/ui-components/src/components/ChartOverlayX/useHoveredDate.tsx
```diff
@@ -0,0 +1,54 @@
...
+ *       </Tooltip>
+ *     )}
+ *     <ChartOverlayX
+ *       onMouseMove={onMouseMove}
+ *       onMouseLeave={onMouseLeave}
+ *       {...otherProps}
+ *     />
+ *   </svg>
+ * );
+ * ```
```
mikelehen, line 30: 🙏  for example!

packages/ui-components/src/components/MetricLineChart/index.ts
mikelehen: Feels a bit weird to export our styles.  This will probably end up being exported from our ui-components module too, right?  Not sure we want that much surface area.

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
```diff
@@ -0,0 +1,151 @@
...
+export interface MetricLineThresholdChartProps extends BaseChartProps {
+  metric: Metric | string;
+  region: Region;
+}
+
+/**
+ * MetricLineThresholdChart renders a line chart where line segments are
+ * colored according to the metric level and threshold.
+ *
+ * For example, if we have a metric with levels two levels: High (red) and
```
mikelehen: levels two levels => two levels

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
```diff
@@ -0,0 +1,151 @@
...
+ * Low (green) separated by a threshold at the value 10, the line will
+ * be red above when above 10 and green below it.
```
mikelehen: above when above => when above

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
```diff
@@ -0,0 +1,151 @@
...
+  height,
+  marginTop = 10,
+  marginBottom = 30,
+  marginLeft = 70,
+  marginRight = 20,
+}: MetricLineThresholdChartProps) => {
+  const metricCatalog = useMetricCatalog();
+  const metric = metricCatalog.getMetric(metricOrId);
+
+  const { data } = useData(region, metric, true);
```
mikelehen: nit: consider documenting the boolean literal for readability, i.e. `/*includeTimeseries=*/true`

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+  const { hoveredPoint, onMouseMove, onMouseLeave } = useHoveredDate<number>(
+    data?.timeseries as Timeseries<number>
```
mikelehen: Why do we need this `as` cast?  It seems incorrect (`data?.timeseries` could be undefined).  In general, I don't see why we need to cast at all.

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+  const dateScale = scaleTime({
+    domain: [minDate, maxDate],
+    range: [0, chartWidth],
+  });
+
+  const yScale = scaleLinear({
+    domain: [0, maxValue],
+    range: [chartHeight, 0],
+  });
```
mikelehen, line 70: Do we want to scale the chart so that all of the thresholds are always visible (and have enough space to include the label)?  E.g. in CAN for Vermont:
image

We either need some heuristic here or to let the user pass in a value to override `maxValue` I guess.  Or both. 😁

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+  const { minDate, maxDate, maxValue } = timeseries;
+
+  const dateScale = scaleTime({
+    domain: [minDate, maxDate],
+    range: [0, chartWidth],
+  });
+
+  const yScale = scaleLinear({
+    domain: [0, maxValue],
```
mikelehen, line 68: FWIW- We might do this in other charts too, but it occurs to me that this will break if we have metrics that can have negative values.  I think it might make more sense to do something like:

```
// Ensure the chart axis will include 0.
const minValue = (timeseries.minValue > 0) ? 0 : timeseries.minValue
```

and then use `minValue` instead of `0` in the rest of the code.  Though maybe we should allow chart axes that don't include 0 in some cases.  Might need an option.

Feel free to punt if we're not trying to support negative values at this point.  We can rework things later.

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+  assert(
+    metric.thresholds?.length && metric.levelSet?.levels,
+    `MetricLineThresholdChart can only be used with metrics that have thresholds and levels ${metric}`
+  );
+
+  const intervals = calculateChartIntervals(
+    metric.levelSet.levels,
+    metric.thresholds,
+    0,
```
mikelehen: nit: consider documenting the numeric literal for readability, i.e. `/*minValue=*/0`

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
```diff
@@ -0,0 +1,151 @@
...
+          const clipHeight = Math.abs(yFrom - yTo);
+          return (
+            <RectClipGroup
+              key={`rect-clip-${interval.level.id}`}
+              y={Math.min(yFrom, yTo)}
+              width={chartWidth}
+              height={clipHeight}
+            >
+              <LineChart
+                timeseries={data.timeseries as Timeseries<number>}
```
mikelehen: Again, not psyched about this cast.  Can we just use`timeseries`?

packages/ui-components/src/components/MetricLineThresholdChart/utils.ts
```diff
@@ -0,0 +1,112 @@
...
+ *   { lowerBound: 20, upperBound: 60, level: LEVEL_HIGH },
+ * ]
+ * ```
+ *
+ * Note that levels that don't intersect with the interval [minValue, maxValue]
+ * won't be returned, since we won't need them in the chart.
+ *
+ * @example
+ * ```ts
+ *  * calculateChartIntervals(
```
mikelehen: Extra `*`?

packages/ui-components/src/components/MetricLineThresholdChart/utils.ts
```diff
@@ -0,0 +1,112 @@
...
+ * @param thresholds List of thresholds.
+ * @param minValue Minimum value to show in the chart.
+ * @param maxValue Maximum value to show in the chart.
+ * @returns List of chart intervals.
+ */
+export function calculateChartIntervals(
+  metricLevels: MetricLevel[],
+  thresholds: number[],
+  startValue: number,
+  endValue: number
```
mikelehen: The doc comments have minValue/maxValue but here you use start/end.  Can we make them match?

packages/ui-components/src/components/MetricLineThresholdChart/utils.ts
```diff
@@ -0,0 +1,112 @@
...
+  assert(
+    metricLevels.length === thresholds.length + 1,
+    `There should be one fewer threshold than levels`
+  );
+
+  const minValue = Math.min(startValue, endValue);
+  const maxValue = Math.max(startValue, endValue);
+
+  // Reverse the metric levels and thresholds if the thresholds are descending
+  if (thresholds.length >= 2 && thresholds[1] - thresholds[0] < 0) {
```
mikelehen: Any reason to do `thresholds[1] - thresholds[0] < 0` instead of just `thresholds[0] > thresholds[1]` ?

FWIW- If you passed in the `metric` you could use `metric.areThresholdsDescending`.

packages/ui-components/src/stories/mockMetricCatalog.ts
```diff
@@ -31,7 +31,7 @@ const testMetricDefs: MetricDefinition[] = [
       providerId: "mock",
       startDate: "2020-01-01",
     },
-    thresholds: [10, 100],
+    thresholds: [40, 100],
```
mikelehen, line 34: This is probably fine, but just checking that it was intentional?

mikelehen (approved):
Left some comments, but overall looks good!

packages/ui-components/src/components/MetricLineChart/index.ts
mikelehen: Yeah, I'd duplicate it or go ahead and import it via the full `../MetricLineChart/MetricLineChart.styles` or whatever. 🤷

packages/ui-components/src/components/MetricLineThresholdChart/utils.ts
pnavarrc: oh that's better!

packages/ui-components/src/stories/mockMetricCatalog.ts
pnavarrc: yeah, the mock cases timeseries wasn't hitting the thresholds, so I changed it - I might update this function to take the metric, in which case I will need to create metric and metric data instances to cover the cases I need

packages/ui-components/src/components/MetricLineChart/index.ts
pnavarrc: I'm going to import it from the full styles path for now, but I think it might make sense to create it as its own component, will add that as a topic for next week!

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
pnavarrc: **EDIT**: I think I found a better way, preserving the explanation below

---

I wand to have `hoveredPoint` to have type `TimeseriesPoint<number> | null` so I can keep the tooltip code simple [here](https://github.com/covid-projections/act-now-packages/blob/782ad8b546002f8e64b7bebc0c0a615e0b3a82a2/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx#L66)

```tsx
{hoveredPoint && (
  <MetricTooltip point={hoveredPoint} {...}>
    <CircleMarker
      cx={dateScale(hoveredPoint.date)}
      cy={yScale(hoveredPoint.value)}
      {...}
    />
  </MetricTooltip>
)}
```

When I call `useHoveredDate<number>(data?.timeseries)`, `data?.timeseries` has type `Timeseries<unknown> | undefined`. The `useHoveredDate` hook is okay with the `undefined` part, but not with `Timeseries<unknown>` (that will break the type of the point?)

Another option is to be more explicit when casting `data?.timeseries`, but maybe there is a better way? If I could initialize the hook after `data.timeseries.assertFiniteNumber()` it would be great, but that's not possible because we have a conditional in the middle 🤷 

```tsx
const { hoveredPoint, onMouseMove, onMouseLeave } = useHoveredDate<number>(
  data?.timeseries as Timeseries<number> | undefined
);

if (!data) {
  return null;
}

const timeseries = data.timeseries.assertFiniteNumbers();

if (!timeseries.hasData()) {
  return null;
}
```

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
pnavarrc: we should support negative values - I'm going to punt on this for now and make a note so we can revisit all charts together, good catch!

pnavarrc:
I'm going to handle the logic to show all the levels and negative metric values on a follow up PR, thanks for the review! 🙏

PR #312, opened 2022-10-19
Reorganize misc files/folders

chasulin:
Don't be spooked by the large diff in this PR. It's all just renaming some files and moving them around.

Per our recent discussion in standup about file structure, this PR reorgs a few files/folders:

- Renames `Axes` folder to `AxexTimeseries` (to match the folder's contents)
- Separates regular and metric-aware implementations of US Maps (Maps folder -> `USMaps` folder and `MetricUSMaps` folder)
- Renames common US map props from RenderMapProps to `BaseUSMapProps`
- Removes geo-albers-us-territories package that we don't seem to be using anywhere

PR #313, opened 2022-10-19
Metric mini map

### LINKED ISSUE #283: Implement MiniMap container with metric selector
We need to implement a metric-aware MiniMap that receives a region and a list of list of metrics, so users can select the metric that they want to use to see the map.

## Notes

There are 3 cases in the mocks: 

1. Only 1 metric - No controls to change the metric
2. Exactly 2 metrics - Uses a Tab component to select the active metric
3.  More than 2 metrics - Includes a dropdown to select the active metric

I think we should only implement 1 and 3 (since the dropdown can be used to select metrics when we only have 2). I would recommend not worrying about making the component sticky for now (as it is in CAN) because that's a concern of the app, we will take care of that once we start implementing better containers in the template.

- [ActNow Design System - MiniMap](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=118%3A4529)

faihegberg:
Closes https://github.com/covid-projections/act-now-packages/issues/283

Implements a mini map container which accepts a region and a collection of metrics.

image

image

image

image

**Notes**
- Currently for the single metric display, if you zoom in you'll notice the top right hand corner has a teeny tiny detail where the map appears to be sitting on top of the paper component and is not conforming to the border radius. Not exactly sure how we can enforce the map to conform to the border radius. Let me know if you have any ideas!
- Currently the metric's extended name is left-aligned instead of center-aligned like the mocks. We will need to adjust `MetricLegendThreshold` to explicitly center the text. I held off on this for now as it might affect other components. But I think it would be good to explicitly center it.


packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+function getEndLabel(metric: Metric): string {
+  if (metric.levelSet) {
+    const levels = metric.levelSet.levels;
+    return startCase(levels[levels.length - 1].id);
+  } else if (metric.categories) {
+    const categories = metric.categories;
+    return startCase(categories[categories.length - 1].label);
+  } else return "High";
+}
```
faihegberg: Wonder whether we should establish a more standardized way to get the start and end label. I currently just take the first item in the level set / categories as the lowest and the last item as the highest, which is pretty fragile (i.e. there's no guaranteeing that the first is always the lowest and the last is always the highest).

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
```diff
@@ -0,0 +1,107 @@
...
+  renderTooltip,
+}) => {
+  const theme = useTheme();
+  const metricCatalog = useMetricCatalog();
+  const [option, setOption] = useState(metrics[0]);
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setOption(event.target.value);
+  };
+  return (
+    <Paper sx={{ width: theme.spacing(39) }}>
```
faihegberg: `theme.spacing(39)` might be a little hard to read. Wonder if we should just stick with pixels here. No strong feelings though!

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
```diff
@@ -0,0 +1,107 @@
...
+          label={<Typography variant="paragraphSmall">Metric</Typography>}
+          value={option}
+          onChange={handleChange}
+          sx={{
+            margin: 0,
+            "& .MuiFilledInput-root": {
+              borderBottomLeftRadius: 0,
+              borderBottomRightRadius: 0,
+            },
+          }}
```
faihegberg: Pretty minor styling so I didn't separate this into its own style file. Let me know if you think otherwise though!

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.stories.tsx
```diff
@@ -0,0 +1,56 @@
+import React from "react";
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { MetricMiniMap } from ".";
+import { MetricId } from "../../stories/mockMetricCatalog";
+import { states, counties, Region, RegionDB } from "@actnowcoalition/regions";
+
+export default {
+  title: "Maps/MetricMiniMap",
```
mikelehen: Should we move the component files into the components/Maps/ directory to match?

Although @chasulin just reworked the `Maps/` directory in https://github.com/covid-projections/act-now-packages/pull/312 (it's now `USMaps/`).  So maybe we should merge that first?

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
```diff
@@ -0,0 +1,113 @@
+import React, { useState } from "react";
+import { RegionDB, Region } from "@actnowcoalition/regions";
+import { MetricUSStateMap } from "../Maps";
```
mikelehen: Given that `MetricMiniMap` only works with states, it feels like the name should be `MetricUSStateMiniMap` or something.  Though I'm also wondering if we can come up with a better differentiator than just "mini"...  I'm not sure though, so maybe just stick with "mini" for now.

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+  component: MetricMiniMap,
+} as ComponentMeta<typeof MetricMiniMap>;
+
+const regionDB = new RegionDB([...states.all, ...counties.all], {
+  getRegionUrl: (region: Region) => `/us/${region.slug}`,
+});
+
+const renderTooltip = (regionId: string) => {
+  return regionDB.findByRegionIdStrict(regionId).fullName;
+};
```
mikelehen: Is there a reason we wouldn't want to just make this the builtin tooltip behavior so you don't have to provide a `renderTooltip` handler?  Would make the component easier to use.

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
```diff
@@ -0,0 +1,113 @@
...
+  stateRegionId: string;
+  currentRegion?: Region;
+  metrics: (Metric | string)[];
+  regionDB: RegionDB;
+  renderTooltip: (regionId: string) => React.ReactElement | string;
+}
+
+function getStartLabel(metric: Metric): string {
+  if (metric.levelSet) {
+    const levels = metric.levelSet.levels;
```
mikelehen: Note that I reworked levels in #319 and so once that lands this will end up being `categorySet` and `categorySet.categories` and you won't need two separate cases for levels vs categories.

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
mikelehen: I feel like this startLabel / endLabel stuff should be built into `MetricLegendThreshold`. Or at least it should have a default implementation that does this.  Can you open a github issue for us to consider doing that?

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
```diff
@@ -0,0 +1,113 @@
...
+    return startCase(levels[0].id);
+  } else if (metric.categories) {
+    return startCase(metric.categories[0].label);
+  } else return "Low";
```
mikelehen: I feel like this should be an error, e.g.:

```
fail(`Metric must have categories to be used with MetricMiniMap. ${metric} does not.`);
```

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
```diff
@@ -0,0 +1,113 @@
...
+}
+
+function getEndLabel(metric: Metric): string {
+  if (metric.levelSet) {
+    const levels = metric.levelSet.levels;
+    return startCase(levels[levels.length - 1].id);
+  } else if (metric.categories) {
+    const categories = metric.categories;
+    return startCase(categories[categories.length - 1].label);
+  } else return "High";
```
mikelehen: ditto

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
```diff
@@ -0,0 +1,113 @@
...
+}
+
+export const MetricMiniMap: React.FC<MetricMiniMapProps> = ({
```
mikelehen: Can you add a doc comment explaining what this component is?

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
mikelehen: Hrm, yeah. I assumed we only use spacing() for spacing between components, etc.  For specifying the actual size of a component, we'd use pixels.  But I'm not sure.

That said, in general it's not obvious to me that this component should have a fixed size (rather than letting the consumer of the component control the size).  So maybe it should accept width/height props?  But I'm not really sure how we expect layout stuff like this to work.   Maybe @pnavarrc can weigh in on how he'd imagine this working.

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
```diff
@@ -0,0 +1,113 @@
...
+              borderBottomRightRadius: 0,
+            },
+          }}
+        >
+          {metrics.map((metric) => (
+            <MenuItem
+              key={metricCatalog.getMetric(metric).id}
+              value={metricCatalog.getMetric(metric).id}
+            >
+              <Typography>{metricCatalog.getMetric(metric).name}</Typography>
```
mikelehen: Rather than doing `metricCatalog.getMetric(metric)` over and over (I see 5 I think), I'd suggest doing:

```
const resolvedMetrics = metrics.map(m => metricCatalog.getMetric(m));
```
at the top and then using `resolvedMetrics.map(metric => ...)` down here so they're guaranteed to be `Metric` objects...

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
```diff
@@ -0,0 +1,113 @@
...
+        stateRegionId={stateRegionId}
+        currentRegion={currentRegion}
+        metric={metric}
+        regionDB={regionDB}
+        renderTooltip={renderTooltip}
+      />
+      <Box padding={3}>
+        <MetricLegendThreshold
+          orientation="horizontal"
+          width={160}
```
mikelehen: using pixels here but spacing() above seems inconsistent.

mikelehen (changes requested):
Generally looks good but left a few questions / suggestions!

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
pnavarrc: Yeah, theme spacing shouldn't be used for component widths. In this case, I would recommend adding an optional `width` prop to the component to match `USStateMap` (we can grab the default width from it too)

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
pnavarrc: yeah, consider using AutoWidth or just the `width` prop (once added)

packages/ui-components/src/index.ts
```diff
@@ -109,3 +109,5 @@ export * from "./components/RectClipGroup";
 export * from "./components/RegionSearch";
 export * from "./components/ShareButton";
 export * from "./components/SparkLine";
+
+export * from "./components/MetricMiniMap";
```
pnavarrc: Maybe put this in order once the component is renamed?

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.stories.tsx
chasulin: i renamed the files/folder to `MetricUSMiniMap`, to express that its a US map (and we could eventually have non-US versions of this down the road.. although they could probably be unified into a single minimap component, but we can cross that bridge/rename when/if it comes to it?)

and, i kept it in its own directory (`MetricUSMiniMap`), separate of the `USMaps` and `MetricUSMaps` directories, because those are for the base map components, and this is a more specific thing composed of multiple components (legend, map, dropdown menu, etc.)

the file structuring can get a little head-y, so im not very tied to any of this^^. if something else makes more sense to anyone reading this, please feel free to chime in 😎

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.stories.tsx
chasulin: hm, nixing `renderTooltip` prop would nix the ability to customize the tooltip beyond just displaying the region name. im not opposed, but might be limiting down the line. we can make `renderTooltip` optional and set this^ as the default for now

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
chasulin: ah yes, see [here](https://github.com/covid-projections/act-now-packages/pull/313/files#r1003665440)

re: "mini", im getting rid of the built-in width, so now it can be flexible/any size/dependent on a parent container. i think we're outgrowing the term "mini" ..

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
chasulin: setting the width to depend on whatever wraps the component to keep things flexible

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
mikelehen: Maybe we should have some term for components that are standalone widgets ready to be dropped into the homepage / location page.  E.g. `MetricUSStateMapWidget` or `MetricUSStateMapBlock` or something?  Idunno.

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.stories.tsx
mikelehen: Sounds good!

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.stories.tsx
mikelehen: Sounds reasonable!

packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.tsx
chasulin: added it to the queue [here](https://github.com/covid-projections/act-now-packages/issues/325)

packages/ui-components/src/components/MetricUSMiniMap/MetricUSMiniMap.style.tsx
```diff
@@ -0,0 +1,13 @@
...
+export const BorderedContainer = styled(Box)`
+  border: ${({ theme }) => `1px solid ${theme.palette.border.default}`};
+`;
+
+export const BorderedContainerLast = styled(BorderedContainer)`
+  border-bottom-left-radius: ${({ theme }) => theme.shape.borderRadius}px;
+  border-bottom-right-radius: ${({ theme }) => theme.shape.borderRadius}px;
+  border-top: none;
+  padding: ${({ theme }) => theme.spacing(3)};
+`;
```
chasulin: put these into a style file because it was a bit crowded to write the second set of styles inline

mikelehen:
Still reviewing, but I'm seeing issues with the thermometer width on some screen sizes (e.g. "Small Mobile" in Storybook):
image

packages/ui-components/src/components/MetricUSMaps/MetricUSNationalMap/MetricUSNationalMap.tsx
```diff
@@ -19,7 +19,9 @@ export const MetricUSNationalMap: React.FC<MetricUSNationalMapProps> = ({
     <USNationalMap
       getFillColor={(regionId: string) => {
         const region = regionDB.findByRegionId(regionId);
-        return region ? data.metricData(region, metric).getColor() : "#eee";
+        return region && data.hasMetricData(region, metric)
+          ? data.metricData(region, metric).getColor()
+          : "#eee";
```
mikelehen: FWIW- If you merge `develop` in we should be able to revert this change here and in `MetricUSStateMap.tsx` below.  I did this as a temporary fix, but the real fix is in now.

chasulin:
> Still reviewing, but I'm seeing issues with the thermometer width on some screen sizes (e.g. "Small Mobile" in Storybook): image

yeah im adding a github issue to work on the responsiveness of the legend/therm. lets ignore the responsiveness that component for the purpose of this pr for now

packages/ui-components/src/components/MultiMetricUSStateMap/MultiMetricUSStateMap.tsx
```diff
@@ -0,0 +1,106 @@
...
+} from "./MultiMetricUSStateMap.style";
+
+export interface MultiMetricUSStateMapProps {
+  /** Region ID of the state being mapped */
+  stateRegionId: string;
+  /** Region corresponding to the page on which the component is being rendered */
+  currentRegion?: Region;
+  /** Array of metric options that can be used to color the map */
+  metrics: (Metric | string)[];
+  /** All regions to include on the map */
```
mikelehen: This could be a bit confusing (we aren't going to include all the regions in the region db on the map).  Perhaps just say "Region DB instance (used for generating region links, etc.)." or similar.

packages/ui-components/src/components/MultiMetricUSStateMap/MultiMetricUSStateMap.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { getStartLabel, getEndLabel } from "./utils";
+import {
+  BorderedContainer,
+  BorderedContainerLast,
+} from "./MultiMetricUSStateMap.style";
+
+export interface MultiMetricUSStateMapProps {
+  /** Region ID of the state being mapped */
+  stateRegionId: string;
+  /** Region corresponding to the page on which the component is being rendered */
```
mikelehen: This makes some assumptions about how this component is being used.  I think I would just say "An optional region to highlight on the map." or something.  As an aside, I think "selectedRegion" would be a bit clearer name, but then we'd need to update the other map components as well.

packages/ui-components/src/components/MultiMetricUSStateMap/MultiMetricUSStateMap.tsx
```diff
@@ -0,0 +1,106 @@
...
+  currentRegion?: Region;
+  /** Array of metric options that can be used to color the map */
+  metrics: (Metric | string)[];
+  /** All regions to include on the map */
+  regionDB: RegionDB;
+  /** Function returning the contents of the map tooltip, given a hovered region */
+  renderTooltip: (regionId: string) => React.ReactElement | string;
```
mikelehen: I think we want this to be optional, so it should be `renderTooltip?: ...` ?

also nit: Maybe use `React.ReactNode` which includes `string` already?  Or is that too permissive for some reason?

packages/ui-components/src/components/MultiMetricUSStateMap/utils.ts
```diff
@@ -0,0 +1,25 @@
+import { Metric } from "@actnowcoalition/metrics";
+import startCase from "lodash/startCase";
+
+// TODO (Chelsi) - move these to MetricLegendThreshold (or somewhere more central than here)
```
mikelehen: You could do "TODO(#325)" to reference the github issue...

mikelehen (approved):
Few small suggestions, but looks great. Thanks for getting this over the finish line.

packages/ui-components/src/components/MultiMetricUSStateMap/MultiMetricUSStateMap.tsx
chasulin: good call

PR #314, opened 2022-10-19
Fix US maps to handle when metric data is not available yet.

mikelehen:


packages/ui-components/src/components/MetricMiniMap/MetricMiniMap.stories.tsx
```diff
@@ -24,7 +24,7 @@ const Template: ComponentStory<typeof MetricMiniMap> = (args) => (
 export const Example = Template.bind({});
 Example.args = {
   stateRegionId: "36",
-  metrics: [MetricId.MOCK_CASES, MetricId.PASS_FAIL, MetricId.PI],
```
mikelehen, line 27: The PI metric doesn't have thresholds or categories to color by so it was erroring for other reasons...

faihegberg (approved):
Thank you Michael!

PR #317, opened 2022-10-20
Overview banner component

### LINKED ISSUE #199: Implement overview banner
[Figma - ActNow Design System](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=32%3A5611)

Image

chasulin:
closes #199 

[Storybook examples](https://act-now-packages--pr317-casulin-overview-ban-4vxac1c3.web.app/storybook/index.html?path=/story/components-overviewbanner--all-content)

[Figma mocks](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System)

### DIFF
--- a/packages/ui-components/src/components/OverviewBanner/OverviewBanner.tsx
+++ b/packages/ui-components/src/components/OverviewBanner/OverviewBanner.tsx
@@ -1,0 +1,56 @@
+import React from "react";
+import { Stack, Typography, Button } from "@mui/material";
+import { useTheme } from "@mui/material";
+
+interface ButtonContent {
+  label: string;
+  url: string;
+  onClick?: () => void;
+}
+
+export interface OverviewBannerProps {
+  title?: string;
+  body: React.ReactNode;
+  primaryButton?: ButtonContent;
+  secondaryButton?: ButtonContent;
+}
+
+export const OverviewBanner: React.FC<OverviewBannerProps> = ({
+  title,
+  body,
+  primaryButton,
+  secondaryButton,
+}) => {
+  const theme = useTheme();
+
+  return (
+    <Stack spacing={2} p={3} bgcolor={theme.palette.grey[100]} borderRadius={1}>
+      {title && <Typography variant="labelLarge">{title}</Typography>}
+      {body && (
+        <Typography pb={1.5} variant="paragraphLarge">
+          {body}
+        </Typography>
+      )}
+      <Stack direction="row" spacing={2}>
+        {primaryButton && (
+          <Button
+            variant="text"
+            href={primaryButton.url}
+            onClick={primaryButton.onClick}
+          >
+            {primaryButton.label}
+          </Button>
+        )}
+        {secondaryButton && (
+          <Button
+            variant="text"
+            href={secondaryButton.url}
+            onClick={secondaryButton.onClick}
+          >
+            {secondaryButton.label}
+          </Button>
+        )}
+      </Stack>
+    </Stack>
+  );
+};
--- a/packages/ui-components/src/components/OverviewBanner/OverviewBanner.tsx
+++ b/packages/ui-components/src/components/OverviewBanner/OverviewBanner.tsx
@@ -3,8 +3,11 @@ import { Stack, Typography, Button } from "@mui/material";
 import { useTheme } from "@mui/material";
 
 interface ButtonContent {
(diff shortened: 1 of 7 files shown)


packages/ui-components/src/styles/theme/components.tsx
```diff
@@ -33,6 +33,13 @@ const components: ThemeOptions["components"] = {
       outlinedPrimary: ({ theme }) => ({
         borderColor: theme.palette.border.default,
       }),
+      text: {
```
chasulin, line 36: text buttons should sit flush to the side of the container (and any necessary spacing should depend on the parent/wrapper), so we don't get something like this, where padding makes the buttons appear more indented than the surrounding content: 

Screen Shot 2022-10-20 at 1 39 25 PM

packages/ui-components/src/components/OverviewBanner/OverviewBanner.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  return (
+    <Stack spacing={2} p={3} bgcolor={theme.palette.grey[100]} borderRadius={1}>
+      {title && <Typography variant="labelLarge">{title}</Typography>}
+      {body && (
+        <Typography pb={1.5} variant="paragraphLarge">
+          {body}
+        </Typography>
+      )}
+      <Stack direction="row" spacing={2}>
```
pnavarrc, line 37: I think it might be better to keep this simple and have a `children` prop that we can use to put any kind of content in the banner instead of having `primaryButton` and `secondaryButton.

The primary and secondary buttons are fairly restrictive - for example, we can't use a `contained` button or add an icon, or have a button that's not a link.

Doing this will make this component almost too simple to exist, but I think that there is value in having consistent spacing and good defaults for the title, background color, rounded borders, etc.

packages/ui-components/src/components/OverviewBanner/OverviewBanner.tsx
```diff
@@ -0,0 +1,59 @@
...
+export const OverviewBanner: React.FC<OverviewBannerProps> = ({
+  title,
+  body,
+  primaryButton,
+  secondaryButton,
+}) => {
+  const theme = useTheme();
+
+  return (
+    <Stack spacing={2} p={3} bgcolor={theme.palette.grey[100]} borderRadius={1}>
```
pnavarrc, line 30: I think users might want to customize the background color (we changed the color every single time), consider making it a prop

packages/ui-components/src/styles/theme/components.tsx
```diff
@@ -33,6 +33,13 @@ const components: ThemeOptions["components"] = {
...
+        padding: 0,
+        "&: hover": {
+          backgroundColor: "transparent",
+          textDecoration: "underline",
+        },
+      },
```
pnavarrc, line 42: I think this change might be too extensive - we probably want to keep the hover styling for non-link text buttons - which will require keeping the padding too.

I see the problem when aligning horizontally though - but maybe the parent can handle that (add padding to the title and body containers, but not to the button container, etc).

If we change the component API to take `children`, when we can use links instead of buttons, which don't have the padding problem.

pnavarrc (commented):
Added some notes, I think we can simplify the API and keep it more flexible

packages/ui-components/src/components/OverviewBanner/OverviewBanner.tsx
chasulin: hm i think this might be an oversimplification. the `OverviewBanner` is a specific component that's mocked up. what you're describing is really a div with border radius and anything inside

packages/ui-components/src/styles/theme/components.tsx
chasulin: > I see the problem when aligning horizontally though - but maybe the parent can handle that (add padding to the title and body containers, but not to the button container, etc).

this^ complicates/breaks convention of how we add spacing in general (ie. using stack and just setting the `spacing` prop). if kept as is, we'll probably often be making adjustments to accommodate spacing, moving the buttons to sit flush but not moving surrounding components, etc.. also, the hover state will then not align with the content above

chasulin:
closing because this component is probably too simple to be its own permanent thing

PR #319, opened 2022-10-20
Unify metric categories and levels.

### LINKED ISSUE #259: Try to unify levelSet and categories for COMMUNITY_RISK_LEVEL metric.

mikelehen:
* Rename `MetricLevel` => `Category` and remove `MetricCategory`.
* Unify `categoryThresholds` and `categoryValues` as two ways to "categorize" metric values into categories.
* Misc. cleanup, documentation, error improvements.

Note: I don't *love* this terminology.  I think for low/medium/high type grading (that we use for thermometers, charts, etc.), "level" is a clearer name.  But "level" doesn't really make sense for categorical metrics (Yes / No, Urban / Rural, etc.). And so "category" feels like a more general-purpose term to use. 🤷  That said, we have historically used graded-metrics more than categorical metrics, so maybe using `Level` for categorical metrics wouldn't be terrible?  I'm torn. Holler if you have suggestions! 🤷 

Fixes #259


packages/metrics/src/Metric/Metric.ts
```diff
@@ -91,101 +105,136 @@ export class Metric {
...
+
+    // Validate categoryValues.
+    if (this.categoryValues) {
+      assert(
+        this.categorySet,
+        `${this} defines categoryValues but does not specify the categorySetId of the categories to use the values with. Add a categorySetId to your metric definition.`
+      );
+      assert(
+        this.categoryValues.length === this.categorySet.categories.length,
+        `${this} defines ${this.categoryValues.length} values in categoryValues but the referenced ${this.categorySetId} category set has ${this.categorySet.categories.length} categories. There should be the same number of categoryValues as there are categories. Add or remove values as necessary.`
```
pnavarrc, line 144: The error messages are great!

packages/metrics/src/Metric/Metric.ts
```diff
@@ -91,101 +105,136 @@ export class Metric {
...
    *
-   * @param value The value to be graded.
-   * @returns The level that the value falls into.
+   * @param value The value to be categorized.
+   * @returns The category that the value falls into.
    */
-  getLevel(value: unknown): MetricLevel {
+  getCategory(value: unknown): Category {
+    assert(
+      this.categoryThresholds || this.categoryValues,
```
pnavarrc, line 169: I think we can use `this.hasCategories()` here, and maybe rely on the validations for having `categorySet && thresholds` OR `categorySet && categoryValues` if we have categories from the constructor.

packages/metrics/src/Metric/Metric.ts
```diff
@@ -91,101 +105,136 @@ export class Metric {
...
+      `${this} does not have categoryThresholds or categoryValues defined in its MetricDefinition so getCategory() cannot be used with it.`
+    );
```
pnavarrc: I found that the validations here were a bit redundant with the constructor (or we could use `this.hasCategories()` to simplify), but I'm guessing we do that for TS reasons

packages/metrics/src/Metric/Metric.ts
```diff
@@ -91,101 +105,136 @@ export class Metric {
...
+      assert(this.categorySet); // validated in constructor.
+      return this.getCategoryByValues(
+        this.categorySet,
+        this.categoryValues,
+        value
+      );
+    } else {
+      fail(
+        `${this} does not have categoryThresholds or categoryValues defined in its MetricDefinition so getCategory() cannot be used with it.`
+      );
```
pnavarrc, line 185: This can never happen right?

pnavarrc (approved):
Looks good! I think that _category_ is a bit better/general than _level_ - an we can try to use a more neutral language in components moving forward. Thanks for cleaning up!

packages/metrics/src/Metric/Metric.ts
mikelehen: Ooops, good catch!  These are redundant now.  I think I refactored and forgot to clean up.  Thanks!

packages/metrics/src/Metric/Metric.ts
mikelehen: I removed the above assert, so now this can happen. 😁

PR #323, opened 2022-10-24
Updates the MetricLineThresholdChart to show all the levels

pnavarrc:
Previously, the `MetricLineThresholdChart` chart will only show the intervals that intersected with the data. Here, we update how intervals are calculated (and the chart) to make sure that all the _categories_ are visible in the chart.

Note that I added the labels just to illustrate the change, but I don't want to add the labels yet (they are not in the mocks and don't look great, I think we need a better solution).

image

I'm going to update the `calculateChartIntervals` function to take a `metric` on a follow-up PR to keep this one focused on the logic.

### DIFF
--- a/packages/ui-components/src/components/MetricLineThresholdChart/utils.ts
+++ b/packages/ui-components/src/components/MetricLineThresholdChart/utils.ts
@@ -86,6 +86,7 @@ export function calculateChartIntervals(
 
   const firstThreshold = thresholds[0];
   const lastThreshold = thresholds[thresholds.length - 1];
+  const delta = 0.2 * (lastThreshold - firstThreshold);
 
   // Build the intervals in the same order as the categories, and then
   // filter out the intervals that are not relevant for the chart.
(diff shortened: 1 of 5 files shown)


packages/ui-components/src/components/MetricLineThresholdChart/utils.ts
```diff
@@ -87,23 +81,30 @@ export function calculateChartIntervals(
...
-          ? Math.min(minVal, firstThreshold)
-          : thresholds[categoryIndex - 1],
-        upper: isLastCategory
-          ? Math.max(maxVal, lastThreshold)
-          : thresholds[categoryIndex],
-      };
-    })
-    .filter((interval) => interval.lower < interval.upper);
+  // If `minValue` or `maxValue` are inside the thresholds, we calculate
+  // a padding to make sure that all the categories are visible in the
```
mikelehen: This phrasing is a bit confusing, since we always calculate the padding (but then later on may or may not use it).  It might also be helpful to reference the category label, which is really what we're trying to make room for.

Maybe just say something like:
> // Calculate a padding to make sure that each category has room for a label to be rendered inside the category. Here, we use ...

mikelehen (approved):
Minor comment suggestion, but looks good!

PR #324, opened 2022-10-24
Fix the Metric constructor to properly compare categoryThresholds

pnavarrc:
Fixes a small bug on the Metric constructor - In order to determine if the thresholds are sorted, we compare the thresholds to sorted thresholds (both ascending and descending). 

https://github.com/covid-projections/act-now-packages/blob/00a645b7c6e10258464f52af3ed26706b73c389d/packages/metrics/src/Metric/Metric.ts#L126-L133

The problem in the original code is that [`Array.prototype.sort`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) converts numbers to strings before sorting. Passing a comparator function solves the issue. I added more test cases to confirm that it only throws when it should.

### DIFF
--- a/packages/metrics/src/Metric/Metric.test.ts
+++ b/packages/metrics/src/Metric/Metric.test.ts
@@ -7,24 +7,14 @@ import { MetricCatalogOptions } from "../MetricCatalog";
 const testMetricDef: MetricDefinition = {
   id: "cases",
   name: "Cases per 100k",
-  categorySetId: "low-medium-high",
+  categorySetId: "low-medium-high-critical",
 };
 
 // Thresholds consistent with the testCategorySet
-const testMetricThresholds = [10, 20];
+const testMetricThresholds = [10, 20, 30];
 
 // Example of a typical CategorySet.
 const testCategorySet: CategorySet = {
-  id: "low-medium-high",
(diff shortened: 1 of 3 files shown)


packages/metrics/src/Metric/Metric.test.ts
```diff
@@ -24,13 +24,24 @@ const testCategorySet: CategorySet = {
...
+const testCategorySet4: CategorySet = {
+  id: "low-medium-high-critical",
+  defaultCategory: { id: "unknown", color: "grey" },
+  categories: [
+    { id: "low", name: "Low", color: "green" },
+    { id: "medium", name: "Medium", color: "yellow" },
+    { id: "high", name: "High", color: "red" },
+    { id: "critical", name: "Critical", color: "maroon" },
+  ],
+};
```
mikelehen, line 26: Not a big deal, but in the interest of keeping tests simpler, I'd be a fan of just adding a "critical" category to the existing `testCategorySet` rather than creating a new set that's used in some tests but not others.  You might have to tweak a few other tests though, not sure.

mikelehen (approved):
Looks good with optional test simplification suggestion.

PR #327, opened 2022-10-24
Rework metric hooks using SWR, and update tests.

### LINKED ISSUE #315: Revisit behavior of asynchronous `useMetricData()` hooks.
In https://github.com/covid-projections/act-now-packages/pull/314 I added a `hasMetricData()` check before calling `metricData()` in the `getColor()` functions of the maps.  This fixes a problem @faihegberg ran into with the `MetricMiniMap` where switching metrics would crash with:

> INTERNAL ASSERTION FAILED: No data for metric mock_cases. Did you forget to specify it when you created the MetricDataStore?

The problem is that `useMetricData()` hooks update asynchronously when you change what metrics you're passing to them and so there's a period of time where you're still getting the old data back and so the data won't have the metric you are asking for.

We had a similar problem during the hackathon that I fixed with https://github.com/covid-projections/act-now-packages/pull/258 but I'm thinking that wasn't a complete enough solution. :/

I'm not sure what the best solution is.  Maybe either:
1. Change `MultiMetricDataStore` and `MultiRegionMultiMetricDataStore` to return "empty" data objects if you ask for a nonexistent region / metric.  I started to do that, but it's nontrivial because in order to create an empty `MetricData` object you need the `Metric` being asked for, but we may only have the metric ID.  To solve that, `MultiMetricDataStore` would need to have a reference to the `MetricCatalog` so it can look up the metric.  That would create a circular reference, which we'd need to either be okay with or break by introducing an interface or something.
2. Reworking the `useMetricData()` hooks to return an empty `DataOrError` object when you change what regions/metrics you're looking for data for.  I am not sure if that's straightforward to do though.  I guess you'd use `useRef()` to track the previously requested regions / metrics or something.  And it might cause flicker when switching between metrics (since we revert to a "loading" state momentarily).

mikelehen:
* Reworked metric data hooks to use `useSWR` hook from https://swr.vercel.app/.  This was surprisingly smooth, and it fixes #315 (the hook will now return to the loading state (data==undefined) when you change the metrics / regions you're asking for).
* Reworked the tests to use `@testing-library/react` instead of `@testing-library/react-hooks` as the latter was triggering a  is soon going to be deprecated and for some reason with SWR was triggering a "A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown." warning.
* minor cleanup


jest.config.js
```diff
@@ -1,7 +1,7 @@
 /** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
 module.exports = {
   preset: "ts-jest",
-  testEnvironment: "node",
+  testEnvironment: "jsdom",
```
mikelehen, line 4: I had to do this after switching to `@testing-library/react`.  The tests still pass, and I think `jsdom` probably makes sense for testing web components anyway, so hopefully it's fine? 🤷

packages/ui-components/package.json
```diff
@@ -39,7 +39,7 @@
     "@storybook/manager-webpack4": "^6.5.12",
     "@storybook/react": "^6.5.12",
     "@storybook/testing-library": "^0.0.13",
-    "@testing-library/react-hooks": "^8.0.1",
+    "@testing-library/react": "^13.4.0",
```
mikelehen, line 42: I switched because react-hooks is [going to be deprecated](https://github.com/testing-library/react-hooks-testing-library/issues/849) and this solved a "A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown" warning I was getting after I started using SWR.

packages/ui-components/src/common/hooks/metric-data.test.tsx
```diff
@@ -53,10 +49,7 @@ describe("metric data hooks", () => {
     expect(catalog.dataFetchesCount).toBe(expectedFetches);
 
     // Render the hook initially.
-    const { result, waitForNextUpdate, rerender } = renderHook<
-      any,
-      DataOrError<MultiRegionMultiMetricDataStore>
-    >(
```
mikelehen, line 59: There was something messed up with the `@testing-library/react-hooks` types that forced me to manually specify the generic types and use `any`.  But with `@testing-library/react` this problem went away.

packages/ui-components/src/common/hooks/metric-data.test.tsx
```diff
@@ -75,9 +68,11 @@ describe("metric data hooks", () => {
     // The catalog should have performed a data fetch, but since it's anync,
     // initially the hook won't return data.
     expect(catalog.dataFetchesCount).toBe(++expectedFetches);
-    expect(result.current.data).toStrictEqual(undefined);
+    expect(result.current.data).toBe(undefined);
     // Wait for initial async fetch to finish.
-    await waitForNextUpdate();
+    await waitFor(() => {
+      expect(result.current.data).not.toBe(undefined);
+    });
```
mikelehen, line 75: `@testing-library/react` got rid of `waitForNextUpdate()` and this is the new way to do it, I guess.  I don't love it, but it's what's in the [migration guide](https://github.com/testing-library/react-hooks-testing-library/blob/chore/migration-guide/MIGRATION_GUIDE.md).

packages/ui-components/src/common/hooks/metric-data.test.tsx
```diff
@@ -93,8 +88,13 @@ describe("metric data hooks", () => {
     // Render the hook again with different regions / metrics.  This should perform a fetch.
     rerender({ regions: [testRegionWA], metrics: [MetricId.E] });
     expect(catalog.dataFetchesCount).toBe(++expectedFetches);
+    // Data should go back to loading state.
+    expect(result.current.data).toBe(undefined);
```
mikelehen: This expectation is new, and it would fail before we switched to SWR.  Now, when you request different regions / metrics, we temporarily revert to the loading state.

packages/ui-components/src/common/hooks/metric-data.test.tsx
```diff
@@ -75,9 +68,11 @@ describe("metric data hooks", () => {
     // The catalog should have performed a data fetch, but since it's anync,
     // initially the hook won't return data.
     expect(catalog.dataFetchesCount).toBe(++expectedFetches);
-    expect(result.current.data).toStrictEqual(undefined);
+    expect(result.current.data).toBe(undefined);
```
mikelehen: Drive-by change for consistency. 🤷

packages/ui-components/src/common/hooks/metric-data.ts
```diff
@@ -54,21 +54,13 @@ export function useDataForMetrics(
   includeTimeseries = false
 ): DataOrError<MultiMetricDataStore> {
   const catalog = useMetricCatalog();
-  let resolvedMetrics = metrics.map((m) => catalog.getMetric(m));
+  const args = { region, metrics, includeTimeseries };
 
-  // In order to allow people to pass in a new array of metrics that
-  // contain the same metrics as before without triggering additional
-  // fetches, we need this caching trick.
```
mikelehen, line 61: Note that I think SWR is going to serialize the list of metrics / regions in order to check if they changed between calls, which is slower than what I was doing here, but hopefully it's fine.  (I'm mostly just nervous about when we do `regions.all` for fetching data for rendering the map / compare table, etc.  Serializing the whole list of regions isn't ideal.)

jest.config.js
pnavarrc: yeah it should be fine :)

packages/ui-components/src/common/hooks/metric-data.test.tsx
pnavarrc: Alternatively, we could use [`toBeUndefined()`](https://jestjs.io/docs/expect#tobeundefined`) and `not.toBeUndefined()` (the Jest docs say it's better practice, but they don't say why)

> You could write `expect(bestDrinkForFlavor('octopus')).toBe(undefined)`, but it's better practice to avoid referring to undefined directly in your code.

pnavarrc (approved):
Looks good! Optionally we could use `.toBeUndefined()` for the tests, but I'm not sure if that's actually better.

packages/ui-components/src/common/hooks/metric-data.test.tsx
mikelehen: Ah!  Interesting.  That's probably a tad cleaner.  And there's a `toBeDefined()` I can use instead of `not.toBeUndefined()` which I definitely like better.  Updated!

FWIW- In the olden days of JavaScript, when you referenced `undefined` in your code, you were actually just referencing a global variable called `undefined` that had the value `undefined` and if you accidentally wrote code like this:

```
var globalToUpdate;
window[globalToUpdate] = "hello";
```

You would accidentally overwrite the `undefined` variable and so now whenever you referred to `undefined` in your code, you'd end up with `"hello"` instead of `undefined`.

And so the "correct" way to check for undefined was `typeof value === "undefined"` instead of `value === undefined`.

But at some point browsers realized that was terrible and started prevented you from overwriting `undefined`. So `window["test"] = "hello"` will create a global called `test`, but `window["undefined"] = "hello"` is a no-op. 😁 

I'm guessing that's part of why jest has those docs and it's not as relevant now.  But `toBeUndefined()` is pretty clean, so I'll go ahead and use it!

PR #328, opened 2022-10-24
Add links to regions in MetricCompareTable

### LINKED ISSUE #277: Link to region pages on the Metric compare table (using MuiLink)
Every cell should be a link to the corresponding region page - Ideally we should highlight the entire row

pnavarrc:
Close #277 - Add links to `MetricCompareTable` rows

image

## Changes

- Updated the styles to highlight rows on hover (default hover overlay, if the table cell has a link inside, the text will have `primary.dark` color on hover)
- Add links to regions on the `MetricCompareTable` (this required restyling the table cells to remove the padding. It also required passing the `RegionDB` to the `MetricCompareTable` (we only need the function to get the URL, but this way the API is consistent with the metric-aware maps)

### DIFF
--- a/.changeset/thirty-kiwis-approve.md
+++ b/.changeset/thirty-kiwis-approve.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/ui-components": minor
+---
+
+Add links to region pages on `MetricCompareTable`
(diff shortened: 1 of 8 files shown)


mikelehen (approved):
Looks good to me!

PR #329, opened 2022-10-24
Fix #316: Disallow null and undefined in timeseries.

### LINKED ISSUE #316: Disallow nulls/undefineds in timeseries.
We discussed this at standup a while ago, but I forgot to open an issue.  We don't think there's a good use case for allowing null/undefined within a Timeseries.  So we should make sure the constructor disallows it and then get rid of the `removeNils()` function.

mikelehen:
Assert in constructor and remove no longer needed `remvoeNils()` method.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,5 +28,5 @@
     "@typescript-eslint/no-explicit-any": "off",
     "lodash/import-scope": "warn"
   },
-  "ignorePatterns": ["!.storybook", "**/dist", "**/lib"]
+  "ignorePatterns": ["!.storybook", "**/dist", "**/lib", "public/**"]
 }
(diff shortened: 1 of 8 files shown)


.eslintrc.json
```diff
@@ -28,5 +28,5 @@
     "@typescript-eslint/no-explicit-any": "off",
     "lodash/import-scope": "warn"
   },
-  "ignorePatterns": ["!.storybook", "**/dist", "**/lib"]
+  "ignorePatterns": ["!.storybook", "**/dist", "**/lib", "public/**"]
```
mikelehen, line 31: Drive-by fix so that `yarn lint` doesn't fail after you've run `yarn generate-docs`

pnavarrc (approved):
Looks good!

PR #330, opened 2022-10-25
Fix #322: Require passing provider ids to all metric data providers.

### LINKED ISSUE #322: Require passing providerId to all metric data providers during construction.
This came up during the hackathon when it wasn't obvious what providerId to use with the CovidActNowDataProvider.  We decided that perhaps you should pass the providerId in when constructing the provider.  That way there's an obvious correlation in your own code.

mikelehen:
Rather than having hardcoded provider ids for some data providers ("mock" for MockDataProvider, "static" for StaticValueDataProvider, etc.), we now require the user to always pass in their own provider IDs.  That way it's a bit more explicit and we don't have to document the IDs for the different providers.

Most of the changes in the PR are just updating tests / stories to use const / enums for provider ids instead of the hardcoded values.

### DIFF
--- a/.changeset/chatty-frogs-wash.md
+++ b/.changeset/chatty-frogs-wash.md
@@ -1,0 +1,6 @@
+---
+"@actnowcoalition/metrics": patch
+"@actnowcoalition/ui-components": patch
+---
+
+Fix #322: Require passing provider ids to all metric data providers.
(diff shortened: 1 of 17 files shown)


smcclure17 (approved):
looks good! thanks for plumbing all these changes through 🙌

mikelehen:
Fixes #322

PR #338, opened 2022-10-25
Update `RegionSearch` API to use `MuiLink` and other improvements

### LINKED ISSUE #333: Update RegionSearch to use MuiLink instead of receiving a LinkComponent
The type of link is customized at the theme level now, so we should be able to remove the `LinkComponent` prop and just use `MuiLink` internally.

We could also remove `getRegionUrl` and pass the `RegionDB` instance to be consistent with the way the compare table and metric-aware maps work.

pnavarrc:
Close https://github.com/covid-projections/act-now-packages/issues/333 

image

## Changes

- Update the API of `RegionSearch` to use `MuiLink` and `regionDB` instead of the props `LinkComponent` and `getRegionUrl`
- Update the styles of `MuiAutocomplete` to remove the margins of `MuiTextField`, which were causing the component to be wider than its container. 

**Before** (I added a container with a fixed width and a black border to show the difference)

image

**After**

image

### DIFF
--- a/packages/ui-components/src/styles/theme/components.tsx
+++ b/packages/ui-components/src/styles/theme/components.tsx
@@ -1,6 +1,6 @@
 /** MUI theme components */
 import React from "react";
-import { ThemeOptions, createTheme, textFieldClasses } from "@mui/material";
+import { ThemeOptions, createTheme } from "@mui/material";
 import KeyboardArrowDownIcon from "@mui/icons-material/KeyboardArrowDown";
 
 const referenceTheme = createTheme();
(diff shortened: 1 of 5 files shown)


packages/ui-components/src/styles/theme/components.tsx
```diff
@@ -12,6 +12,13 @@ const components: ThemeOptions["components"] = {
       disableClearable: true,
       popupIcon: <KeyboardArrowDownIcon />,
     },
+    styleOverrides: {
+      root: {
+        [`& .${textFieldClasses.root}`]: {
+          margin: 0,
```
mikelehen: FWIW- I don't really understand what's going on / why this is necessary.  Consider adding a comment if there's a straightforward explanation.

mikelehen (approved):
Looks good!

packages/ui-components/src/styles/theme/components.tsx
pnavarrc: The `MuiTextField` component had some margins, so I wanted to counteract that - but actually, I removed the extra margin and the components look the same 🤷‍♂️  (I checked all the ones that should have been affected).

PR #340, opened 2022-10-26
Misc. small fixes

chasulin:
- Renames `currentRegion` prop -> `highlightedRegion`
- Updates type of `renderTooltip` prop to return `React.ReactNode` instead of `string | React.ReactElement`


mikelehen (approved):
Can you add a changeset?  Otherwise, looks good!  Thanks for following up on these!

PR #344, opened 2022-10-26
Clean up fetch code and use p-retry for retries.

mikelehen:


smcclure17 (approved):
Looks great! 🙌 🚢 

I really like how you've implemented `fetchData()` with the `responseHandler` param--that feels very clean 🤩

Issue #241, opened 2022-09-28
Make sure CovidActNowDataProvider doesn't fetch the same API endpoint multiple times.


mikelehen:
I believe right now if you do something like:

metricCatalog.fetchData(region, metric1)
metricCatalog.fetchData(region, metric2)

We will end up fetching data for the region twice (you'll see two network requests in the browser).

We will need to rework the code so that we can cache the fetch `Promise`, something akin to https://github.com/covid-projections/act-now-packages/pull/231

mikelehen:
See also #249 maybe we can find a networking library that helps us solve both.

PR #348, opened 2022-10-27
Fix #249: Use p-limit for throttling requests more elegantly.

### LINKED ISSUE #249: Improve parallelization and retries in CovidActNowDataProvider fetching.
I'm implementing some very crude parallelization / retries, but we should clean up the code so it's reusable or better yet, find an existing library that can do this for us.  Maybe it could fix the caching issue in #241 as well.

mikelehen:
Switch to using the "p-limit" npm module for parallelizing requests to the CAN API, which I think is a bit more elegant and slightly more efficient (we don't wait for all 50 to complete before starting the next batch of 50).

Tested via running `yarn generate-data-snapshot` in hackathon repo.

Fixes #249.

### DIFF
--- a/.changeset/hot-moons-battle.md
+++ b/.changeset/hot-moons-battle.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/metrics": patch
+---
+
+Use p-limit for throttling CAN API requests more elegantly.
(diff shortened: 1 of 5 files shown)


pnavarrc (approved):
Looks good! 

Not necessarily related, but I noticed that in the hackathon repo (location page), we trigger 6 requests to the same endpoint (I imagine that they are triggered around the same time, while waiting for the response of the first request)

image

Maybe we could mark `apiJson[cacheKey]` as loading or something once we trigger the request (instead of once we get the response) to avoid that? (not sure how we could handle failures though)

mikelehen:
Thanks!  That's part of my next PR to rework caching and fix #241 . 😁

PR #350, opened 2022-10-27
WorldMap and MetricWorldMap components

### LINKED ISSUE #58: World map (WHO)

chasulin:
- Adds WorldMap + MetricWorldMap components
- [Disables typedoc for the time being](https://github.com/covid-projections/act-now-packages/pull/350/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R31) (removes `yarn generate-docs` from the build script), because the generation of docs is causing a "javascript heap out of memory" error that we can't get to the bottom of at the moment ([TODO to investigate further](https://github.com/orgs/covid-projections/projects/6/views/4))

closes #58


packages/ui-components/src/components/WorldMap/WorldMap.tsx
```diff
@@ -0,0 +1,127 @@
...
+        {/* Clickable region overlay */}
+        {countries.features.map((geo) => (
+          <>
+            <Tooltip title={renderTooltip(`${geo.id}`) ?? ""}>
+              <Link href={getRegionUrl(`${geo.id}`)}>
+                <RegionOverlay d={geoPath(geo) ?? ""} />
+              </Link>
+            </Tooltip>
+          </>
+        ))}
```
pnavarrc, line 70: Obviously from before, but we could attach the tooltips to the features above instead of duplicating every shape here. Not sure why we did it this way 🤷

pnavarrc (approved):
Looks good!

packages/ui-components/src/components/WorldMap/WorldMap.tsx
chasulin: [the above shapes](https://github.com/covid-projections/act-now-packages/pull/350/files#diff-16a9774defe809c0484d4dbb458186fa6f1a04d21988315ba54bb4c3a4e3adacR50-R59) can receive a fill color. these shapes are fully transparent and become a partially transparent white when hovered, so the colored shape below has a white overlay when hovered

chasulin: and the tooltip is attached to the shape that sits on top

PR #351, opened 2022-10-27
Implement `LineIntervalChart` component

pnavarrc:
Implements the `LineThresholdChart`, the non-metric aware version of `MetricLineThresholdChart`. While implementing #341, I realized that the dependency between metrics and the coloured line is fairly weak, and that we can use the non-metric aware version for both `MetricLineThresholdChart` and for the `SeriesChart`.

I also implemented the `createTimeseriesScales` function to simplify the stories that need scales based on `timeseries.

### DIFF
--- a/packages/ui-components/src/components/LineChart/LineChart.stories.tsx
+++ b/packages/ui-components/src/components/LineChart/LineChart.stories.tsx
@@ -1,8 +1,9 @@
 import React from "react";
 import { ComponentStory, ComponentMeta } from "@storybook/react";
-import { scaleLinear, scaleTime } from "@visx/scale";
-import { assert } from "@actnowcoalition/assert";
-import { appleStockTimeseries } from "../../stories/mockData";
+import {
+  appleStockTimeseries as timeseries,
+  createTimeseriesScales,
+} from "../../stories/mockData";
 import { LineChart } from ".";
 
 export default {
@@ -11,7 +12,6 @@ export default {
 } as ComponentMeta<typeof LineChart>;
 
 const [width, height] = [600, 400];
-const padding = 20;
 
 const Template: ComponentStory<typeof LineChart> = (args) => (
   <svg width={width} height={height} style={{ border: "solid 1px #eee" }}>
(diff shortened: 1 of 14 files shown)


packages/ui-components/src/components/LineChart/LineChart.stories.tsx
```diff
@@ -11,39 +12,26 @@ export default {
...
-const xScale = scaleTime({
-  domain: [minDate, maxDate],
-  range: [padding, width - padding],
-});
-
-const yScale = scaleLinear({
-  domain: [minValue, maxValue],
-  range: [height - 2 * padding, padding],
-});
+const { xScale, yScale } = createTimeseriesScales(timeseries, width, height);
```
mikelehen, line 22: Nice!

packages/ui-components/src/components/LineThresholdChart/LineThresholdChart.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+export interface LineThresholdChartProps {
+  /**
+   * Array defining the intervals on the y-axis on which the line will
+   * be rendered with the color specified in the interval.
+   */
+  intervals: LineThresholdInterval[];
+  /** Timeseries to render as a line */
+  timeseries: Timeseries<number>;
+  /** d3-scale to transform point dates to pixel positions on the x-axis */
```
mikelehen: Should this be y-axis?

packages/ui-components/src/components/LineThresholdChart/LineThresholdChart.tsx
```diff
@@ -0,0 +1,101 @@
...
+  yScale: ScaleLinear<number, number>;
+  /** d3-scale to transform point values to pixel positions on the y-axis */
```
mikelehen: x-axis?

packages/ui-components/src/components/LineThresholdChart/LineThresholdChart.tsx
```diff
@@ -0,0 +1,101 @@
...
+ * between 10 and 20, and in red where its y-coordinates are between 20 and 40.
+ * If the timeseries has values outside these intervals, those line segments
+ * won't be rendered.
+ *
+ * @param timeseries Timeseries to represent as a line.
+ * @param xScale d3-scale to transform point dates to pixel positions on the x-axis.
+ * @param yScale d3-scale to transform point values to pixel positions on the y-axis.
+ * @param intervals intervals that define the color of the line for different segments.
+ *
+ * @returns An SVG group containing the line.
```
mikelehen: Note that nothing in the description mentions "thresholds" which suggests we might want to rename this component.  LineIntervalChart or LineColoredIntervalsChart or something?  Not sure.

mikelehen (commented):
Couple doc nits, but otherwise looks good!

packages/ui-components/src/components/LineThresholdChart/LineThresholdChart.tsx
pnavarrc: Good call! That's probably better, renaming!

PR #352, opened 2022-10-31
Fix CovidActNowDataProvider caching (#241) and rework caching in general.

### LINKED ISSUE #241: Make sure CovidActNowDataProvider doesn't fetch the same API endpoint multiple times.

mikelehen:
It turns out that CachingMetricDataProviderBase wasn't that helpful for implementing a MetricDataProvider with caching. It kind of works for a provider that just has one big asset to cache (like the CsvDataProvider, though we messed it up on the first try). But it does not work well when you have more granular assets to cache (e.g. one per region like with CAN).  So I have:

* Replaced CachingMetricDataProviderBase with SimpleMetricDataProviderBase that helps you write a provider that fetches a single region and metric at a time but doesn't pretend to help with caching.
* Reworked the CovidActNowDataProvider to just implement MetricDataProvider directly and correctly cache data on a per-region basis, and avoid having multiple outstanding requests for the same region file.
* Minor cleanup of CsvDataProvider to simplify the caching implementation.

Fixes #241.


packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
```diff
@@ -77,6 +77,35 @@ export class MultiRegionMultiMetricDataStore<T = unknown> {
     return new MultiRegionMultiMetricDataStore(data);
   }
 
+  /**
+   * Static async constructor to create a MultiRegionMultiMetricDataStore from a list
+   * of regions, a list of metrics, and a dataFetcher callback that returns
+   * Promise<MetricData> for each region and metric combination.
```
mikelehen, line 83: This is basically identical to the above `fromRegionsAndMetrics()` static constructor, except this one is Promise-based (your dataFetcher to get data for each region+metric can be asynchronous).

packages/metrics/src/data_providers/CovidActNowDataProvider.ts
```diff
@@ -1,5 +1,4 @@
 import pLimit from "p-limit";
-import { CachingMetricDataProviderBase } from "./CachingMetricDataProviderBase";
```
mikelehen, line 2: We now implement MetricDataProvider directly.

packages/metrics/src/data_providers/CovidActNowDataProvider.ts
```diff
@@ -22,66 +24,86 @@ const MAX_CONCURRENT_REQUESTS = 50;
...
  * - For the CAN Community Level, set metric.dataReference.column to `communityLevels.canCommunityLevel`.
  */
-export class CovidActNowDataProvider extends CachingMetricDataProviderBase {
+export class CovidActNowDataProvider implements MetricDataProvider {
   /** Valid Covid Act Now API key to use in API calls. */
   private readonly apiKey: string;
 
   /** Cached CAN API responses indexed by the FIPS codes of the regions fetched. */
-  private apiJson: { [regionId: string]: DataRow };
+  private apiJson: { [regionId: string]: Promise<DataRow> };
```
mikelehen, line 32: This is the core of the fix for #241.  We now cache the promise for fetching each region from the API so that subsequent calls to the data provider will wait on the existing promise instead of potentially starting another request to the API.

packages/metrics/src/data_providers/CovidActNowDataProvider.ts
```diff
@@ -22,66 +24,86 @@ const MAX_CONCURRENT_REQUESTS = 50;
...
-   * @param providerId A unique provider id to associate with the provider (e.g.
+   * @param id A unique provider id to associate with the provider (e.g.
    * "can-api"). This ID can be used from a {@link MetricDataReference} in a
    * metric to reference the data from this provider.
    * @param apiKey Valid Covid Act Now API key to use in API calls.
    * @param data JSON data to put directly into the cache. Used primarily for testing.
    */
   constructor(
-    providerId: string,
+    readonly id: string,
```
mikelehen, line 44: We have to implement the `id` property directly since we're implementing `MetricDataProvider` ourselves now.

packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -67,14 +67,10 @@ describe("CsvDataProvider", () => {
     expect(metricData.hasTimeseries()).toBe(false);
   });
 
-  test("populateCache() fails if csv does not have at least one row.", async () => {
-    const provider = new CsvDataProvider("PROVIDER_ID", {
-      regionColumn: "region",
-      csvText: `region,cool_metric`,
-    });
-    expect(async () => {
-      await provider.populateCache();
```
mikelehen, line 76: populateCache() is no longer a publicly accessible method, so I reworked this to use the public API.

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -37,12 +38,11 @@ export interface CsvDataProviderOptions {
...
  * ```
  */
-export class CsvDataProvider extends CachingMetricDataProviderBase {
+export class CsvDataProvider extends SimpleMetricDataProviderBase {
   private readonly regionColumn: string;
   private readonly dateColumn?: string;
   private readonly url?: string;
-  private readonly csvText?: string;
-  private fetchedTextPromise: Promise<string> | undefined;
+  private fetchedText: Promise<string> | undefined;
```
mikelehen, line 45: `csvText` and `fetchedTextPromise` were a bit redundant so I simplified to a single `fetchedText` property (which is a promise).

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -87,11 +89,15 @@ export class CsvDataProvider extends CachingMetricDataProviderBase {
     this.dataRowsByRegionId = dataRowsByRegionId;
   }
 
-  getDataFromCache(
+  async fetchDataForRegionAndMetric(
```
mikelehen, line 92: This is the new method we need to override from `SimpleMetricDataProviderBase` (and it's now async, returning a promise).

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -37,12 +38,11 @@ export interface CsvDataProviderOptions {
  * |CA     |2022-02-01 |31   |66   |
  * ```
  */
-export class CsvDataProvider extends CachingMetricDataProviderBase {
+export class CsvDataProvider extends SimpleMetricDataProviderBase {
```
mikelehen, line 41: I got rid of `CachingMetricDataProviderBase` since it wasn't that helpful and replaced it with `SimpleMetricDataProviderBase`.

packages/metrics/src/data_providers/MockDataProvider.test.ts
```diff
@@ -19,7 +19,10 @@ describe("MockDataProvider", () => {
         endDate: "2022-01-05",
       },
     });
-    const data = dataProvider.getDataFromCache(testRegion, metric);
+    const data = await dataProvider.fetchDataForRegionAndMetric(
+      testRegion,
+      metric
+    );
```
mikelehen, line 25: Had to rework all these tests to use the public API of `SimpleMetricDataProviderBase` now.

packages/metrics/src/data_providers/SimpleMetricDataProviderBase.ts
```diff
@@ -0,0 +1,48 @@
...
+import { Metric } from "../Metric";
+import { MetricData, MultiRegionMultiMetricDataStore } from "../data";
+
+/**
+ * Base class to help implement a MetricDataProvider that fetches one
+ * region and metric at a time. To support fetching multiple regions
+ * and metrics simultaneously, implement MetricDataProvider directly.
+ */
+export abstract class SimpleMetricDataProviderBase
+  implements MetricDataProvider
```
mikelehen, line 13: This class is pretty similar to the deleted `CachingMetricDataProviderBase` but it doesn't have a `populateCache()` method and `getDataFromCache()` is now `fetchDataForRegionAndMetric()` and returns a Promise.

pnavarrc (approved):
Looks good!

PR #353, opened 2022-10-31
Implement `MetricSeries` chart

pnavarrc:
The MetricSeries component allows to show multiple time series in the same chart. This is only appropriate for time series that share the units on the y-axis (for % of vaccinations, or for normalized metrics such as weekly cases per 100k). A Series object has a type (LINE or BAR), the metric, region and optional props to customize the underlying chart.

**Example** If we want to show a single metric as both a bar chart and a line chart, we can do:

```tsx
const series = [
  {
    region: states.findByRegionIdStrict("53"),
    metric: metricCatalog.getMetric(MetricId.MOCK_CASES),
    type: SeriesType.BAR,
  },
  {
    region: states.findByRegionIdStrict("53"),
    metric: metricCatalog.getMetric(MetricId.MOCK_CASES),
    type: SeriesType.LINE,
    lineProps: {
      stroke: "black",
    },
  },
];

// Render the chart
<MetricSeriesChart
  series={series}
  width={width}
  height={height}
/>
```

More examples in storybook.

**Vaccination** 

image

**Trends Multiple Locations**

image

**Trends Single Locations**

image

### DIFF
--- a/packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
+++ b/packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
@@ -92,7 +92,7 @@ export const MetricSeriesChart = ({
   });
 
   return (
-    <svg width={width} height={height} style={{ border: "solid 1px #ddd" }}>
+    <svg width={width} height={height}>
       <Group top={marginTop} left={marginLeft}>
         <AxesTimeseries
           yScale={yScale}
--- a/packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
+++ b/packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
@@ -5,6 +5,7 @@ import { scaleLinear, scaleTime } from "@visx/scale";
 import uniq from "lodash/uniq";
 import min from "lodash/min";
 import max from "lodash/max";
+import isNumber from "lodash/isNumber";
 
 import { assert } from "@actnowcoalition/assert";
 import { Timeseries } from "@actnowcoalition/metrics";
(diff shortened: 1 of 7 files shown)


packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
```diff
@@ -0,0 +1,187 @@
...
+    }))
+    .filter(({ timeseries }) => timeseries.hasData());
+
+  const [minDate, maxDate] = getDateRange(timeseriesList);
+  const [minDataValue, maxValue] = getValueRange(timeseriesList);
+
+  // Note: We use minValue if provided. If not, we default to start from zero,
+  // which is best in most cases, unless minDataValue is negative, in which
+  // case we use that value instead.
+  const minYValue = isNumber(minValue) ? minValue : Math.min(0, minDataValue);
```
faihegberg, line 85: Perhaps include stories with negative min values?

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
```diff
@@ -0,0 +1,187 @@
...
+ * @param timeseriesList List of timeseries.
+ * @returns [minValue, maxValue]
+ */
+function getValueRange(timeseriesList: Timeseries<number>[]): [number, number] {
+  const minValue = min(timeseriesList.map(({ minValue }) => minValue));
+  const maxValue = max(timeseriesList.map(({ maxValue }) => maxValue));
+
+  assert(
+    typeof minValue === "number" && typeof maxValue === "number",
+    "min and max value"
```
faihegberg: Would probably want revise this to be a clearer error statement instead of just "min and max value".

faihegberg (approved):

PR #355, opened 2022-11-01
Fix #242: Drop timeseries data if it wasn't asked for.

### LINKED ISSUE #242: Make MetricCatalog drop fetched timeseries data if it wasn't requested.
I ran into an issue when I set up the metric data caching with snapshots where the mock metric data provider was always returning timeseries data which then meant that I was storing timeseries data in the snapshot even though I didn't want / need it, leading to a giant snapshot.

I think we should update MetricCatalog.fetchDataForRegionsAndMetrics() to automatically drop any timeseries data that was returned from the data provider so that the user always gets what they asked for.

mikelehen:
Fixes #242.

### DIFF
--- a/.changeset/friendly-mugs-repair.md
+++ b/.changeset/friendly-mugs-repair.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/metrics": patch
+---
+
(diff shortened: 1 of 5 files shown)


smcclure17 (approved):
lgtm!

Issue #368, opened 2022-11-07
Some components assume a light background (Sparklines, CompareTable, RegionSearch) and their visual attributes are not taken from the theme


chasulin:

chasulin:
Partial solution: https://github.com/covid-projections/act-now-packages/pull/372

PR #372, opened 2022-11-07
Theme defaults

chasulin:
Partial solution for #368 

During the hackathon, we learned that some elements (chart lines, text, axes) assumed white backgrounds in their styling and were hard to configure for dark mode.

This PR [adds 2 variables to our palette module](https://github.com/covid-projections/act-now-packages/pull/372/files#diff-e8db90eaf056409aa60f99ae7e58bc599864cd6e284b72fcade7570c4697b1f8R72-R75), `chart.main` (applied to lines in line charts and bars in bar charts) and `chart.axis` (applied to chart axes) so that these color defaults can be controlled at the theme level in a single location. This PR also changes the typography object to reference `palette.text.{primary/secondary}`, so these defaults also can be controlled at the theme level in a single location.

Still to do:
`palette.text` only allows for `primary` and `secondary`, which is incompatible with our design system having 3 text colors. We need to edit the theme module for the text property to allow a third option

### DIFF
--- a/packages/ui-components/src/components/Axis/Axis.style.tsx
+++ b/packages/ui-components/src/components/Axis/Axis.style.tsx
@@ -31,7 +31,7 @@ export const AxisLeft = styled((props: AxisLeftProps) => (
       dx: "-.25em",
       ...baseTickLabelProps,
     })}
-    stroke={theme.palette.border.default}
+    stroke={theme.palette.chart.axis}
     {...props}
   />
 ))``;
@@ -46,7 +46,7 @@ export const AxisBottom = styled((props: AxisBottomProps) => (
       verticalAnchor: "start",
       ...baseTickLabelProps,
     })}
-    stroke={theme.palette.border.default}
+    stroke={theme.palette.chart.axis}
     {...props}
   />
 ))``;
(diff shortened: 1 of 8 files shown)


pnavarrc:
@chasulin remember to add the changeset! I'm reviewing now!

packages/ui-components/src/components/Axis/Axis.style.tsx
```diff
@@ -31,7 +31,7 @@ export const AxisLeft = styled((props: AxisLeftProps) => (
       dx: "-.25em",
       ...baseTickLabelProps,
     })}
-    stroke={theme.palette.border.default}
+    stroke={theme.palette.chart.axis}
```
pnavarrc, line 34: I think the tick labels should have the same color as the axis lines right? (I made `chart.axis` blue here). Alternatively we could use `typography.paragraphSmall.color` for the axis line stroke color to ensure that the tick labels and lines match.

image

pnavarrc (approved):
Looks good, remember to add the changeset file!

packages/ui-components/src/components/Axis/Axis.style.tsx
chasulin: will do this in a follow up

PR #382, opened 2022-11-08
Theme defaults cont. 

chasulin:
Continuation to PR #372 (re: issue #368)

Connects coloring of spark line to theme (`palette.chart.main`)

Adds `axisLabel` to palette and implements in `Axis.style.tsx`


packages/ui-components/src/components/SparkLine/SparkLine.tsx
```diff
@@ -65,7 +63,7 @@ export const SparkLine: React.FC<SparkLineProps> = ({
             timeseries={timeseriesBarChart}
             xScale={xScaleBar}
             yScale={yScaleBar}
-            fill={theme.palette.border.default}
+            fillOpacity={0.3}
```
chasulin, line 73: removing `fill` and adding `fillOpacity` allows the fill color [to be set at the BarChart level](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/components/BarChart/BarChart.tsx#L73), which is linked to the theme (palette.chart.main)

packages/ui-components/src/styles/theme/palette.ts
```diff
@@ -80,6 +80,7 @@ const palette = {
   chart: {
     main: colors.common.black,
     axis: colors.border.default,
+    axisLabel: colors.text.deemphasized,
```
chasulin, line 84: added `axisLabel` rather than just tying the label coloring to `axis`, in case the axis lines and axis labels should ever be different colors

packages/ui-components/src/components/SparkLine/SparkLine.tsx
mikelehen: I wonder if we'd ever want to customize this via the theme and if there's still a way to do that?  I guess we can cross that bridge if we come to it though.  Not a big deal.

mikelehen (approved):
Looks reasonable to me!

PR #388, opened 2022-11-09
Accommodate negative values in charts

### LINKED ISSUE #362: Bugs - MetricThresholdLineChart
https://www.dropbox.com/scl/fi/ri48yepzt0462ceno0okn/2022-Q4-Hackathon-Notes.paper?dl=0&rlkey=stj3wqol8iipokdngogl6zkpm

- 🐞 Doesn’t work with negative values
- 🐞 Doesn’t crop the chart content to the charting area (the line overlaps the x-axis)
- 🐞 Should have more padding above the maxValue (Note Pablo: The RectClipGroup should allow for some space for the line width on the top most interval so the line is not cut - but the line needs to be cut like that on other intervals)

faihegberg:
Enable our chart to accommodate negative values and address the lack of padding issue at the top of the chart.

Closes https://github.com/covid-projections/act-now-packages/issues/362

This PR does the following:
- Add a new test data set (i.e. NYC temperature), which I convert to degrees celsius so we can better see negative values.
- Display negative values on charts when there is negative data.

Before:
image

After:
image

- Add white space to the top of the line chart by 10% of the maximum chart value. This is to prevent the highest point of the line chart from being cropped off. Not sure if this is the best solution though so ideas are welcome!

Before:
image

After:
image

### DIFF
--- a/packages/ui-components/src/components/Axis/Axis.style.tsx
+++ b/packages/ui-components/src/components/Axis/Axis.style.tsx
@@ -24,7 +24,6 @@ export const AxisLeft = styled((props: AxisLeftProps) => (
     axisClassName={props.className ?? ""}
     numTicks={props.numTicks ?? 5}
     hideTicks={props.hideTicks}
-    hideZero={false}
     tickStroke={theme.palette.border.default}
     tickLabelProps={() => ({
       textAnchor: "end", // Horizontal anchor
--- a/packages/ui-components/src/components/Axis/Axis.style.tsx
+++ b/packages/ui-components/src/components/Axis/Axis.style.tsx
@@ -24,6 +24,7 @@ export const AxisLeft = styled((props: AxisLeftProps) => (
     axisClassName={props.className ?? ""}
     numTicks={props.numTicks ?? 5}
     hideTicks={props.hideTicks}
+    hideZero={false}
     tickStroke={theme.palette.border.default}
     tickLabelProps={() => ({
       textAnchor: "end", // Horizontal anchor
--- a/packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
+++ b/packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
@@ -83,9 +83,17 @@ export const LineIntervalChart = ({
         return (
           <RectClipGroup
             key={`rect-clip-${intervalIndex}`}
-            y={Math.min(yFrom, yTo)}
+            y={
+              intervalIndex === intervals.length - 1
+                ? Math.min(yFrom, yTo) - 50
+                : Math.min(yFrom, yTo)
+            }
             width={width}
-            height={clipHeight}
+            height={
+              intervalIndex === intervals.length - 1
+                ? clipHeight + 50
+                : clipHeight
+            }
           >
             <LineChart
               timeseries={timeseries}
--- a/packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
+++ b/packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
@@ -83,9 +83,17 @@ export const LineIntervalChart = ({
         return (
           <RectClipGroup
             key={`rect-clip-${intervalIndex}`}
-            y={Math.min(yFrom, yTo)}
+            y={
+              intervalIndex === intervals.length - 1
+                ? Math.min(yFrom, yTo) - 5
+                : Math.min(yFrom, yTo)
+            }
             width={width}
-            height={clipHeight}
+            height={
+              intervalIndex === intervals.length - 1
+                ? clipHeight + 5
+                : clipHeight
+            }
           >
             <LineChart
               timeseries={timeseries}
--- a/packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
+++ b/packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
@@ -27,6 +27,8 @@ export interface LineIntervalChartProps {
   xScale: ScaleTime<number, number>;
   /** d3-scale to transform point dates to pixel positions on the y-axis */
   yScale: ScaleLinear<number, number>;
+  /** Optional offset on the y-axis to pad the highest and lowest point on the chart */
+  topIntervalOffset?: number;
 }
 
 /**
@@ -48,6 +50,7 @@ export interface LineIntervalChartProps {
  *       xScale={xScale}
  *       yScale={yScale}
  *       intervals={intervals}
+ *       topIntervalOffset={5}
  *     />
  *   </svg>
  * )
@@ -56,12 +59,13 @@ export interface LineIntervalChartProps {
  * In this case, the line will be rendered in green where its y-coordinates are
  * between 10 and 20, and in red where its y-coordinates are between 20 and 40.
  * If the timeseries has values outside these intervals, those line segments
- * won't be rendered.
+ * won't be rendered. The chart will also have a padding of 5 at the top.
  *
  * @param timeseries Timeseries to represent as a line.
  * @param xScale d3-scale to transform point dates to pixel positions on the x-axis.
  * @param yScale d3-scale to transform point values to pixel positions on the y-axis.
  * @param intervals intervals that define the color of the line for different segments.
+ * @param topIntervalOffset Optional offset on the y-axis to pad the highest point on the chart.
  *
  * @returns An SVG group containing the line.
  */
@@ -70,13 +74,17 @@ export const LineIntervalChart = ({
   xScale,
   yScale,
   intervals,
+  topIntervalOffset,
 }: LineIntervalChartProps) => {
   const [xMin, xMax] = xScale.range();
   const width = Math.abs(xMax - xMin);
+  // Since this is a SVG, the highest point on the chart has the lowest y-coordinate.
+  const topPoint = Math.min(...intervals.map((interval) => interval.lower));
 
   return (
     <Group>
       {intervals.map((interval, intervalIndex) => {
+        console.log(interval);
         const yFrom = yScale(interval.lower);
         const yTo = yScale(interval.upper);
         const clipHeight = Math.abs(yFrom - yTo);
@@ -84,14 +92,14 @@ export const LineIntervalChart = ({
           <RectClipGroup
             key={`rect-clip-${intervalIndex}`}
             y={
-              intervalIndex === intervals.length - 1
-                ? Math.min(yFrom, yTo) - 5
+              interval.lower === topPoint
+                ? Math.min(yFrom, yTo) - (topIntervalOffset ?? 0)
                 : Math.min(yFrom, yTo)
             }
             width={width}
             height={
-              intervalIndex === intervals.length - 1
-                ? clipHeight + 5
+              interval.lower === topPoint
+                ? clipHeight + (topIntervalOffset ?? 0)
                 : clipHeight
             }
           >
--- a/packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
+++ b/packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
@@ -78,13 +78,10 @@ export const LineIntervalChart = ({
 }: LineIntervalChartProps) => {
   const [xMin, xMax] = xScale.range();
   const width = Math.abs(xMax - xMin);
-  // Since this is a SVG, the highest point on the chart has the lowest y-coordinate.
-  const topPoint = Math.min(...intervals.map((interval) => interval.lower));
-
+  const topPoint = Math.max(...intervals.map((interval) => interval.upper));
   return (
     <Group>
       {intervals.map((interval, intervalIndex) => {
-        console.log(interval);
         const yFrom = yScale(interval.lower);
         const yTo = yScale(interval.upper);
         const clipHeight = Math.abs(yFrom - yTo);
@@ -92,13 +89,13 @@ export const LineIntervalChart = ({
           <RectClipGroup
             key={`rect-clip-${intervalIndex}`}
             y={
-              interval.lower === topPoint
+              interval.upper === topPoint
                 ? Math.min(yFrom, yTo) - (topIntervalOffset ?? 0)
                 : Math.min(yFrom, yTo)
             }
             width={width}
             height={
-              interval.lower === topPoint
+              interval.upper === topPoint
                 ? clipHeight + (topIntervalOffset ?? 0)
                 : clipHeight
             }
--- a/packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
+++ b/packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
@@ -27,7 +27,7 @@ export interface LineIntervalChartProps {
   xScale: ScaleTime<number, number>;
   /** d3-scale to transform point dates to pixel positions on the y-axis */
   yScale: ScaleLinear<number, number>;
-  /** Optional offset on the y-axis to pad the highest and lowest point on the chart */
+  /** Optional offset on the y-axis to pad the highest point on the chart */
   topIntervalOffset?: number;
 }
 
--- a/packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
+++ b/packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
@@ -83,17 +83,9 @@ export const LineIntervalChart = ({
         return (
           <RectClipGroup
             key={`rect-clip-${intervalIndex}`}
-            y={
-              intervalIndex === intervals.length - 1
-                ? Math.min(yFrom, yTo) - 50
-                : Math.min(yFrom, yTo)
-            }
+            y={Math.min(yFrom, yTo)}
             width={width}
-            height={
-              intervalIndex === intervals.length - 1
-                ? clipHeight + 50
-                : clipHeight
-            }
+            height={clipHeight}
           >
             <LineChart
               timeseries={timeseries}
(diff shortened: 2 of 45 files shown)


packages/ui-components/src/stories/mockData.ts
```diff
@@ -14,6 +14,17 @@ const appleStockPoints = appleStock.map(
...
 export const appleStockTimeseries = new Timeseries(appleStockPoints);
 
+// We format the points from cityTemperature to match TimeseriesPoint<number>
+// so we can use them to initialize Timeseries.
+const NYCtemp = cityTemperature.map(
+  (p: { date: string; "New York": string }): TimeseriesPoint<number> => ({
+    date: new Date(p.date),
+    value: (parseFloat(p["New York"]) - 32) * (5 / 9),
+  })
+);
```
faihegberg: Convert to degrees celsius just so we can see more negative values in our stories.

packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
```diff
@@ -83,9 +83,17 @@ export const LineIntervalChart = ({
...
+                ? Math.min(yFrom, yTo) - 5
+                : Math.min(yFrom, yTo)
+            }
             width={width}
-            height={clipHeight}
+            height={
+              intervalIndex === intervals.length - 1
+                ? clipHeight + 5
+                : clipHeight
+            }
```
pnavarrc: Since the intervals are provided as prop, we shouldn't count on them to be sorted in a specific way. For example, this wouldn't work if the intervals are passed in reverse order.

Maybe find the interval that has the lower `y` starting point instead?

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
```diff
@@ -93,7 +95,7 @@ export const MetricLineThresholdChart = ({
           yScale={yScale}
           axisLeftProps={{
             tickFormat: (value) => metric.formatValue(value),
-            tickValues: thresholds,
+            tickValues: [0].concat(thresholds),
```
pnavarrc: I don't think we should force `0` to be part of the tick values (it can always be added to the thresholds if the value is relevant for a particular metric

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
```diff
@@ -77,10 +77,12 @@ export const MetricLineThresholdChart = ({
     range: [0, chartWidth],
   });
 
+  const minChartValue = min(intervals.map(({ lower }) => lower)) ?? minValue;
   const maxChartValue = max(intervals.map(({ upper }) => upper)) ?? maxValue;
 
   const yScale = scaleLinear({
-    domain: [0, maxChartValue],
+    // Add 10% to maxChartValue to prevent the chart's highest point from being cropped off.
+    domain: [minChartValue, maxChartValue * 1.1],
```
pnavarrc: I don't think we need to change the `yScale` domain to avoid cropping the intervals (I changed this in the PR and all the stories look the same)

packages/ui-components/src/stories/mockData.ts
```diff
@@ -14,6 +14,17 @@ const appleStockPoints = appleStock.map(
 
 export const appleStockTimeseries = new Timeseries(appleStockPoints);
 
+// We format the points from cityTemperature to match TimeseriesPoint<number>
+// so we can use them to initialize Timeseries.
+const NYCtemp = cityTemperature.map(
+  (p: { date: string; "New York": string }): TimeseriesPoint<number> => ({
+    date: new Date(p.date),
+    value: (parseFloat(p["New York"]) - 32) * (5 / 9),
+  })
```
pnavarrc, line 27: Maybe write a function to convert from Fahrenheit to Celsius to improve readability (even if we only use it once).

Consider fully spelling the name (e.g. `newYorkTemperature`) to improve readability

packages/ui-components/src/stories/mockData.ts
```diff
@@ -14,6 +14,17 @@ const appleStockPoints = appleStock.map(
...
+);
+
+export const NYCtempTimeseries = new Timeseries(NYCtemp);
```
pnavarrc: Same with the timeseries!

packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
```diff
@@ -83,9 +83,17 @@ export const LineIntervalChart = ({
         return (
           <RectClipGroup
             key={`rect-clip-${intervalIndex}`}
-            y={Math.min(yFrom, yTo)}
+            y={
+              intervalIndex === intervals.length - 1
+                ? Math.min(yFrom, yTo) - 5
```
pnavarrc: Maybe name the `5` for readability? it could also be an optional parameter of the chart (`firstIntervalOffset` or something)

packages/ui-components/src/components/Axis/Axis.style.tsx
```diff
@@ -24,6 +24,7 @@ export const AxisLeft = styled((props: AxisLeftProps) => (
     axisClassName={props.className ?? ""}
     numTicks={props.numTicks ?? 5}
     hideTicks={props.hideTicks}
+    hideZero={false}
```
pnavarrc: I don't think we should change this

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.stories.tsx
```diff
@@ -26,3 +26,11 @@ AppleStock.args = {
   metric: MetricId.APPLE_STOCK,
   region: newYork,
 };
+
+export const NewYorkCityTemperature = Template.bind({});
+NewYorkCityTemperature.args = {
+  width,
+  height,
+  metric: MetricId.NYC_TEMP,
```
pnavarrc: Consider fully spell the name `NYC_TEMPERATURE` for clarity

packages/ui-components/src/stories/NYCtempDataProvider.ts
```diff
@@ -0,0 +1,39 @@
...
+ * ranges of values.
+ *
+ * @example
+ * ```
+ * dataReference: {
+ *   providerId: "nyc_temperature"
+ * },
+ * ```
+ */
+export class NYCtempDataProvider extends SimpleMetricDataProviderBase {
```
pnavarrc: Not to increase scope too much, but maybe we could return the temperature data for other cities/states if we have them? That way we can use this provider with multiple states  to test Trends, for example

pnavarrc (changes requested):
Added some suggestions!

packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
mikelehen: What is this 5 for, exactly?  Padding on the bottom of the chart?  But we did padding on the top of the chart via a 1.1 multiplier in MetricLineChart, right?  Is there a way to make the padding work the same way?

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
mikelehen: I was telling Fai that these charts without a labeled 0 value look weird to me:
image

But I guess that's how CAN works and it's never been an issue.  I was thinking maybe if a chart has min/max values then we should put those on the axis labels, but not 100% sure.  Anyway, I'm fine with punting on this labeling for now.

faihegberg: I agree with Michael here, but I don't feel strongly. I added it not only because I think it's intuitive to see a zero on the chart, but putting it into the threshold would mean that zero is explicitly an upperbound of one level and a lowerbound of the next level, which is not necessarily true for the metric we're defining.

I'll leave out the zero for now and if there's a clarity issue in the future, we can always add it.

packages/ui-components/src/stories/NYCtempDataProvider.ts
faihegberg: Sounds great! I'll do this in a follow-up PR.

packages/ui-components/src/stories/mockData.ts
faihegberg: I assume you're referring to the renaming.

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
faihegberg: You're right that we don't need this to avoid cropping the intervals, but this does provide the extra vertical padding at the top of the chart. So the highest point of the line is not at the same vertical position as the top of the y-axis. Without this 10% padding (with the cropped line issue now fixed) it would look like:
image

vs. with the 10% padding it would look like:
image

Personally I think the latter option looks better, but I don't feel strongly, since there are probably other, more elegant ways to add white space. So I'll remove it for now, given that the cropped line issue is the prioritized fix and that is already achieved.

packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
faihegberg: @pnavarrc I renamed the 5 and made it an optional variable 🙂 

@mikelehen The 5 is variable that determines the amount of padding _just at the top interval_ of the chart. This is the key ingredient that resolves the line getting cropped issue, as it adjusts the `RectClipGroup` of only the top interval. In contrast, the 1.1 multiplier in the `MetricLineChart` expands the domain of the yScale by increasing its upperbound. This forces the y-axis to extend longer at the top. So although both of these introduce white space, they do so in different ways. That said, I just got rid of this 1.1 multiplier ([see this comment](https://github.com/covid-projections/act-now-packages/pull/388#discussion_r1023225213)). I can also explain on a call if you want 🙂

packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
```diff
@@ -70,10 +74,11 @@ export const LineIntervalChart = ({
   xScale,
   yScale,
   intervals,
+  topIntervalOffset,
 }: LineIntervalChartProps) => {
   const [xMin, xMax] = xScale.range();
   const width = Math.abs(xMax - xMin);
-
+  const topPoint = Math.max(...intervals.map((interval) => interval.upper));
```
pnavarrc: This won't work because if the yScale is reversed, then the top most interval will be the one that has lower `upper` and `lower` values. For example, with:

```tsx
const yScale = scaleLinear({
  domain: [0, 1000],
  range: [0, 100]
})
```

will cause `topPoint` will be the interval at the bottom of the chart, which is not what we want.

Also, `upper` could end up at the bottom of the interval (again, depending on the scale), so we will need to compare both the `upper` and the `lower` ends of the interval (that's why we use `Math.min(yFrom, yTo)` instead of just `yFrom` or `yTo`).

faihegberg: Good point 😅

packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
```diff
@@ -27,6 +27,8 @@ export interface LineIntervalChartProps {
   xScale: ScaleTime<number, number>;
   /** d3-scale to transform point dates to pixel positions on the y-axis */
   yScale: ScaleLinear<number, number>;
+  /** Optional offset on the y-axis to pad the highest point on the chart */
```
mikelehen: This is a pixel offset, right?  Can you say "Optional offset (in pixels)" to make that clear? Also, maybe add `@default 5` per Chelsi's note at standup.

packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
```diff
@@ -56,12 +59,13 @@ export interface LineIntervalChartProps {
  * In this case, the line will be rendered in green where its y-coordinates are
  * between 10 and 20, and in red where its y-coordinates are between 20 and 40.
  * If the timeseries has values outside these intervals, those line segments
- * won't be rendered.
+ * won't be rendered. The chart will also have a padding of 5 at the top.
```
mikelehen: I'd remove this and let the `topIntervalOffset` documentation communicate this.

packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
```diff
@@ -56,12 +59,13 @@ export interface LineIntervalChartProps {
...
  *
  * @param timeseries Timeseries to represent as a line.
  * @param xScale d3-scale to transform point dates to pixel positions on the x-axis.
  * @param yScale d3-scale to transform point values to pixel positions on the y-axis.
  * @param intervals intervals that define the color of the line for different segments.
+ * @param topIntervalOffset Optional offset on the y-axis to pad the highest point on the chart.
```
mikelehen: I don't think we need / want all of these `@param`s here.  They seem redundant and don't show up in the generated docs: https://act-now-packages.web.app/docs/functions/ui_components_src.LineIntervalChart.html

If in doubt, maybe double-check with @chasulin since she's auditing prop docs right now.

packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
```diff
@@ -83,9 +93,17 @@ export const LineIntervalChart = ({
...
+                ? Math.min(yFrom, yTo) - (topIntervalOffset ?? 0)
+                : Math.min(yFrom, yTo)
+            }
             width={width}
-            height={clipHeight}
+            height={
+              Math.min(yFrom, yTo) === topPoint
+                ? clipHeight + (topIntervalOffset ?? 0)
+                : clipHeight
+            }
```
mikelehen: Rather than duplicate this logic twice (which makes it a bit harder to read and risks them getting out-of-sync, etc.), how about add this above:

```
const topOffset = (topIntervalOffset && Math.min(yFrom, yTo) === topPoint) ? topIntervalOffset : 0;
```

And then you can just do `y={Math.min(yFrom, yTo) - topIntervalOffset}` and `height={clipHeight + topIntervalOffset}`

packages/ui-components/src/stories/NYCtempDataProvider.ts
```diff
@@ -0,0 +1,39 @@
...
+  SimpleMetricDataProviderBase,
+  MetricData,
+  Metric,
+} from "@actnowcoalition/metrics";
+import { NYCtemperatureTimeseries } from "./mockData";
+
+/**
+ * Mock data provider with constant timeseries data for all regions. Useful
+ * for charts and other components when we would like stable dates and
+ * ranges of values.
```
mikelehen: I don't really understand how this description relates to what this data provider actually does or is for.  Shouldn't it say "Data provider that provides NYC temperature data for all regions. " or something?  And maybe include a TODO about making it include real data for other cities if we plan to do that in the future. 🤷 

FWIW- I'm also not 100% sure why we need this provider.  We could use the existing MockDataProvider with a negative `minValue` if all we want is timeseries data that includes negative numbers.  

That said, it doesn't seem like a bad idea to have some more realistic data available for testing, so I'm not opposed to adding temperature data for storybook testing purposes if we want to. 🤷

packages/ui-components/src/stories/NYCtempDataProvider.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @example
+ * ```
+ * dataReference: {
+ *   providerId: "nyc_temperature"
+ * },
+ * ```
+ */
+export class NYCtempDataProvider extends SimpleMetricDataProviderBase {
```
mikelehen: nit: I think this should be `NycTempDataProvider` to match our convention with e.g. CsvDataProvider...  or are we inconsistent on that?

Also I'd spell out temperature, since "temp" is somewhat ambiguous (temp => temporary)

packages/ui-components/src/stories/mockData.ts
```diff
@@ -14,6 +14,21 @@ const appleStockPoints = appleStock.map(
 
 export const appleStockTimeseries = new Timeseries(appleStockPoints);
 
+function toCelsius(fahrenheit: number): number {
+  return (fahrenheit - 32) * (5 / 9);
+}
+
+// We format the points from cityTemperature to match TimeseriesPoint<number>
+// so we can use them to initialize Timeseries.
+const NYCtemperature = cityTemperature.map(
```
mikelehen: Similarly, this should perhaps be nycTemperature?

packages/ui-components/src/stories/mockData.ts
```diff
@@ -14,6 +14,21 @@ const appleStockPoints = appleStock.map(
...
+  (p: { date: string; "New York": string }): TimeseriesPoint<number> => ({
+    date: new Date(p.date),
+    value: toCelsius(parseFloat(p["New York"])),
+  })
+);
+
+export const NYCtemperatureTimeseries = new Timeseries(NYCtemperature);
```
mikelehen: And nycTemperatureTimeseries?

mikelehen (approved):
A few nits, but overall looks good!

packages/ui-components/src/components/LineIntervalChart/LineIntervalChart.tsx
faihegberg: I wonder whether this is not showing up in the generated doc because it's an optional prop, but I'll remove this line.

packages/ui-components/src/stories/NYCtempDataProvider.ts
faihegberg: Whoops, yes, this comment is supposed to be updated. I'll revise it. I think it's beneficial to have some variety of realistic data for us to test so I'll keep it. I'll also make sure this provider provides temperature for other cities too as a follow-up PR.

PR #398, opened 2022-11-15
Error / loading state infrastructure improvements.

mikelehen:
* Ensures that useData() hooks always log any errors encountered.
* Adds ability to have mock metrics simulate loading delays and errors.
* Disables SWR caching so that we can actually see simulated delays.
* Adds a couple mock metrics that simulate loading delays / errors and adds stories to all metric-aware components to use them.

This does not actually improve the loading state or error handling for any components yet.

### DIFF
--- a/packages/metrics/package.json
+++ b/packages/metrics/package.json
@@ -34,6 +34,7 @@
     "@actnowcoalition/regions": "^0.1.1",
     "@actnowcoalition/time-utils": "^0.1.0",
     "@types/papaparse": "^5.3.3",
+    "delay": "^5.0.0",
     "lodash": "^4.17.21",
     "node-fetch": "^2.6.7",
     "p-limit": "^3.1.0",
--- a/packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
+++ b/packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
@@ -95,13 +95,20 @@ export class MultiRegionMultiMetricDataStore<T = unknown> {
     dataFetcher: (region: Region, metric: Metric) => Promise<MetricData<T>>
   ): Promise<MultiRegionMultiMetricDataStore<T>> {
     const data: { [regionId: string]: MultiMetricDataStore<T> } = {};
-    for (const region of regions) {
-      const regionData: { [metricId: string]: MetricData<T> } = {};
-      for (const metric of metrics) {
-        regionData[metric.id] = await dataFetcher(region, metric);
-      }
-      data[region.regionId] = new MultiMetricDataStore<T>(region, regionData);
-    }
+    await Promise.all(
+      regions.map(async (region) => {
+        const regionData: { [metricId: string]: MetricData<T> } = {};
+        await Promise.all(
+          metrics.map(async (metric) => {
+            regionData[metric.id] = await dataFetcher(region, metric);
+          })
+        ),
+          (data[region.regionId] = new MultiMetricDataStore<T>(
+            region,
+            regionData
+          ));
+      })
+    );
 
     return new MultiRegionMultiMetricDataStore(data);
   }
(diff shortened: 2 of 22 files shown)


packages/ui-components/src/components/MetricOverview/MetricOverview.stories.tsx
```diff
@@ -50,6 +50,18 @@ NoExtendedName.args = {
...
+LoadingDelay.args = {
+  region: newYorkState,
+  metric: MetricId.MOCK_CASES_DELAY_1S,
+};
+
+export const LoadingError = Template.bind({});
+LoadingError.args = {
+  region: newYorkState,
+  metric: MetricId.MOCK_CASES_ERROR,
+};
```
mikelehen: @pnavarrc I'm not sure if it makes sense to go through every component and add stories that use `MetricId.MOCK_CASES_DELAY1S` and `MetricId.MOCK_CASES_ERROR` so that we can verify the loading / error handling or if there's an easier / cleaner way.  Any thoughts?

packages/metrics/package.json
```diff
@@ -34,6 +34,7 @@
     "@actnowcoalition/regions": "^0.1.1",
     "@actnowcoalition/time-utils": "^0.1.0",
     "@types/papaparse": "^5.3.3",
+    "delay": "^5.0.0",
```
mikelehen, line 37: FWIW- This module just creates a promise that resolves after a specified timeout and we could trivially implement it ourselves which I'm happy to do if we prefer not use trivial modules like this.

packages/metrics/src/data/MultiRegionMultiMetricDataStore.ts
```diff
@@ -95,13 +95,20 @@ export class MultiRegionMultiMetricDataStore<T = unknown> {
...
+          metrics.map(async (metric) => {
+            regionData[metric.id] = await dataFetcher(region, metric);
+          })
+        ),
+          (data[region.regionId] = new MultiMetricDataStore<T>(
+            region,
+            regionData
+          ));
+      })
+    );
```
mikelehen, line 111: `MockDataProvider` uses `fromRegionsAndMetricsAsync()` to help it construct the `MultiRegionMultiMetricDataStore` with the "fetched" data and I noticed that when I introduced the delay functionality, components that were requesting many metrics or regions (e.g. a map) were loading *very* slowly.  It turns out this code was awaiting each promise serially so that the delays added up, instead of doing it in parallel so the delays all completed simultaneously.  So I rewrote it using `Promise.all()` to make sure we were waiting for all the promises in parallel.

packages/ui-components/src/components/MetricOverview/MetricOverview.stories.tsx
mikelehen: We discussed at standup and Pablo said this seemed like the best approach.

packages/ui-components/src/common/hooks/metric-data.ts
```diff
@@ -37,7 +37,11 @@ export function useData(
     return catalog.fetchData(region, metric, includeTimeseries);
   };
 
-  return useSWR(args, fetcher);
+  return useSWRWrapper(
+    args,
+    fetcher,
+    `useData(region=${region}, metric=${metric}) failed: `
```
smcclure17: nit: Having `failed: ` both in this message and in [`useSWRWrapper()`](https://github.com/covid-projections/act-now-packages/blob/cd6d1f955f843733d31d0838ebc45355d6a1d9d9/packages/ui-components/src/common/hooks/metric-data.ts#L122) creates a message like `...Metric(mock_cases): Cases (mock)]) failed: failed: Error: Simulated ...`, maybe consider removing one of the two?

smcclure17 (approved):
LGTM! thanks for going through and updating the stories 🙏

packages/ui-components/src/common/hooks/metric-data.ts
mikelehen: Oops, great catch!

PR #399, opened 2022-11-15
Reenable typedoc with workarounds to avoid out-of-memory crash.

### LINKED ISSUE #356: Investigate 'Javascript heap out of memory' error
[An error saying "Javascript heap out of memory"](https://github.com/covid-projections/act-now-packages/actions/runs/3371400883/jobs/5593571363) popped up in the [WorldMap component PR](https://github.com/covid-projections/act-now-packages/pull/350), and blocked the pr from building successfully (and deploying to firebase). The source of the error seems to be the generation of typedocs, and [a quick fix was just to disable typedocs](https://github.com/covid-projections/act-now-packages/pull/350/commits/5bf4d08d66c17b6290c13fb8ba0cd1e392ac9cce) (remove yarn generate-docs from the build script) for now, so as to not block merging the WorldMap.

Things we've tried so far: 
- Doubling the amount of allotted memory
- Upgrading typedocs

mikelehen:
Typedoc has issues with monorepo support using lots of memory.  See https://github.com/covid-projections/act-now-packages/issues/356 and https://github.com/TypeStrong/typedoc/issues/1606.

This avoids the crash in two different ways:
1. I increased the node heap size to 8192 which seems to be sufficient to make the build succeed for now.
2. For now I am pointing typedoc at the packages' source code directly rather than using the monorepo support. This uses a lot less memory and is a lot faster, though the resulting docs are not formatted as nicely (it looks like all the code lives in one big package).  We can revisit this when https://github.com/TypeStrong/typedoc/issues/1835 is fixed.

I'm also upgrading typedoc just for good measure, though I don't think it fixes anything.

Not including a changeset, since this has no effect on our packages.

Fixes #356 for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,9 +26,10 @@
     "release": "yarn build:all && yarn changeset publish",
     "create-package": "yarn plop package",
     "create-component": "yarn plop component",
-    "generate-docs": "yarn typedoc --out public/docs --entryPointStrategy packages --name 'Act Now Packages' .",
+    "generate-module-docs": "yarn typedoc --out public/docs --entryPointStrategy packages --name 'Act Now Packages' . # See #356",
+    "generate-src-docs": "yarn typedoc --out public/docs packages/*/src/index.ts --name 'Act Now Packages' # See #356",
     "generate-storybook": "yarn build-storybook -c packages/ui-components/.storybook/ -o public/storybook",
-    "build-public": "yarn build:all && yarn generate-storybook"
+    "build-public": "yarn build:all && yarn generate-storybook && yarn generate-src-docs"
   },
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -54,7 +54,7 @@
     "ts-jest": "^28.0.2",
     "ts-node": "^10.7.0",
     "tsconfig-paths": "^4.0.0",
-    "typedoc": "^0.23.10",
+    "typedoc": "^0.23.21",
```
mikelehen, line 58: Upgrading for good measure, though it wasn't actually necessary.

chasulin (approved):
🙌🏼🙌🏼

PR #402, opened 2022-11-16
CSVDataProvider: Fix handling of missing values

### LINKED ISSUE #366: CSVDataProvider incorrectly parsing missing values
`CSVDataProvider` seems to be parsing missing values as `""` instead of `null` or `undefined` which means they don’t get filtered out automatically, and we have `""` in the middle of numeric time series, which makes `assertFiniteNumbers()` fail.

mikelehen:
Also add/improve some tests to make sure we handle things that people hit during the hackathon.

Fixes #366.

### DIFF
--- a/packages/metrics/src/data_providers/CsvDataProvider.test.ts
+++ b/packages/metrics/src/data_providers/CsvDataProvider.test.ts
@@ -10,6 +10,7 @@ const mockCsv = `region,cool_metric
 
 const csvTimeseries = `region,date,cool_metric
 36,2022-08-02,150
+36,2022-08-03,
 12,2022-08-02,`;
 
 const newYork = states.findByRegionIdStrict("36");
@@ -29,14 +30,16 @@ const testMetric = new Metric({
 const testFetchingCsvData = async (
   data: string,
   includeTimeseries: boolean,
-  dateCol?: string
+  dateCol?: string,
+  metric?: Metric
 ) => {
+  metric = metric ?? testMetric;
   const provider = new CsvDataProvider(PROVIDER_ID, {
     regionColumn: "region",
     dateColumn: dateCol,
     csvText: data,
   });
-  return (await provider.fetchData([newYork], [testMetric], includeTimeseries))
+  return (await provider.fetchData([newYork], [metric], includeTimeseries))
     .regionData(newYork)
     .metricData(testMetric);
 };
@@ -54,8 +57,11 @@ describe("CsvDataProvider", () => {
     );
     expect(metricDataNoTs.currentValue).toBe(150);
     expect(metricDataNoTs.hasTimeseries()).toBe(false);
-    expect(metricDataTs.currentValue).toBe(150);
+
+    expect(metricDataTs.hasTimeseries()).toBe(true);
+    expect(metricDataTs.timeseries.length).toBe(1);
     expect(metricDataTs.timeseries.lastValue).toBe(150);
+    expect(metricDataTs.currentValue).toBe(150);
   });
 
   test("fetchData() returns non-timeseries data if timeseries data is not available.", async () => {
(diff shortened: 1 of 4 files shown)


packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -10,6 +10,7 @@ const mockCsv = `region,cool_metric
 
 const csvTimeseries = `region,date,cool_metric
 36,2022-08-02,150
+36,2022-08-03,
```
mikelehen, line 13: We want to test that missing values turn into nulls (and then get removed from the timeseries) rather than parsed as empty strings.

packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -29,14 +30,16 @@ const testMetric = new Metric({
 const testFetchingCsvData = async (
   data: string,
   includeTimeseries: boolean,
-  dateCol?: string
+  dateCol?: string,
+  metric?: Metric
```
mikelehen, line 34: Make it possible to use `testFetchingCsvData()` with a specified metric, so I can test bad metric definitions.

packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -54,8 +57,11 @@ describe("CsvDataProvider", () => {
     );
     expect(metricDataNoTs.currentValue).toBe(150);
     expect(metricDataNoTs.hasTimeseries()).toBe(false);
-    expect(metricDataTs.currentValue).toBe(150);
```
mikelehen, line 57: moved this below.

packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -54,8 +57,11 @@ describe("CsvDataProvider", () => {
...
+
+    expect(metricDataTs.hasTimeseries()).toBe(true);
+    expect(metricDataTs.timeseries.length).toBe(1);
```
mikelehen, line 62: Prior to this PR, we would have parsed the timeseries to have 2 values with one being an empty string.

packages/metrics/src/data_providers/CsvDataProvider.test.ts
```diff
@@ -73,6 +79,24 @@ describe("CsvDataProvider", () => {
     ).rejects.toThrow("CSV must not be empty.");
   });
 
+  test("fetchData() fails if metric is missing a 'column' property.", async () => {
```
mikelehen, line 82: This error was already present and worked, but I added a test since people complained that it was failing "silently" during the hackathon.  In reality, the error was being surfaced by the `useData()` hooks, but they were not handling it.  So #398 added logging to make it harder to ignore errors from the hooks.

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -175,7 +175,9 @@ function sanitizeRows(rows: DataRow[], excludeColumns: string[]): void {
         continue;
       }
       const v = row[c] as string;
-      if (/^-?[\d.eE]+$/.test(v)) {
+      if (v.length === 0) {
+        row[c] = null;
```
mikelehen, line 179: This is the main fix.  We will treat `""` as `null` and then `dataRowsToMetricData()` will filter it out when it constructs the timeseries.

packages/metrics/src/data_providers/data_provider_utils.test.ts
```diff
@@ -34,6 +34,7 @@ describe("dataRowsToMetricData()", () => {
       "date"
     );
     expect(data.currentValue).toBe(100);
+    expect(data.timeseries.length).toBe(2);
```
mikelehen, line 37: This was me just sanity-checking that rows with missing data get removed, so the 3rd row is removed and our timeseries has length 2.

faihegberg (approved):
LGTM!

PR #415, opened 2022-11-17
Consistent component syntax + prop interface file structure

chasulin:
**This PR makes the following things consistent throughout our ui-components package:**
- Component syntax
- Storybook syntax
- Props file-structure (no longer using `interfaces.ts` files)

**We are using the following conventions:**
- Component syntax
    - We’re no longer using `React.FC` type declaration for components. (Declaring React.FC used to automatically give the `children` prop, but that’s no longer the case. It doesn’t seem to be widely used anymore.)
    - We use the syntax used [in these Typescript-React docs](https://github.com/Microsoft/TypeScript-React-Starter#creating-a-component)
- Prop interfaces are written directly in the component’s file (rather than in a separate interfaces.ts file) and are named `{{componentName}}Props`
- Storybook files use `ComponentStory` and `ComponentMeta` for typing (rather than `Story` and `Meta`, which are also included in @storybook/react). This lets us avoid importing the component’s props into the storybook file. [See here for a full explanation with examples](https://github.com/storybookjs/storybook/discussions/16650#discussioncomment-1617480).

**Notes for reviewer 🙋🏻‍♀️ :**
- US maps were previously grouped, with one directory called `USMaps` and one called `MetricUSMaps`. These are now further separated for consistency, now as `USStateMap`, `USNationalMap`, `MetricUSStateMap`, and `MetricUSNationalMap`. These files will look newly added/newly deleted, but they're really just rearranged (+all changes relevant to this PR)
- Compare still has an interfaces.ts file. I haven't landed on the best way to nix this one. Punting to another PR.
- There are a few instances of `Story` left in the code (rather than `ComponentStory`) - [see here for one](https://github.com/covid-projections/act-now-packages/blob/casulin/component-syntax/packages/ui-components/src/components/LegendThreshold/LegendThreshold.stories.tsx#L15-L18). All of the remaining uses of Story are for components that use generics in their prop interfaces. I couldn't figure out how to use `ComponentStory` with generics.. [added a gh issue](https://github.com/covid-projections/act-now-packages/issues/428) and punting to another PR.


packages/ui-components/src/common/utils/charts.ts
```diff
@@ -0,0 +1,8 @@
+export interface BaseChartProps {
+  width: number;
+  height: number;
+  marginTop?: number;
+  marginBottom?: number;
+  marginLeft?: number;
+  marginRight?: number;
+}
```
chasulin, line 8: If this doesn't feel like the right place for this interface, lmk

packages/ui-components/src/common/utils/maps.ts
```diff
@@ -1,5 +1,12 @@
 import { RegionDB, Region } from "@actnowcoalition/regions";
 
+export interface BaseUSMapProps {
+  renderTooltip: (regionId: string) => React.ReactNode;
+  getFillColor?: (regionId: string) => string;
+  getRegionUrl?: (regionId: string) => string | undefined;
+  width?: number;
+}
+
```
chasulin, line 9: [Same for this one](https://github.com/covid-projections/act-now-packages/pull/415/files#r1028581965). I moved these two commonly re-used interfaces (this one + the above `BaseChartProps`) to common/utils, in their respective category's file (ie, maps, charts..)

plopfile.mjs
```diff
@@ -98,7 +98,7 @@ export interface {{pascalCase name}}Props {
 
 }
 
-export const {{pascalCase name}}: React.FC<{{pascalCase name}}Props> = () => {
+export const {{pascalCase name}} = (props: {{pascalCase name}}Props) => {
```
chasulin, line 101: Updated plopfile to match our convention

packages/ui-components/src/common/utils/charts.ts
mikelehen: Makes sense to me!

packages/ui-components/src/components/CompareTable/CompareTable.stories.tsx
```diff
@@ -18,7 +18,7 @@ export default {
   component: CompareTable,
 } as ComponentMeta<typeof CompareTable>;
 
-interface RowItem {
+export interface RowItem {
```
mikelehen, line 21: Why are we exporting this? 🤔

packages/ui-components/src/components/MetricCompareTable/utils.tsx
```diff
@@ -14,7 +12,7 @@ import {
   getAriaSort,
 } from "../CompareTable";
 import { StyledTableCell, StyledLink } from "./MetricCompareTable.style";
-import { Row } from "./interfaces";
+import { Row } from "./MetricCompareTable";
```
mikelehen, line 15: FWIW- I'm having slight reservations about getting rid of the interfaces.ts file in this case since now we have a circular dependency between MetricCompareTable.tsx and utils.tsx.  Circular dependencies are allowed in JavaScript / TypeScript but they can be kind of dicey (e.g. if for some reason you tried to do `const table = MetricCompareTable()` at the top-level of `utils.tsx` it would probably fail at runtime with a probably-not-very-friendly error).

In this case, I think it's fine, but I guess I'm realizing that I'm not opposed to having `interfaces.ts` files in cases like this where we have interfaces that are shared across multiple files.  But I am still opposed to just arbitrarily creating interfaces.ts files for components when there's no sharing going on.  FWIW another option in this case would just be to define `Row` in this `utils.tsx` file. That would be fine with me too.

Anyway, feel free to keep this as-is. I don't think it's a big deal. Just sharing my evolving style preferences.

packages/ui-components/src/components/MetricLegendThreshold/utils.ts
```diff
@@ -1,6 +1,6 @@
 import { assert } from "@actnowcoalition/assert";
 import { Metric } from "@actnowcoalition/metrics";
-import { CategoryItem } from "./interfaces";
+import { CategoryItem } from "./MetricLegendThreshold";
```
mikelehen, line 3: Same deal here (and maybe other places we were using interfaces.ts 🤷 )

packages/ui-components/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -26,7 +32,7 @@ export interface ProgressBarProps {
  * currentValue=35.
  *
  * By default, the role of the component is 'meter', since the role
- * 'progressbar' is appropiate to show progress of tasks.
+ * 'progressbar' is appropriate to show progress of tasks.
```
mikelehen, line 35: 👏

packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
```diff
@@ -85,3 +89,31 @@ export const RegionSearch: React.FC<RegionSearchProps> = ({
     />
   );
 };
+
+export interface RegionSearchItemProps {
```
mikelehen, line 93: Any particular reason you refactored SearchItem.tsx into this file?  I don't particularly mind, just curious.

packages/ui-components/src/components/USNationalMap/StatesMap.tsx
```diff
@@ -1,26 +1,29 @@
 import React from "react";
 import { GeoPath } from "d3-geo";
 import { Tooltip, Link } from "@mui/material";
-import { statesGeographies } from "../../../common/geo-shapes";
-import { RegionOverlay, RegionShapeBase } from "../Maps.style";
+import { statesGeographies } from "../../common/geo-shapes";
+import { RegionOverlay, RegionShapeBase } from "../../styles/common/Maps.style";
 
-const StatesMap: React.FC<{
+// TODO (#427) - improve upon this props interface
```
mikelehen: I'm hesitant to give this feedback since I appreciate the diligence of logging things in github, but I will very gently suggest that some TODOs may be fine to continue living as just TODOs in code, in particular if it's just code cleanup and it's not something urgent or that has ramifications across the codebase.  In that case, just having a note in the code as a reminder to whoever looks at this code next might be sufficient?

For Issues like #427, in my experience if they don't get fixed immediately, they will probably just sit around for weeks / months / years because there will always be more pressing work to do, so they can end up being a bit noisy / distracting. 🤷   That said, I don't care that strongly. We can always close it later.  And if you're planning to tackle it right away, then cool. 👍

mikelehen (approved):
This looks great!  Thanks for the thorough cleanup, the thorough PR description, and the comments to explain some of the changes.  It all makes sense to me.  I left a couple questions / comments, but mostly as food for thought.  Feel free to ship this as-is.

packages/ui-components/src/components/MetricCompareTable/utils.tsx
chasulin: ooooh this is a very good point! i definitely agree, not a fan of introducing circular dependencies.

instead of going back through this PR, maybe it would be easier if i just merge as-is, and make the edits pertaining to this in a follow-up PR today? easier to review if its separated out, etc..

packages/ui-components/src/components/CompareTable/CompareTable.stories.tsx
chasulin: oh i just exported interfaces that werent previously exported for the sake of typedoc//making sure everything was exposed. was this one a step too far?

packages/ui-components/src/components/MetricCompareTable/utils.tsx
mikelehen: Works for me!

packages/ui-components/src/components/USNationalMap/StatesMap.tsx
chasulin: good point. ill nix it and just leave a note in the code

packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
chasulin: i appended `Region` to the front of the naming since it was only used by RegionSearch, and i thought it was overall kind of regiony-specific. but actually it just takes strings and an icon color, so maybe its generic enough to be its own thing in its own folder, to be used by other types of search..

i moved it into the parent component's file to match what we have for MetricTooltip/MetricTooltipContent ([both in the same file](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/components/MetricTooltip/MetricTooltip.tsx#L32-L69)). trying to make things match/consistent is a rabbit hole 🌀

mikelehen: 👍  Heh, yeah.  The good news is that it'll never be perfect but this PR moves us drastically in the right direction and will make it easier to know what "consistent" even means going forward. 😁

PR #426, opened 2022-11-21
Fix floating zero on y-axis

### LINKED ISSUE #401: Chart: Y-axis padding is making the zero-tick float
Currently for `MetricLineThresholdChart`, we're adding padding at the bottom of the lowest threshold and at the top of the highest threshold ([code](https://github.com/covid-projections/act-now-packages/blob/209259be4e953cd3e0a4d564f0383a6118bf3709/packages/ui-components/src/components/MetricLineThresholdChart/utils.ts#L81-L105)). This padding causes a tick value (in this case, 0) to float above to x-axis line, resulting in the chart looking off, as follows:
Image
In this case, the 0 tick value is higher than the lower bound of the lowest threshold interval.

The padding was originally implemented to enable white space at the top and bottom of the chart, but this can potentially be achieved in a different way. Below are some considerations based on my discussion with Pablo:
- This padding is currently calculated in `utils.ts` in a pretty arbitrary way (just multiplying by 0.2), and without knowledge of how this would transform to svg units, which happens later in `LineIntervalChart.tsx`.
- We have a stronger need to pad the top of the chart than the bottom. It's worth exploring keeping some form of padding at the top, but removing the padding at the bottom.
- Since currently this padding is calculated based on data values, not svg units, we should perhaps remove this padding calculation and explore adding padding in `LineIntervalChart.tsx` instead.

faihegberg:
Closes https://github.com/covid-projections/act-now-packages/issues/401

The original issue linked above states that we should explore removing the padding from `MetricLineThresholdChart/utils.ts` and adding padding calculation to `LineIntervalChart.tsx` instead. On second thought, I think there is value is keeping this padding calculation in the utils file so we don't overcomplicate our `LineIntervalChart` implementation (although I don't feel strongly). I also don't think we necessarily need to remove padding from the bottom, since the only time we should is when the data's minimum value is above zero but the lower bound of the first interval is below zero.

So this PR simply implements a conditional where if the data's minimum value is higher than zero, then the lowest y-axis label should not be lower than zero.

Before
image

After
image

To test
- Adjust `tickValues` prop of `AxesTimeseries` in `MetricThresholdChart.tsx` to include zero (i.e. `tickValues: [0].concat(thresholds)` OR
- Play around with metric thresholds (i.e. add zero as a threshold).

### DIFF
--- a/packages/ui-components/src/components/MetricLineThresholdChart/utils.ts
+++ b/packages/ui-components/src/components/MetricLineThresholdChart/utils.ts
@@ -91,7 +91,7 @@ export function calculateChartIntervals(
 
   // If minValue is higher than 0, don't pad below 0 on the y-axis.
   const lowestBound = Math.min(minVal, firstThreshold - padding);
-  const chartMin = lowestBound >= 0 ? Math.max(0, lowestBound) : lowestBound;
+  const chartMin = minVal >= 0 ? Math.max(0, lowestBound) : lowestBound;
   const chartMax = Math.max(maxVal, lastThreshold + padding);
 
   // Build the intervals in the same order as the categories.
--- a/packages/ui-components/src/components/MetricLineThresholdChart/utils.ts
+++ b/packages/ui-components/src/components/MetricLineThresholdChart/utils.ts
@@ -89,26 +89,19 @@ export function calculateChartIntervals(
       ? 0.2 * (lastThreshold - firstThreshold)
       : 0.2 * (maxVal - minVal);
 
+  // If minValue is higher than 0, don't pad below 0 on the y-axis.
+  const lowestBound = Math.min(minVal, firstThreshold);
+  const chartMin = lowestBound >= 0 ? 0 : lowestBound - padding;
+  const chartMax = Math.max(maxVal, lastThreshold + padding);
+
   // Build the intervals in the same order as the categories.
   return metricCategories.map((category, categoryIndex) => {
     const isFirstCategory = categoryIndex === 0;
     const isLastCategory = categoryIndex === metricCategories.length - 1;
-    const lowestBound = Math.min(minVal, firstThreshold - padding);
-    // If minValue is higher than 0, don't pad below 0 on the y-axis.
-    const firstCategoryLowerThreshold =
-      isFirstCategory && minValue > 0 ? Math.max(0, lowestBound) : lowestBound;
-    const lastCategoryUpperThreshold = Math.max(
-      maxVal,
-      lastThreshold + padding
-    );
     return {
       category,
-      lower: isFirstCategory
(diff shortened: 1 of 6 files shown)


packages/ui-components/src/components/MetricLineThresholdChart/utils.ts
```diff
@@ -91,13 +93,21 @@ export function calculateChartIntervals(
   return metricCategories.map((category, categoryIndex) => {
     const isFirstCategory = categoryIndex === 0;
     const isLastCategory = categoryIndex === metricCategories.length - 1;
+    const lowestBound = Math.min(minVal, firstThreshold - padding);
+    // If minValue is higher than 0, don't pad below 0 on the y-axis.
+    const firstCategoryLowerThreshold =
+      isFirstCategory && minValue > 0 ? Math.max(0, lowestBound) : lowestBound;
```
mikelehen: There are a few things about this code that I think makes it a bit more complicated than necessary:
1. `firstCategoryLowerThreshold` will only be used below if `isFirstCategory` is true, which means that the `lowestBound` case of this ternary operator is irrelevant.  It's best practice to never have irrelevant code since it makes the code more complicated and it's hard to know why the code is there.
2. `firstCategoryLowerThreshold` and `lastCategoryLowerThreshold` are being recalculated every time through the `.map()` loop even though the value won't really change.  It's best not to recalculate things if you don't need to for clarity and performance.
3. It could be a bit confusing to have `firstThreshold` and `firstCategoryLowerThreshold` as different variables. Based on the names I can't really tell how they are different.

I would probably suggest:
1. Renaming to `chartMin` and `chartMax` or similar.
2. Calculating them once at the top of the function, outside the `.map()` loop.

I also think you could simplify the logic a tad:
```
const lowestBound = Math.min(minVal, firstThreshold);
const chartMin = (lowestBound >= 0) ? 0 : lowestBound - padding;
```

mikelehen (approved):
This approach looks good to me, but I think it could be simplified a bit.

Also, would it be possible to tweak one of our stories in storybook to demonstrate the effect of your change, like the screenshots you included in the PR description?  That way if somebody changes this code in the future it'll be easier to tell if they broke something.  Thanks!

I'm approving but let me know if you want a second set of eyes on it before merging.

packages/ui-components/src/components/MetricLineThresholdChart/utils.ts
faihegberg: Thanks Michael! Also adding a note here since we further discussed this on Slack. The solution above definitely eliminates the bug, although this may imply that a chart with a positive `minVal` will have its y-axis start at zero, regardless of how high the `minVal` is. This is fine for now and we can fine-tune how we deal with y-axis scaling once we have a better idea of user needs.

PR #439, opened 2022-11-22
Add parseCsv() and generateCsv() helpers to the metrics package.

mikelehen:
Fixes https://github.com/covid-projections/act-now-template/issues/50

### DIFF
--- a/packages/metrics/src/data_providers/CsvDataProvider.ts
+++ b/packages/metrics/src/data_providers/CsvDataProvider.ts
@@ -7,10 +7,10 @@ import {
   DataRow,
   dataRowToMetricData,
   dataRowsToMetricData,
+  parseCsv,
 } from "./data_provider_utils";
 import groupBy from "lodash/groupBy";
 import isEmpty from "lodash/isEmpty";
-import Papa from "papaparse";
 import { fetchText } from "./utils";
 
 export interface CsvDataProviderOptions {
@@ -82,7 +82,7 @@ export class CsvDataProvider extends SimpleMetricDataProviderBase {
       "We should have initialized fetchedText directly above or in the constructor"
     );
     const csvText = await this.fetchedText;
-    const csv = parseCsv(csvText, this.regionColumn);
+    const csv = parseCsv(csvText, [this.regionColumn]);
     assert(csv.length > 0, "CSV must not be empty.");
     const dataRowsByRegionId = groupBy(csv, (row) => row[this.regionColumn]);
     assert(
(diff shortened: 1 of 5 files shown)


packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -141,48 +141,3 @@ export class CsvDataProvider extends SimpleMetricDataProviderBase {
...
-        row[c] = null;
-      } else if (/^-?[\d.eE]+$/.test(v)) {
-        const num = Number.parseFloat(v.replace(/,/g, ""));
-        if (!Number.isNaN(num)) {
-          row[c] = num;
-        }
-      }
-    }
-  }
-}
```
mikelehen, line 188: These functions moved basically as-is to data_provider_utils.ts

packages/metrics/src/data_providers/data_provider_utils.test.ts
```diff
@@ -137,3 +139,48 @@ describe("dataRowToMetricData()", () => {
     ).toBe(null);
   });
 });
+
+describe("parseCsv() and generateCsv()", () => {
```
mikelehen, line 143: These tests aren't particularly exhaustive (especially the parsing), but since we're just wrapping Papaparse, I think that's hopefully okay. If there's anything in particular you think would be useful to test, let me know.

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -86,3 +91,122 @@ export function dataRowsToMetricData(
...
+ * to be written to a CSV.
+ */
+function generateCsvValue(value: unknown): string {
+  if (value === null || value === undefined) {
+    return "";
+  } else if (
+    typeof value === "string" &&
+    (value.includes(",") || value.includes('"'))
+  ) {
+    // Escape " with "" in strings.
```
smcclure17: nit: It wasn't immediately clear to me why we're adding an extra `"` here until I did a quick [google](https://stackoverflow.com/a/769820/14034347), maybe it's worth adding a comment on how double quotes are handled?

smcclure17 (commented):
looks great to me 🚢!  one small nit,

and just for clarity, I think the issue in the description is wrong and should instead link to https://github.com/covid-projections/act-now-template/issues/50

packages/metrics/src/data_providers/data_provider_utils.ts
mikelehen: Thanks. I updated the comment and also added handling for newlines within CSV values which I had omitted before. 🙏

mikelehen:
Ooops, yes! I linked to #50 in the wrong repo.  Thanks!

PR #440, opened 2022-11-22
Route MetricCatalog through SimpleMetricDataProviderBase.

mikelehen:
This was an oversight when I implemented `SimpleMetricDataProviderBase`.  This will make it easier to use `SimpleMetricDataProviderBase` to write a data provider that transforms multiple existing metrics, etc.


faihegberg (approved):
LGTM!

PR #441, opened 2022-11-22
Log and throw errors when CSV file contains invalid region IDs.

### LINKED ISSUE #424: Raise error / warning if importing CSV data with bad region IDs.

mikelehen:
* Gives MetricCatalog access to the RegionDb so we can use it for validation.
* Makes CsvMetricDataProvider validate the encountered regionIds. If any are invalid, they will be logged. If all are invalid, then it will throw an error.

Fixes #424


mikelehen:
I reworked this to be simpler (passing RegionDB directly to the data provider instead of going through the metric catalog).

PR #442, opened 2022-11-22
Log and throw errors when CSV file contains invalid region IDs.

mikelehen:
Makes CsvDataProvider validate the encountered regionIds against a `RegionDB` that must now be provided during construction.  If any region IDs are invalid, they will be logged. If all of them are invalid, then it will throw an error.


smcclure17 (approved):
nice! LGTM 🙏

PR #443, opened 2022-11-22
Props cleanup follow-ups

chasulin:
Misc. follow-ups to #415:

- Renames some prop interfaces that were missed in the last PR
- Removes circular dependencies by moving prop interfaces into `utils` files where needed
- Moves RegionSearchItem code out of RegionSearch and into its own folder (now SearchItem, since nothing region-specific is in the component)


packages/ui-components/src/components/MetricMultiProgressBar/MetricMultiProgressBar.tsx
```diff
@@ -44,6 +39,11 @@ export const MetricMultiProgressBar = ({
   );
 };
 
+export interface MetricProgressBarItem {
```
chasulin, line 42: just moved down in the file

packages/ui-components/src/components/MetricLegendThreshold/utils.ts
```diff
@@ -1,6 +1,16 @@
...
+export interface CategoryItem {
+  /** Category name (e.g. "High") */
+  name: string;
+  /** Category color */
+  color: string;
+  /** Description of the category */
+  description: string | undefined;
+  /** Formatted value of the threshold at the end of the current category */
+  endThreshold?: string;
+}
```
chasulin, line 13: moved from `MetricLegendThreshold.tsx` to avoid circ dependency

packages/ui-components/src/components/MetricCompareTable/utils.tsx
```diff
@@ -12,7 +12,16 @@ import {
...
+import { Region } from "@actnowcoalition/regions";
+
+export interface Row {
+  /** Unique ID for the row. */
+  rowId: string;
+  /** Region corresponding to this row. */
+  region: Region;
+  /** multiMetricDataStore instance to make metric data accessible to every cell */
+  multiMetricDataStore: MultiMetricDataStore;
+}
```
chasulin, line 24: moved from `MetricCompareTable.tsx` to avoid circ dependency

packages/ui-components/src/components/MetricMultiProgressBar/MetricMultiProgressBar.tsx
```diff
@@ -44,6 +39,11 @@ export const MetricMultiProgressBar = ({
...
+  currentValue: number;
+  label: string;
+}
+
```
chasulin, line 46: just moved down in the file

packages/ui-components/src/common/hooks/useHoveredDate.tsx
```diff
@@ -1,6 +1,6 @@
 import { useState } from "react";
 import { Timeseries, TimeseriesPoint } from "@actnowcoalition/metrics";
-import { ChartOverlayXProps } from "./ChartOverlayX";
+import { ChartOverlayXProps } from "../../components/ChartOverlayX";
```
mikelehen: This feels funky for code in `common` to import code from `components`.  I'd normally expect it to always go in the opposite direction.  While not strictly necessary, preserving this sort of architectural layering can help avoid accidental circular dependencies, etc.

In this case it looks like we're just doing it so we can share the `ChartOverlayXProps["onMouseMove"]` type definition.  Can we just duplicate it here instead to remove the import?

packages/ui-components/src/common/hooks/useHoveredPoint.ts
```diff
@@ -1,7 +1,10 @@
 import { useState } from "react";
 import isNumber from "lodash/isNumber";
 import { Timeseries, TimeseriesPoint } from "@actnowcoalition/metrics";
-import { ChartOverlayXYProps, HoveredPointInfo } from "./ChartOverlayXY";
+import {
+  ChartOverlayXYProps,
+  HoveredPointInfo,
+} from "../../components/ChartOverlayXY";
```
mikelehen: Similar feedback here.  For `HoveredPointInfo` maybe we can just move that interface into this file?

mikelehen (approved):
Small feedback about having code in `common` depend on code in `components` but otherwise looks good!

packages/ui-components/src/common/hooks/useHoveredPoint.ts
chasulin: hm moving the `HoveredPointInfo` interface into /hooks feels weird too..

which do you think is less weird -- having the `HoveredPointInfo` interface in `/hooks` and imported into `/components`, or keeping both hooks where they originally were (in `/components`)?

mikelehen: Good question.  Probably moving the hook back to components (sorry!).  Given the hook is specifically designed to be used in conjunction with the component (I think?), keeping them together makes sense to me.

chasulin: yeah i agree, this makes sense. back they go

mikelehen: 👍 🙏

PR #458, opened 2022-11-29
Import sorting

chasulin:
Sorts imports in `/packages` using [@trivago/prettier-plugin-sort-imports](https://github.com/trivago/prettier-plugin-sort-imports), using the following sort order:
- react
- third party imports
- @actnowcoalition
- local imports

### Example (SparkLine.tsx):

**Before sort:**

```
import React from "react";
import { Group } from "@visx/group";
import { Timeseries } from "@actnowcoalition/metrics";
import { BarChart } from "../BarChart";
import { LineChart } from "../LineChart";
import { scaleUtc, scaleLinear } from "@visx/scale";
import { useTheme } from "@mui/material";
```

**After sort:**

```
import React from "react";

import { useTheme } from "@mui/material";
import { Group } from "@visx/group";
import { scaleLinear, scaleUtc } from "@visx/scale";

import { Timeseries } from "@actnowcoalition/metrics";

import { BarChart } from "../BarChart";
import { LineChart } from "../LineChart";
```

### DIFF
--- a/packages/ui-components/src/components/BarChart/BarChart.stories.tsx
+++ b/packages/ui-components/src/components/BarChart/BarChart.stories.tsx
@@ -1,11 +1,12 @@
-import React from "react";
-import { ComponentStory, ComponentMeta } from "@storybook/react";
-import { scaleLinear, scaleUtc } from "@visx/scale";
-import { Group } from "@visx/group";
 import { assert } from "@actnowcoalition/assert";
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import { Group } from "@visx/group";
+import { scaleLinear, scaleUtc } from "@visx/scale";
+import React from "react";
+
+import { BarChart } from ".";
 import { appleStockTimeseries } from "../../stories/mockData";
 import { LineChart } from "../LineChart";
-import { BarChart } from ".";
 
 export default {
   title: "Charts/BarChart",
--- a/packages/ui-components/src/components/BarChart/BarChart.stories.tsx
+++ b/packages/ui-components/src/components/BarChart/BarChart.stories.tsx
@@ -1,7 +1,8 @@
+import React from "react";
+
 import { ComponentMeta, ComponentStory } from "@storybook/react";
 import { Group } from "@visx/group";
 import { scaleLinear, scaleUtc } from "@visx/scale";
-import React from "react";
 
 import { assert } from "@actnowcoalition/assert";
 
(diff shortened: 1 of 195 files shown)


packages/ui-components/src/components/BarChart/BarChart.stories.tsx
```diff
@@ -1,11 +1,12 @@
-import React from "react";
-import { ComponentStory, ComponentMeta } from "@storybook/react";
-import { scaleLinear, scaleUtc } from "@visx/scale";
-import { Group } from "@visx/group";
 import { assert } from "@actnowcoalition/assert";
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import { Group } from "@visx/group";
+import { scaleLinear, scaleUtc } from "@visx/scale";
+import React from "react";
```
chasulin: i cant figure out how to get imports without `@` (ie. `'react'`, `'lodash'`) sorted _above_ imports with @ (ie. `@visx/scale`) [using importOrder](https://github.com/covid-projections/act-now-packages/pull/458/files#diff-979a4a52db5287c891eac28a141bfa97bf8b0732ed5afae848d90956ef80e78dR1-R5)

1- does anyone agree/disagree with this being something that should be fixed?
2- does anyone who is better versed in regex know how to edit the `importOrder` array accordingly? 😁

right now im just using the array [in the docs example](https://github.com/trivago/prettier-plugin-sort-imports#importorder)

mikelehen: I don't know of any particular reason we'd want third-party imports without `@` to be sorted above third-party imports with `@`.  Either way seems fine to me.

I *do* think it would be nice to separate the `@actnowcolation/*` imports from third-party imports though, so here's the sort order I'd suggest:

```
  "importOrder": ["<THIRD_PARTY_MODULES>", "^@actnowcoalition/", "^[./]"]
```

If we *do* want to sort the non-`@` imports before the `@` imports, it would be a bit tricky because we will want to separate `@actnowcoalition` imports from other `@` imports, but I think we could do:
```
  "importOrder": ["<THIRD_PARTY_MODULES>", "^@(?!actnowcoalition/)", "^@actnowcoalition/", "^[./]"]
```

The `^@(?!actnowcoalition/)` pattern is using a negative lookahead to say that "actnowcoalition/" must not appear after the "@", so it'll match all the `@` imports except `@actnowcoalition/`.

Anyway, my preference would be to keep it simple and go with my first one:

```
  "importOrder": ["<THIRD_PARTY_MODULES>", "^@actnowcoalition/", "^[./]"]
```

But I don't care particularly strongly.  As long as the order is consistent, it doesn't matter that much what it is. 🤷

pnavarrc: I think that the order proposed by @mikelehen is good enough (third party, then `@actnowcoalition`, then local imports).

I kind of like having `react` first, then other 3rd party modules, then `@actnowcoalition` and then local imports, but I'm 100% fine with not having a special handling for react, I'm sure we will get used to this config in no time.

mikelehen (approved):
Looks good! Thanks for figuring this all out.

PR #459, opened 2022-11-30
Add default metric format options

### LINKED ISSUE #326: Improve how the y-axis tick labels are formatted
In some cases, the format of the metric values doesn't work well for the y-axis. In the chart below, we have `0.1` and `0.2` twice `http://localhost:3000/us/washington-wa`

Image

On the vaccination chart, we have the opposite problem - the number format doesn't need to have 1 decimal place on the y-axis, but it does in other places.

Image

One option is to set the format of the y-axis automatically with better defaults, or maybe to add that information to the metric.

faihegberg:
Closes https://github.com/covid-projections/act-now-packages/issues/326

This PR adds a default format option for each metric.

### DIFF
--- a/.changeset/empty-carrots-learn.md
+++ b/.changeset/empty-carrots-learn.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/ui-components": patch
+---
+
+Add default metric format options and decimal points mock data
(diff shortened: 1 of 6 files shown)


smcclure17 (approved):
I don't have the best context on this but I think it LGTM! 🚢

PR #460, opened 2022-11-30
Fix the initial sorting state of MetricCompareTable

### LINKED ISSUE #392: Compare Table: Ability to sort a column initially
Slightly related to #391 - We would like to provide a way to sort the table by a specific column and ordering direction without it breaking the ability to sort the table after.

pnavarrc:
Close https://github.com/covid-projections/act-now-packages/issues/392 - Fix MetricCompareTable so its sorting state can be set initially (and the the table remains sortable after) 

The bug was happening because:

- `MetricCompareTable` wasn't initializing the sorting state using the `sortColumnId` and `sortDirection` passed as props (but it was still passing them down to `CompareTable`).
- The `CompareTable` was sorting the table using the `sortColumnId` and `sortDirection` props (which were not connected to the sorting state of `MetricCompareTable`).

Since the `CompareTable` doesn't have a sorting state (or any sorting logic really), it doesn't make sense that it will sort the rows. I removed `sortColumnId` and `sortDirection` from `CompareTableProps` and added them to `MetricCompareTable`.

**Storybook**
- [CompareTable](https://act-now-packages--pr460-pablo-compare-table-fz3lmhdc.web.app/storybook/index.html?path=/story/table-comparetable--example)
- [MetricCompareTable - Default Sort Order](https://act-now-packages--pr460-pablo-compare-table-fz3lmhdc.web.app/storybook/index.html?path=/story/metrics-metriccomparetable--default-sort-order) 
- [MetricCompareTable - Initial Sort State](https://act-now-packages--pr460-pablo-compare-table-fz3lmhdc.web.app/storybook/index.html?path=/story/metrics-metriccomparetable--initial-sort-order)

I also moved some functions to `components/CompareTable/utils.ts` since they are not being used in the `CompareTable` component anymore. I will cleanup a bit more (get rid of `interfaces.ts`) on a follow up PR.

### DIFF
--- a/packages/ui-components/src/components/CompareTable/CompareTable.stories.tsx
+++ b/packages/ui-components/src/components/CompareTable/CompareTable.stories.tsx
@@ -14,6 +14,7 @@ import {
   TableCell,
   TableContainer,
   compare,
+  sortTableRows,
 } from ".";
 
 export default {
@@ -136,27 +137,23 @@ const StatefulCompareTable = ({ rows }: { rows: RowItem[] }) => {
     },
   ];
 
-  const [sortColumnId, setSortColumnId] = useState(columns[0].columnId);
-  const [sortDirection, setSortDirection] = useState(SortDirection.ASC);
+  const [sortState, setSortState] = useState({
+    sortColumnId: columns[0].columnId,
+    sortDirection: SortDirection.ASC,
+  });
+  const { sortColumnId, sortDirection } = sortState;
 
   const onClickSort = (direction: SortDirection, columnId: string) => {
-    setSortColumnId(columnId);
-    setSortDirection(direction);
+    setSortState({ sortColumnId: columnId, sortDirection: direction });
   };
 
-  return (
-    <CompareTable
-      rows={rows}
-      columns={columns}
-      sortColumnId={sortColumnId}
-      sortDirection={sortDirection}
-    />
-  );
+  const sortedRows = sortTableRows(rows, columns, sortState);
+  return <CompareTable rows={sortedRows} columns={columns} />;
 };
 
 export const Example = () => {
   return (
-    <TableContainer sx={{ maxWidth: 400, height: 600 }}>
+    <TableContainer sx={{ maxWidth: 700, height: 600 }}>
       <StatefulCompareTable rows={rows} />
     </TableContainer>
   );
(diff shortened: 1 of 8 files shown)


packages/ui-components/src/components/CompareTable/CompareTable.stories.tsx
```diff
@@ -136,27 +137,23 @@ const StatefulCompareTable = ({ rows }: { rows: RowItem[] }) => {
     },
   ];
 
-  const [sortColumnId, setSortColumnId] = useState(columns[0].columnId);
-  const [sortDirection, setSortDirection] = useState(SortDirection.ASC);
+  const [sortState, setSortState] = useState({
+    sortColumnId: columns[0].columnId,
+    sortDirection: SortDirection.ASC,
+  });
+  const { sortColumnId, sortDirection } = sortState;
```
pnavarrc, line 144: Decided to merge the state variables into a single object, since changing any of them will force a re-rendering of the table anyways.

packages/ui-components/src/components/CompareTable/CompareTable.stories.tsx
```diff
@@ -136,27 +137,23 @@ const StatefulCompareTable = ({ rows }: { rows: RowItem[] }) => {
...
-    />
-  );
+  const sortedRows = sortTableRows(rows, columns, sortState);
+  return <CompareTable rows={sortedRows} columns={columns} />;
 };
 
 export const Example = () => {
   return (
-    <TableContainer sx={{ maxWidth: 400, height: 600 }}>
+    <TableContainer sx={{ maxWidth: 700, height: 600 }}>
```
pnavarrc, line 156: It was narrower to make it easier to test horizontal scrolling, but it was a bit annoying to test sorting.

packages/ui-components/src/components/CompareTable/CompareTable.tsx
```diff
@@ -1,13 +1,7 @@
...
-import {
-  CompareTableProps,
-  SortDirection,
-  Table,
-  TableBody,
-  TableHead,
-  TableRow,
-} from ".";
+import { Table, TableBody, TableHead, TableRow } from "./CompareTable.style";
+import { CompareTableProps } from "./interfaces";
```
pnavarrc, line 4: Removed the imports from `"."` to prevent potential circular dependencies (except in storybook files, for those I want to check that all the things we use are exported by the `index.ts` file)

faihegberg (approved):
LGTM!

PR #461, opened 2022-11-30
Fix #417: Render Skeleton component in charts while data is loading to avoid layout shift.

### LINKED ISSUE #417: Charts - Better loading state to avoid layout shift
See:
* [MetricLineChart](https://act-now-packages.web.app/storybook/index.html?path=/story/charts-metriclinechart--loading-delay&globals=outline:true;backgrounds.grid:false)
* [MetricLineThresholdChart](https://act-now-packages.web.app/storybook/index.html?path=/story/charts-metriclinethresholdchart--loading-delay&globals=outline:true;backgrounds.grid:false)
* [MetricSeriesChart](https://act-now-packages.web.app/storybook/index.html?path=/story/charts-metricserieschart--loading-delay&globals=outline:true;backgrounds.grid:false) - It looks to me like it has placeholder space but it's not quite the right size (missing padding or something?) so there's still a slight shift when it loads.
* [MetricSparklines](https://act-now-packages.web.app/storybook/index.html?path=/story/metrics-metricsparklines--loading-delay&globals=outline:true;backgrounds.grid:false) - It looks to me like it has placeholder space but it's not quite the right size (missing padding or something?) so there's still a slight shift when it loads.

mikelehen:
See:
* https://act-now-packages--pr461-mikelehen-chart-load-s1kw1ydm.web.app/storybook/index.html?path=/story/charts-metriclinechart--loading-delay
* https://act-now-packages--pr461-mikelehen-chart-load-s1kw1ydm.web.app/storybook/index.html?path=/story/charts-metriclinethresholdchart--loading-delay
* https://act-now-packages--pr461-mikelehen-chart-load-s1kw1ydm.web.app/storybook/index.html?path=/story/charts-metricserieschart--loading-delay
* https://act-now-packages--pr461-mikelehen-chart-load-s1kw1ydm.web.app/storybook/index.html?path=/story/metrics-metricsparklines--loading-delay

### DIFF
--- a/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
+++ b/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
@@ -11,6 +11,7 @@ import { PointMarker } from "../PointMarker";
 import { BaseChartProps } from "../../common/utils/charts";
 import { Metric } from "@actnowcoalition/metrics";
 import { Region } from "@actnowcoalition/regions";
+import { Skeleton } from "@mui/material";
 
 export interface MetricLineChartProps extends BaseChartProps {
   metric: Metric | string;
@@ -37,7 +38,7 @@ export const MetricLineChart = ({
     useHoveredDate(timeseries);
 
   if (!data || !timeseries?.hasData()) {
-    return null;
+    return <Skeleton variant="rectangular" width={width} height={height} />;
   }
 
   const chartHeight = height - marginTop - marginBottom;
@@ -56,7 +57,7 @@ export const MetricLineChart = ({
   });
 
   return (
-    <svg width={width} height={height}>
+    <svg width={width} height={height} style={{ display: "block" }}>
       <Group left={marginLeft} top={marginTop}>
         <AxesTimeseries
           height={chartHeight}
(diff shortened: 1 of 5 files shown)


packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
```diff
@@ -56,7 +57,7 @@ export const MetricLineChart = ({
   });
 
   return (
-    <svg width={width} height={height}>
+    <svg width={width} height={height} style={{ display: "block" }}>
```
mikelehen, line 63: By default svg elements are inline-block which causes them to be rendered like text with some extra space below them based on your line-height, etc.  Displaying as "block" makes sure the exact size used by the svg will match the width/height passed (which we use in the skeleton component) and so we won't get layout shift.

chasulin (approved):
✅

not blocking, but i feel like the grey box implementation is a bit dark + makes the transition between loading state and chart a little jarring. maybe a lighter grey would make the things look smoother + make the transition more subtle?

mikelehen:
👍  I'll play with centralizing our loading state color and tweaking the shade in another PR.

PR #462, opened 2022-11-30
Fix #422: Make MetricMultiProgressBar render Skeleton while data is loading to avoid layout shift.

### LINKED ISSUE #422: MetricMultiProgressBar - better loading state
Currently there's a layout shift when it renders.

On error I'm not sure what we should do.  It'll probably usually be rendered in conjunction with a value and so rendering nothing might be fine / expected.

mikelehen:
See https://act-now-packages--pr462-mikelehen-progressba-64no4ucj.web.app/storybook/index.html?path=/story/metrics-metricmultiprogressbar--loading-delay

### DIFF
--- a/packages/ui-components/src/components/MetricMultiProgressBar/MetricMultiProgressBar.tsx
+++ b/packages/ui-components/src/components/MetricMultiProgressBar/MetricMultiProgressBar.tsx
@@ -4,6 +4,8 @@ import { Metric, MultiMetricDataStore } from "@actnowcoalition/metrics";
 import {
   MultiProgressBar,
   BaseMultiProgressBarProps,
+  DEFAULT_WIDTH,
+  DEFAULT_HEIGHT,
 } from "../MultiProgressBar";
 import { useDataForMetrics } from "../../common/hooks";
 import { Skeleton } from "@mui/material";
@@ -20,8 +22,8 @@ export interface MetricMultiProgressBarProps extends BaseMultiProgressBarProps {
 export const MetricMultiProgressBar = ({
   region,
   metrics,
-  width = 100,
-  height = 16,
+  width = DEFAULT_WIDTH,
+  height = DEFAULT_HEIGHT,
   ...otherProgressBarProps
 }: MetricMultiProgressBarProps) => {
   const { data } = useDataForMetrics(region, metrics, false);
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/components/MetricMultiProgressBar/MetricMultiProgressBar.tsx
```diff
@@ -19,12 +20,14 @@ export interface MetricMultiProgressBarProps extends BaseMultiProgressBarProps {
 export const MetricMultiProgressBar = ({
   region,
   metrics,
+  width = 100,
+  height = 16,
```
mikelehen: These are the same defaults used by the ProgressBar component, but I'm now specifying them here so we can use them to size the Skeleton component.

packages/ui-components/src/components/MultiProgressBar/MultiProgressBar.tsx
```diff
@@ -54,7 +54,13 @@ export const MultiProgressBar = <T,>({
   );
 
   return (
-    <svg width={width} height={height} aria-labelledby={titleId} role="meter">
+    <svg
+      width={width}
+      height={height}
+      aria-labelledby={titleId}
+      role="meter"
+      style={{ display: "block" }}
```
mikelehen, line 67: By default svg elements are inline-block which causes them to be rendered like text with some extra space below them based on your line-height, etc.  Displaying as "block" makes sure the exact size used by the svg will match the width/height passed (which we use in the skeleton component) and so we won't get layout shift.

packages/ui-components/src/components/MetricMultiProgressBar/MetricMultiProgressBar.tsx
chasulin: nit - should we export these defaults from MultiProgressBar to ensure they stay in sync?

chasulin (approved):
looks good! [one small comment](https://github.com/covid-projections/act-now-packages/pull/462#discussion_r1036376382)

packages/ui-components/src/components/MetricMultiProgressBar/MetricMultiProgressBar.tsx
mikelehen: Good flag, done!

PR #463, opened 2022-11-30
Fix #421: Render placeholder text in MetricOverview while data is loading to avoid layout shift.

### LINKED ISSUE #421: Better loading & error states: MetricOverview, MetricScoreOverview, MetricValue
When loading there is some layout shift. See:
* https://act-now-packages.web.app/storybook/index.html?path=/story/metrics-metricoverview--loading-delay&globals=outline:true;backgrounds.grid:false
* https://act-now-packages.web.app/storybook/index.html?path=/story/metrics-metricscoreoverview--loading-delay&globals=outline:true;backgrounds.grid:false
* https://act-now-packages.web.app/storybook/index.html?path=/story/metrics-metricvalue--loading-delay&globals=outline:true;backgrounds.grid:false

When there's an error, parts just don't render.  I think we should render "--" or similar instead.

mikelehen:
See:
* https://act-now-packages--pr463-mikelehen-metric-ove-844e7efn.web.app/storybook/index.html?path=/story/metrics-metricvalue--loading-delay
* https://act-now-packages--pr463-mikelehen-metric-ove-844e7efn.web.app/storybook/index.html?path=/story/metrics-metricoverview--loading-delay
* https://act-now-packages--pr463-mikelehen-metric-ove-844e7efn.web.app/storybook/index.html?path=/story/metrics-metricscoreoverview--loading-delay

### DIFF
--- a/.changeset/brave-files-heal.md
+++ b/.changeset/brave-files-heal.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/ui-components": patch
+---
+
+Fix #421: Render placeholder text in MetricOverview while waiting for data to load to avoid layout shift.
--- a/packages/ui-components/src/components/MetricValue/MetricValue.tsx
+++ b/packages/ui-components/src/components/MetricValue/MetricValue.tsx
@@ -29,6 +29,10 @@ export const MetricValue = ({
 
   const { data, error } = useData(region, metric, /*includeTimeseries=*/ false);
 
+  // If there's an error we render "---" to match when there's no data.
+  // While we are waiting for data to load we render "\u00A0" (a non-breaking
+  // space) just so the height renders correctly and we don't get a layout shift
+  // when the data is available.
   const formattedValue = error
     ? "---"
     : !data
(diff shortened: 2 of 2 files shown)


chasulin (approved):
nice!

chasulin:
one thing- for the sake of systematizing how we handle loading states // settling on some sort of convention, would you say that the intention is to use `\u00A0` as loading state for text, and use a `Skeleton` as a loading state for components?

mikelehen:
I think so?  For the maps I didn't use Skeleton because I didn't know the size ahead-of-time.  So far `MetricValue` is the only component where I've needed to render placeholder text and I kinda' arbitrarily chose `\u00A0` to avoid the layout shift. If somebody were to propose an alternative approach (like wrapping the text in a fixed-height container or something), I would be open to other ideas.  

But in short, yes, for now I think the convention will be to use `Skeleton` when that works / makes sense, and we can use `\u0A00` as needed to avoid layout shifts due to loading text. 🤷

Issue #454, opened 2022-11-29
CsvDataProvider parses CSV repeatedly while cache is populating.


mikelehen:
I noticed after I added region ID error handling to Fai's hackathon project the console showed:

![Image](https://user-images.githubusercontent.com/206364/204586784-25d23e67-4550-435c-ae4c-54d5fcf25549.png)

I think that means we are parsing the CSV content 702 times during page load. 🤯 

I think we need to rework the way [populateCache](https://github.com/covid-projections/act-now-packages/blob/c771c22c7ea50401a70d02ada204337af3c4d36f/packages/metrics/src/data_providers/CsvDataProvider.ts#L82) is called / works so that we have a single cached promise that represents the entire fetching / parsing process (rather than right now where fetchedText just represents the fetching process).

Issue #377, opened 2022-11-08
Implement JSON Data Provider


pnavarrc:
Can we get example JSON from @chasulin  and @faihegberg  (and anybody else that ran into this?)

smcclure17:
Not examples from real use cases, but possible forms to consider could maybe be: 

Record style, e.g:
```json
{
  "data": [
    {"region": "USA", "metric1:" "value"},
    {"region": "CAN", "metric1:" "value"},
    ...
  ]
}
```

Region/entity indexed (like the output of lodash's `keyBy()`):
```json
{
  "USA": {"metric1": "value"}, 
  "CAN": {"metric2": "value"},
  ...
}
```

Metric indexed:
```JSON
{
  "metric1": {"USA":"value", "CAN":"value"}
  "metric2": {"USA":"value2", "CAN":"value2"}
  ...
}
```

Maybe there's a world in which we can pass an "example" or "template" structure to the provider, and the provider could use that template to parse the actual file? Not sure what that would take to implement...

pnavarrc:
I think that something that mimics the structure of the CSV could work (so record style?)

```json
{
  "data": [
    { "region": "CAN", "date": "2020-01-01", "metric1": 1.23, "metric2": 1.34 },
  ]
}
```

I'm not sure if we need the "data" key in the JSON (it could just be an array), but we could make the provider configurable so we can tell how to access the array. Open to more ideas of course!

In general I like to organize data using the principles in [Tidy Data](https://cran.r-project.org/web/packages/tidyr/vignettes/tidy-data.html) (basically the CSV format that we use for the `CSVDataProvider`). 

- https://cran.r-project.org/web/packages/tidyr/vignettes/tidy-data.html
- https://www.jstatsoft.org/article/view/v059i10

smcclure17:
That makes sense--I agree, I'll start with this. 

Thinking out loud, Python Pandas has a [JSON loader with an "orient" option](https://pandas.pydata.org/docs/reference/api/pandas.read_json.html) that can accept multiple formats--maybe we could (eventually) model this to be similar? I think it's best to prioritize one or two formats to start with, though. And, TBH, I've always found importing JSON into Pandas a little finicky

smcclure17:
#465

pnavarrc:
[...]

PR #465, opened 2022-11-30
JSON records provider

smcclure17:
addresses #377 and #454 in the process.

 I think there's a bit of an open question of what/how many JSON formats we want to handle here. I think ultimately the answer is "as many as we can", but it isn't a big priority at the moment, and we don't know exactly what would be useful to implement. I can make this a topic at standup, but I think it makes sense to start with this for now. 

Restructures the CSV provider slightly by creating an abstract base class, as there was a fair amount of code overlap between this JSON provider and the existing CSV provider.

See example of using JSON files here: https://github.com/covid-projections/fai-hackathon-nov-2022/pull/23


packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -71,17 +60,16 @@ export class CsvDataProvider extends SimpleMetricDataProviderBase {
...
     this.regionDb = options.regionDb;
     this.regionColumn = options.regionColumn;
-    this.dateColumn = options.dateColumn;
     this.url = options.url;
     this.fetchedText = options.csvText
       ? Promise.resolve(options.csvText)
       : undefined;
   }
 
-  private async populateCache(): Promise<void> {
```
smcclure17, line 84: I'm not sure how important it is to have some of these methods be private (though it's definitely preferred). Using an abstract base class makes that impossible as far as I'm aware 😕

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -94,73 +82,15 @@ export class CsvDataProvider extends SimpleMetricDataProviderBase {
...
-        throw new Error(
-          `Failed to parse CSV ${url}: All region IDs were invalid.`
-        );
-      }
-    }
-
     this.dataRowsByRegionId = dataRowsByRegionId;
   }
 
-  async fetchDataForRegionAndMetric(
```
smcclure17, line 125: I've moved this to an abstract base class `DataRowMetricProviderBase` (naming?) as this would have been copied over 1:1 to the JSON provider

packages/metrics/src/data_providers/DataRowMetricProviderBase.ts
```diff
@@ -0,0 +1,85 @@
...
+  DataRow,
+  dataRowsToMetricData,
+  dataRowToMetricData,
+} from "./data_provider_utils";
+import { SimpleMetricDataProviderBase } from "./SimpleMetricDataProviderBase";
+
+export interface DataRowMetricProviderBaseOptions {
+  /**
+   * Name of column containing valid ISO 8601 date-time values.
+   * Required if the JSON file contains timeseries data, else it should not be specified.
```
mikelehen: This says JSON but it's meant to be shared between CSV and JSON, right?

TBH I think I would maybe make `CsvDataProviderOptions` and `JsonDataProviderOptions` *not* extend this interface and be 100% standalone.  That will:
1. Let us name / document the options in a way that is maximally user-friendly (refer to CSV columns for CSV and JSON fields for JSON or whatever).
2. Result in simpler auto-generated documentation, I think (the user can go to one page and see all the options rather than having to look at the base options separately).

packages/metrics/src/data_providers/CsvDataProvider.ts
mikelehen: I think you can mark it `protected`, though see my other comment about using inheritance.

packages/metrics/src/data_providers/CsvDataProvider.ts
mikelehen: Would it be possible to use a shared helper function or class instead?  In general, while inheritance can be a very useful tool in some cases, it can also make the code harder to read, especially if the parent class is basically a collection of useful code as opposed to a well-defined class that can be easily described / reasoned about on its own. The fact that there isn't an obvious name for the parent class suggests this is more of the former case. 

Using explicit helper functions / classes can often be a simpler solution.

Alternatively, I would actually say this is a small enough chunk of code that repeating it once wouldn't be the end of the world.  I think we can actually drop the `metricData.dropTimeseries()` call.  I added code to MetricCatalog to do that automatically for all providers.

packages/metrics/src/data_providers/DataRowMetricProviderBase.ts
```diff
@@ -0,0 +1,85 @@
...
+   */
+  abstract populateCache(): Promise<void>;
+
+  async fetchDataForRegionAndMetric(
+    region: Region,
+    metric: Metric,
+    includeTimeseries: boolean
+  ): Promise<MetricData<unknown>> {
+    if (isEmpty(this.dataRowsByRegionId)) {
+      await this.populateCache();
```
mikelehen: Sorry for not bringing this up before you started on this PR, but there's actually an issue with the CSV data provider where it will parse the CSV file repeatedly while we're waiting for the network fetch to complete. https://github.com/covid-projections/act-now-packages/issues/454 This is 100% my fault. I fixed it to not repeat the network fetches a while ago, but did it in a naive way that didn't prevent us from re-parsing the CSV content. 🤦 

Especially if we're going to be sharing code, it might make sense to tackle #454 first so that we don't repeat the same mistake in the JSON provider.

mikelehen (changes requested):
This looks good but I have a couple semi-disruptive comments (about inheritance and fixing a preexisting caching issue with the CSV data provider). Sorry!

packages/metrics/src/data_providers/DataRowMetricProviderBase.ts
smcclure17: Np! In 89f6f3158d7ffeb0af13e834df338e227bb230bb I've reworked it to a similar structure to how the CSV fetching promises are handled, which does mean that I slightly changed the way the cache gets populated. LMK if this doesn't feel like a clean/good solution and I can put some more thought into it

packages/metrics/src/data_providers/CsvDataProvider.ts
smcclure17: Yeah, that reasoning makes sense! I've restructured this to a shared util function in `data_provider_utils.ts`

Thanks!

packages/metrics/package.json
```diff
@@ -35,6 +35,7 @@
     "@actnowcoalition/time-utils": "^0.1.0",
     "@types/papaparse": "^5.3.3",
     "delay": "^5.0.0",
+    "encoding": "^0.1.13",
```
smcclure17: When testing in https://github.com/covid-projections/fai-hackathon-nov-2022/pull/23 I was getting an error of something like `module encoding not found`. Adding it here fixed it, but I'm unsure if this should be necessary?

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -79,7 +78,7 @@ export class CsvDataProvider extends SimpleMetricDataProviderBase {
       : undefined;
   }
 
-  async populateCache(): Promise<void> {
+  private async getDataForCache(): Promise<{ [regionId: string]: DataRow[] }> {
     if (this.url) {
       // We might already be fetching the CSV, in which case we can just wait on
       // the existing promise.
```
mikelehen: I think this comment is kinda' wrong / misleading now (we're no longer at risk of hitting this code when we're already fetching the CSV) and also we don't need the `fetchedText` promise anymore. I think we could remove it and change the constructor to populate `this.dataRowsByRegionId` in the case that you pass in test csv data to use.

packages/metrics/package.json
mikelehen: Hrm.  Strange.  Any idea what triggered it?  It doesn't seem like we are actually accessing any functionality from it.  It's possible that some module requires this as a peer dependency or something, except you're not adding any new modules in this PR, so I'm confused.  To be clear was this error related to this PR in any way?

packages/metrics/src/data_providers/JsonDataProvider.test.ts
```diff
@@ -0,0 +1,143 @@
...
+  id: "metric",
+  dataReference: { providerId: PROVIDER_ID, column: "cool_metric" },
+});
+
+/**
+ * Create a JsonDataProvider from passed arguments and fetch data for test region and metric.
+ *
+ * @param data JSON data to be imported.
+ * @param includeTimeseries Whether to include timeseries.
+ * @param dateCol Name of date column.
```
mikelehen: Missing documentation for `metric`

packages/metrics/src/data_providers/JsonDataProvider.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    region: "36",
+    cool_metric: 150,
+  },
+  {
+    region: "12",
+    cool_metric: null,
+  },
+];
+
+const jsonTimeseries = [
```
mikelehen: nit: `mockJson` and `jsonTimeseries` are used similarly but one is labeled as `mock` and one is not.  I'd probably drop "mock" entirely and use "test" in both (similar to `testMetric`), but don't feel strongly.

packages/metrics/src/data_providers/JsonDataProvider.ts
```diff
@@ -0,0 +1,113 @@
...
+   * incoming data.
+   */
+  regionDb: RegionDB;
+  /** Name of column containing valid Region IDs. */
+  regionColumn: string;
+  /**
+   * Name of column containing valid ISO 8601 date-time values.
+   * Required if the CSV contains timeseries data, else it should not be specified.
+   */
+  dateColumn?: string;
```
mikelehen: nit: Does it make sense to use "column" when referring to JSON data?  I think I'd expect "property" or "field", but could be convinced otherwise if we think there's value in making the providers use similar terminology. 🤷

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -210,3 +212,86 @@ function getAllColumnsFromRows(rows: DataRow[]): string[] {
...
+ * @param dateKey Name of key containing date values.
+ * @returns MetricData for the provided region and metric.
+ */
+export async function getMetricDataFromDataRows(
+  dataRowsByRegionId: { [regionId: string]: DataRow[] },
+  region: Region,
+  metric: Metric,
+  dateKey?: string
+): Promise<MetricData<unknown>> {
+  const metricKey = metric.dataReference?.column;
```
mikelehen: nit: Here again I would expect to use `field` or `property` for JSON data.  So consider passing `metricKey` in instead of `metric`.  That said, again, if you feel like there's value in keeping this uniform between using CSV and JSON data, I could be talked into that.

packages/metrics/package.json
smcclure17: Not entirely sure what causes it atm 😕, but it does seem to be specific to this PR. Running the template on the published `metrics` package doesn't result in the same issue.

FWIW, this is the error I get without adding the `encoding` dependency:

image

packages/metrics/src/data_providers/JsonDataProvider.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      dataReference: {
+        providerId: PROVIDER_ID,
+        /* missing column specifier. */
+      },
+    });
+    expect(async () =>
+      testFetchingJsonData(
+        mockJson,
+        /*includeTimeseries=*/ true,
+        /*datColumn=*/ "date",
```
mikelehen: dateColumn

packages/metrics/src/data_providers/JsonDataProvider.ts
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * The regions that this JSON file contains data for. Used for validating the
+   * incoming data.
+   */
+  regionDb: RegionDB;
+  /** Name of column containing valid Region IDs. */
+  regionColumn: string;
+  /**
+   * Name of column containing valid ISO 8601 date-time values.
+   * Required if the CSV contains timeseries data, else it should not be specified.
```
mikelehen: CSV => JSON

packages/metrics/src/data_providers/CsvDataProvider.ts
smcclure17: Ah, yeah true--updated this in 74397fd2b9f11939baee1056078ffe6f074c5313

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -210,3 +212,86 @@ function getAllColumnsFromRows(rows: DataRow[]): string[] {
...
+/**
+ * Helper to group DataRows by region ID key and validate the region IDs against a regionDb.
+ *
+ * @param dataRows Data rows to group and validate regions for.
+ * @param regionDb RegionDB to validate against.
+ * @param regionKey Region column name to use for grouping and validation.
+ * @param url URL to use for error messages.
+ * @returns
+ */
+export function groupAndValidateRegionIds(
```
mikelehen: Nice helper function! I'd maybe tweak the name to `groupAndValidateRowsByRegionId`

packages/metrics/src/data_providers/data_provider_utils.ts
```diff
@@ -210,3 +212,86 @@ function getAllColumnsFromRows(rows: DataRow[]): string[] {
...
+  dataRows: DataRow[],
+  regionDb: RegionDB,
+  regionKey: string,
+  url?: string
+) {
```
mikelehen: Can you add the return type for readability?

packages/metrics/src/data_providers/CsvDataProvider.ts
```diff
@@ -76,89 +76,47 @@ export class CsvDataProvider extends SimpleMetricDataProviderBase {
...
+  ): Promise<{ [regionId: string]: DataRow[] }> {
+    if (!csvText) {
+      assert(this.url, "URL or csvText must be provided to populate cache.");
+      csvText = await this.fetchCsvText();
     }
     assert(
-      this.fetchedText,
+      csvText,
       "We should have initialized fetchedText directly above or in the constructor"
     );
```
mikelehen: Is this assert redundant with the above if-check now?  Will TypeScript let us remove it?

mikelehen (commented):
Basically looks good but I want to look at the `encoding` dependency before approving.  I'm confused.

packages/metrics/package.json
mikelehen: So I'm seeing a warning in the console when I build act-now-packages from `develop` and `yarn link` that into fai-hackathon-2022, so I think maybe this is related to `yarn link` but not your specific changes.

```
warn  - ../act-now-packages/packages/metrics/node_modules/node-fetch/lib/index.js
Module not found: Can't resolve 'encoding' in '/Users/mikelehen/src/act-now-packages/packages/metrics/node_modules/node-fetch/lib'
```

Based on what I'm seeing, node-fetch has an optional peer dependency on the encoding module: https://github.com/node-fetch/node-fetch/blob/1ef4b560a17e644a02a3bfdea7631ffeee578b35/package.json#L47

So it will use it if it's there, but it shouldn't fail when it's not there.

My suspicion is that this is a bug / quirk with `yarn link` and if you just merge your changes without adding `encoding` as a dependency and publish the packages, the published packages will work fine.  That said, if that means `yarn link` won't work for us going forward, maybe we should just explicitly add `encoding` even though we don't need it.  I don't really like bloating our dependencies needlessly though, so I'm kinda' on the fence.  Will leave it up to you!

As an aside, the latest versions of node-fetch don't [seem](https://github.com/node-fetch/node-fetch/blob/main/package.json) to have the `encoding` peer dependency, but last I checked we couldn't use the latest version because it only supports EcmaScript style imports, not CommonJS, which doesn't work with our next.js config, so upgrading probably isn't an option. :-/

mikelehen (approved):

packages/metrics/src/data_providers/data_provider_utils.ts
smcclure17: Yeah, this makes sense/I agree we should use format-specific language 

I've renamed the affected fields and moved this logic as you've suggested 🙏

packages/metrics/package.json
smcclure17: Ahh, interesting 🤔 thanks for digging into this! 🙏 

I agree with not adding dependencies unless strictly necessary, so let's try this without adding the encoding dependency (I've removed it in 7d4c89b0c3504b492d47e16fcbff259e65c2569e), then later if we feel like it's necessary/helpful we can reconsider if need be. 

Thanks again for looking into this

PR #467, opened 2022-12-01
Add disabledBackground theme color and use for loading states.

mikelehen:
Uses `theme.palette.action.disabledBackground` for our loading states (and missing map regions) and updates the color from `#EEE` to the slightly brighter `#F8F8F8`.

### DIFF
--- a/packages/ui-components/src/components/WorldMap/WorldMap.style.ts
+++ b/packages/ui-components/src/components/WorldMap/WorldMap.style.ts
@@ -1,5 +1,3 @@
-import { Box } from "@mui/material";
-
 import { styled } from "../../styles";
 
 export const colorDisputedAreas = "#aaa";
@@ -166,10 +164,4 @@ export const DisputedBorderPath = styled("path")`
   &.Western_Sahara {
     ${styleSolid};
   }
-`;
-
-export const ErrorMessage = styled(Box)`
-  display: grid;
-  place-items: center;
-  background-color: #eee;
 `;
--- a/packages/ui-components/src/styles/theme/components.tsx
+++ b/packages/ui-components/src/styles/theme/components.tsx
@@ -185,6 +185,13 @@ const components: ThemeOptions["components"] = {
       },
     },
   },
+  MuiSkeleton: {
+    styleOverrides: {
+      root: ({ theme }) => ({
+        backgroundColor: theme.palette.action.disabledBackground,
+      }),
+    },
+  },
 };
 
 export default components;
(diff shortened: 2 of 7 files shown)


packages/ui-components/src/components/WorldMap/WorldMap.style.ts
```diff
@@ -167,9 +165,3 @@ export const DisputedBorderPath = styled("path")`
     ${styleSolid};
   }
 `;
-
-export const ErrorMessage = styled(Box)`
```
mikelehen, line 171: When I was searching for things using color `#eee` I found this and it's not used, so I deleted it.

packages/ui-components/src/styles/theme/components.tsx
```diff
@@ -185,6 +185,13 @@ const components: ThemeOptions["components"] = {
       },
     },
   },
+  MuiSkeleton: {
+    styleOverrides: {
+      root: ({ theme }) => ({
+        backgroundColor: theme.palette.action.disabledBackground,
```
mikelehen, line 191: Updating `<Skeleton>` to use the same `disabledBackground` color...

packages/ui-components/src/styles/theme/palette.ts
```diff
@@ -15,6 +15,7 @@ const colors = {
     default: "#5f6c7233", // TODO - Josh to translate to hex in figma
   },
   action: {
+    disabledBackground: "#F8F8F8",
```
mikelehen, line 18: So:
1)  I don't really follow the naming scheme between this `colors` object and the MUI theme naming scheme.  I'm not sure if the `colors` names are supposed to match what Josh used in Figma or something, but I just named this the same as the MUI theme color (palette.action.disabledBackground).  If this doesn't seem right, I'm happy to change it.
2) Chelsi [mentioned](https://github.com/covid-projections/act-now-packages/pull/461#pullrequestreview-1199899110) the loading skeleton color seemed a bit dark, so I brightened it a tad from `#EEEEEE` to `#F8F8F8` although I think part of why it looks dark is because the opacity animates from 0 to 100%.  We could tweak it more, but I think this is probably fine for now. 🤷

chasulin: re: # 1 - hmm yeah, originally `colors` was supposed to match whats in figma. adding disabledBackground here makes sense to me though

chasulin: re: # 2 - this lightened grey looks better to me!

mikelehen: Another option would be to not include it in `colors` and just specify it directly in the theme below.  Let me know if that feels any better. 🤷

chasulin (approved):

PR #471, opened 2022-12-01
Fixes #371: renames renderTooltip to getTooltip

### LINKED ISSUE #371: Rename renderTooltip => getTooltip
Since it's implemented as a callback I assumed it would only be called when the tooltip needs to be rendered (i.e. when you hover over a country) but it looks like we call `renderTooltip` for every country in advance, before hovering which isn't ideal from a performance perspective.  Can we change it to be on-demand?  Or no?

Same is probably true for the other maps, but I haven't tried.

chasulin:
Fixes #371

### DIFF
--- a/.changeset/smooth-jars-attend.md
+++ b/.changeset/smooth-jars-attend.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/ui-components": minor
+---
+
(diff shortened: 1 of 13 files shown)


mikelehen (approved):
Don't forget the changeset!

Issue #413, opened 2022-11-17
Chart: Build MultiRegionMultiMetricChart (explore / trends)


mikelehen:
[Figma mock](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=100%3A5329&t=QXxopbSRoEZ2BwMa-1):

![Image](https://user-images.githubusercontent.com/206364/202799997-1ce45285-829a-4c3b-b9f8-943f1dfd6fc5.png)

Note that Josh said:
> Pablo mentioned to me a few weeks ago he’s interested in opening this design back up for exploration some, with the goal to solve the limited space problem in the area containing location names, perhaps with a legend. I’m still waiting on him for some more specifics before I start exploring it, but FYI

He also said:

> you can see another unsolved problem in this component being temporarily/partially addressed here in the [Climate Act Now Finland mockups](https://www.figma.com/file/v3nLVBInPHpP6tJPMW95Nk/Climate-Act-Now-Finland?node-id=48%3A25203&t=LaKpVCCpJQh9T7pu-1), as I’m attempting to include both the short and long metric names in the dropdown select. I’m fairly confident that the dropdown in this state should look something like this, but I haven’t yet explored the “open state”

![Image](https://user-images.githubusercontent.com/206364/202800205-899a6665-e5e1-478f-802f-de7cd6f96990.png)

- [x] https://github.com/covid-projections/act-now-packages/pull/472 Add labels to the `MetricSeriesChart`
- [x] https://github.com/covid-projections/act-now-packages/pull/477 `Select` and `MultiSelect` components
- [ ] Implement the `MultiRegionMultiMetricChart` component

PR #472, opened 2022-12-02
Add support for labels on `MetricSeriesChart`

pnavarrc:
Part of #413 - Add support for labels on `MetricSeriesChart`. This is necessary to implement the Trends Chart, but mostly independent of the rest of the component.

Add support for optional labels for the `MetricSeriesChart`. I modified the algorithm we used at CAN to prevent the labels from overlapping. Basically, we go label by label in ascending order of y-coordinate, and adjust the position of each label if it would overlap with one of the previous labels.

In this screenshot, the label for _Indiana_ is moved down a little to prevent it from overlapping with the label of _New York_. 

image

## Notes

- Adding labels on the `Series` object is a bit more flexible than using the region names (we could show region names and last value, for example)
- It's responsibility of the user to adjust the `marginRight`, since that's already a prop of the component and we can't make a good guess of how much space we will need.


packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
```diff
@@ -18,14 +19,19 @@ import { ChartOverlayXY, useHoveredPoint } from "../ChartOverlayXY";
 import { useMetricCatalog } from "../MetricCatalogContext";
 import { MetricTooltip } from "../MetricTooltip";
 import { PointMarker } from "../PointMarker";
-import { Series, SeriesType } from "../SeriesChart";
-import { SeriesChart } from "../SeriesChart";
+import { Series, SeriesChart, SeriesType } from "../SeriesChart";
```
pnavarrc, line 22: Merged the two imports, it would be nice to have a plugin that does that for us (I'm going to take a look)

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.stories.tsx
```diff
@@ -28,20 +35,16 @@ Vaccination.args = {
   height,
   series: [
     {
-      region: states.findByRegionIdStrict("56"),
-      metric: metricCatalog.getMetric(MetricId.MOCK_CASES),
+      region: WY,
+      metric: metricMockCases,
```
mikelehen: Not a big deal, but why not just pass the metric ID (`MetricId.MOCK_CASES`)? 🤷

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.stories.tsx
```diff
@@ -28,20 +35,16 @@ Vaccination.args = {
...
       type: SeriesType.LINE,
-      lineProps: {
-        stroke: theme.palette.gradient[100],
-      },
+      lineProps: { stroke: theme.palette.gradient[100] },
```
mikelehen, line 39: Not a big deal, but I'm confused why this got reformatted...  Did you do that?  Is prettier doing weird things?

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
mikelehen: Ah, yeah. I'm surprised that doesn't happen automatically. :/

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.stories.tsx
```diff
@@ -87,25 +88,52 @@ TrendsMultipleLocations.args = {
...
+      lineProps: { stroke: schemeCategory10[2], strokeDasharray: "4 4" },
+    },
+  ],
+};
+
+export const MultipleLocationsWithLabels = Template.bind({});
+MultipleLocationsWithLabels.args = {
+  width,
+  height,
+  marginRight: 100,
```
mikelehen, line 113: Part of me disagrees with your "It's responsibility of the user to adjust the marginRight, since that's already a prop of the component and we can't make a good guess of how much space we will need." because it's not obvious how they are in a better position to know how much space we will need to render the labels.  We could presumably calculate it automatically using something akin to https://stackoverflow.com/a/58705306

Though maybe in some cases they will want the labels truncated?  I'm a bit on the fence, but making them specify the margin is definitely easier, so we can do that for now.

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
```diff
@@ -18,14 +19,19 @@ import { ChartOverlayXY, useHoveredPoint } from "../ChartOverlayXY";
...
+import { SeriesLabel } from "./MetrisSeriesChart.style";
 
 export interface MetricSeriesChartProps extends BaseChartProps {
   /** List of series to be rendered */
   series: Series[];
   /** Minimum value for the y-axis. It defaults to 0. */
   minValue?: number;
+  /**
+   * Show labels for each series. The labels will be shown on the right side
+   * of the chart, closer to the last value of each series.
```
mikelehen: closer => close ?

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
```diff
@@ -109,6 +118,23 @@ export const MetricSeriesChart = ({
...
+      assert(timeseries.hasData(), `The timeseries should have data`);
+
+      return {
+        y: yScale(timeseries.lastValue),
+        label: series.label,
+        fill: getSeriesColor(series, defaultTextColor),
+      };
+    });
+
+  const labelPositions = calculateLabelPositions(initialLabelPositions, 24);
```
mikelehen: nit: I can't guess what `24` refers to so consider adding a `/*labelHeight=*/24` comment.  I also wonder if this should be based on `theme.typography.paragraphSmall.fontSize` or something (in which case I could probably guess what it's for and you wouldn't need the comment 😁 ).  But not sure if that font size is guaranteed to be a pixel value or if we'd need to somehow convert it to pixels...  Feel free to add a TODO and punt for now.

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
```diff
@@ -131,6 +157,21 @@ export const MetricSeriesChart = ({
...
           />
         ))}
+        {showLabels && (
+          <Group>
+            {labelPositions.map((item, itemIndex) => (
+              <SeriesLabel
+                key={`label-${item.label}-${itemIndex}`}
+                x={chartWidth}
+                y={item.y}
+                dx={5}
```
mikelehen: Out of curiosity, how is this different / better than doing `x={chartWidth+5}` or something?

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
```diff
@@ -195,10 +239,42 @@ function getValueRange(timeseriesList: Timeseries<number>[]): [number, number] {
...
+
+  return sortedItems.reduce(
+    (adjusted: LabelInfo[], item: LabelInfo, index: number) => {
+      const dy = labelHeight / 2;
+
+      // Take the maximum y-coordinate from the items that are already
+      // adjusted, and compare it with the y-coordinate of the current
+      // item. If the items will overlap, we adjust the position of the
+      // current item.
+      const maxY = max(adjusted.map((p) => p.y)) ?? item.y;
```
mikelehen: The `?? item.y` looks "wrong" to me (we would in theory be trying to adjust item to be above itself), but I guess max() will only be undefined when index==0 and then we don't use maxY below so it doesn't matter.

Could we do `?? Number.NEGATIVE_INFINITY` and then get rid of the "index > 0" check below?  I think that'd be simpler / clearer.

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
```diff
@@ -195,10 +239,42 @@ function getValueRange(timeseriesList: Timeseries<number>[]): [number, number] {
...
+function calculateLabelPositions(
+  items: LabelInfo[],
+  labelHeight: number
+): LabelInfo[] {
+  // Sort the items by SVG y-coordinate, in ascending order.
+  const sortedItems = sortBy(items, (item) => item.y);
+
+  return sortedItems.reduce(
+    (adjusted: LabelInfo[], item: LabelInfo, index: number) => {
+      const dy = labelHeight / 2;
```
mikelehen: It's not obvious to me why this is sufficient to keep the labels from overlapping.  I'd expect them to overlap 50%.  But maybe the passed in labelHeight is actually 2x the font size or something?  Seems like there might be some fragile assumptions here.  Or I'm just confused.

mikelehen (approved):
Nice!  Looks good with some minor nits.

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.stories.tsx
pnavarrc: I did, just wanted to have a more compact list 😅

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.stories.tsx
pnavarrc: That's fair, I considered a few options:

- Adding a new prop so the user can determine how much space to use for the margins. They could test using the Trends chart and decide how much space for the labels they want (they can adjust the actual labels too, for example, using short region names or state codes). There is also `@visx/text` that allows for text wrapping (but that means updating the algorithm to position the labels vertically)
- Automatically adding space for the labels. A bit tricky since we would have to update the space after rendering the labels (they can only be measures after rendering), and we will have to handle labels wider than the chart, etc.

I think a new prop with a reasonable default and no wrapping is probably okay for now, I'm going to update the PR to have that.

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
pnavarrc: Yeah, I think that grabbing the values as numbers might get complicated (plus even by using the pixel value of `lineHeight` the labels didn't look quite right, so I had to tweak the number until it looked better).

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
pnavarrc: ~~Is not different, maybe that's better~~ There are subtle differences (more noticeable when using `tspan` inside `text` elements), but those are not relevant in this case, updating!

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
pnavarrc: It wouldn't quite work, since for the first element `maxY + dy > item.y` will be `true` and then we will add the `dy` incorrectly in that case.

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.stories.tsx
mikelehen: Ah, cool.  I assumed prettier would force this to be one way or the other.  I guess it's less opinionated than I thought!

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.stories.tsx
mikelehen: Cool, thanks for the explanation.  I'm fine with keeping it simple and just using marginRight for now if you don't feel like adding another prop. 🤷

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.stories.tsx
pnavarrc: yeah, I think Prettier is cool with either option 🤷

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.stories.tsx
pnavarrc: Good call!

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
pnavarrc: Good point, I reviewed the implementation, we actually just need to pass the calculated `line-height` of `paragraphSmall` (in pixels) and we don't need the division by 2. I couldn't find a way to get the pixel value from the lineHeight prop in the typography though, so I left a comment for now.

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.stories.tsx
pnavarrc: okay, going to leave the marginRight for now 😁

Issue #420, opened 2022-11-21
Compare table error state


mikelehen:
Currently the compare table just renders a blank square if there's an error.  I think it should say "Data is not available." or similar.

mikelehen:
Completed.

PR #474, opened 2022-12-05
Fix #418,#419,#420: Better error states for charts, maps, and compare table.

### LINKED ISSUE #418: Maps - Better error states
Currently our components just don't render or they stay in their "loading state" when there's an error.  I think we should probably render a blank square with "Data is not available" or similar...

mikelehen:
Introduces an `ErrorBox` component and uses it to render a grey error square when a metric fails to load:

image

See:
* https://act-now-packages--pr474-mikelehen-error-stat-lm5yzgbw.web.app/storybook/index.html?path=/story/components-metriclinechart--loading-error
* https://act-now-packages--pr474-mikelehen-error-stat-lm5yzgbw.web.app/storybook/index.html?path=/story/components-metriclinethresholdchart--loading-error
* https://act-now-packages--pr474-mikelehen-error-stat-lm5yzgbw.web.app/storybook/index.html?path=/story/components-metricserieschart--loading-error
* https://act-now-packages--pr474-mikelehen-error-stat-lm5yzgbw.web.app/storybook/index.html?path=/story/components-metricsparklines--loading-error
* https://act-now-packages--pr474-mikelehen-error-stat-lm5yzgbw.web.app/storybook/index.html?path=/story/components-metricusnationalmap--loading-error
* https://act-now-packages--pr474-mikelehen-error-stat-lm5yzgbw.web.app/storybook/index.html?path=/story/components-metricusstatemap--loading-error
* https://act-now-packages--pr474-mikelehen-error-stat-lm5yzgbw.web.app/storybook/index.html?path=/story/components-metricworldmap--loading-error
* https://act-now-packages--pr474-mikelehen-error-stat-lm5yzgbw.web.app/storybook/index.html?path=/story/components-multimetricusstatemap--loading-error

Notes:
* I'm rendering pretty generic, unhelpful error text with the expectation that we don't want to show error details to users most of the time.  But the error will always be logged in the console.
* For maps, I am using a "default" width/height that won't match the actual map size.  Alternatively I could export the aspect ratio so that we could calculate the error box appropriately, but since errors should not occur during "normal" experience, I don't think it's worth adding complexity.
* I just show a blank box (no error text) for sparklines since they are usually rendered small and so there isn't much room for text. Plus it might be noisy since sparklines are usually rendered right next to each other.

### DIFF
--- a/packages/ui-components/src/index.ts
+++ b/packages/ui-components/src/index.ts
@@ -87,6 +87,7 @@ export * from "./components/BarChart";
 export * from "./components/ChartOverlayX";
 export * from "./components/ChartOverlayXY";
 export * from "./components/CompareTable";
+export * from "./components/ErrorBox";
 export * from "./components/Grid";
 export * from "./components/InfoTooltip";
 export * from "./components/LabelIcon";
@@ -124,5 +125,3 @@ export * from "./components/SparkLine";
 export * from "./components/USNationalMap";
 export * from "./components/USStateMap";
 export * from "./components/WorldMap";
-
-export * from "./components/ErrorBox";
--- a/packages/ui-components/src/index.ts
+++ b/packages/ui-components/src/index.ts
@@ -124,3 +124,5 @@ export * from "./components/SparkLine";
 export * from "./components/USNationalMap";
 export * from "./components/USStateMap";
 export * from "./components/WorldMap";
+
+export * from "./components/ErrorBox";
--- a/.changeset/olive-bananas-travel.md
+++ b/.changeset/olive-bananas-travel.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/ui-components": patch
+---
+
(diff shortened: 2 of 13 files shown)


packages/ui-components/src/index.ts
```diff
@@ -124,3 +124,5 @@ export * from "./components/SparkLine";
 export * from "./components/USNationalMap";
 export * from "./components/USStateMap";
 export * from "./components/WorldMap";
+
+export * from "./components/ErrorBox";
```
pnavarrc: Consider sorting this export (I select the exports in VSCode and use `cmd + shift + p` → _Sort Lines Ascending_

pnavarrc (approved):
Looks good!

packages/ui-components/src/index.ts
mikelehen: Oh, nice trick! I just used plop to add the component.  Would be neat if we could somehow automate the sorting, but I don't see a trivial way.  Oh well!

Issue #336, opened 2022-10-25
Implement metric MetricTabs component


pnavarrc:
Implement Tabs that show the metric overview and a metric chart in the tab panel. The component should receive a region and a list of metrics. The tabs and content should be responsive.

![Image](https://user-images.githubusercontent.com/114084/197887190-d0684b6b-6ec3-4f3b-8a44-47f83f330fab.png)

smcclure17:
#479

PR #479, opened 2022-12-06
Implement MetricChartBlock component

smcclure17:
addresses #336 | [stories](https://act-now-packages--pr479-smcclure17-metric-me-9otxsxp6.web.app/storybook/index.html?path=/story/components-metrictabs--example) | [mocks](https://www.figma.com/file/YN7x6bJmSfRHaovP5ABKu2/ActNow-Design-System?node-id=720%3A22518&t=5nRqpMZnfpZltOs6-0)

Implements only clickable tabs. 

I'm unsure how exactly we want to handle tab-content alignment for metrics that use multiple lines. Here is the current formatting:
image


packages/ui-components/src/components/MetricTabs/MetricTabs.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+import { useMetricCatalog } from "../../components/MetricCatalogContext";
+import { MetricLineChart } from "../MetricLineChart";
+import { MetricLineThresholdChart } from "../MetricLineThresholdChart";
+import { MetricValue } from "../MetricValue";
+import { MetricTab } from "./MetricTabs.style";
+
+export interface MetricTabsProps {
+  region: Region;
+  metrics: Array<Metric> | Array<string>;
```
chasulin: i assume the array can contain both `Metric` and `string` at the same time (rather than `metrics: Array<Metric> | Array<string>` declaring that it will be an array of entirely one or the other i think), so instead: `metrics: (Metric | string)[]` or `metrics: Array<string | Metric>`

packages/ui-components/src/components/MetricTabs/MetricTabs.tsx
```diff
@@ -0,0 +1,90 @@
...
+const TabContent = ({ metric, region }: { metric: Metric; region: Region }) => {
+  return (
+    <Stack spacing={1}>
+      <Typography variant="labelLarge" textAlign="left">
+        {metric.name}
+      </Typography>
+      <MetricValue
+        metric={metric}
+        region={region}
+        variant="dataEmphasizedLarge"
```
chasulin: `dataEmphasizedLarge` is the default variant for `MetricValue` ([see here](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/components/MetricValue/MetricValue.tsx#L27)), so you dont need to pass it here

chasulin (commented):
Looks good! A few thoughts/suggestions-

- I think we can actually just call this `MetricChartBlock`, since it includes both the tabs and the chart.
- [This bit of css (81-84)](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/styles/theme/components.tsx#L81-L84) that is already at the theme level is causing some weird styling. i think we can delete it.
- Re: the alignment, I think having the tab content aligned to the bottom of the container works, so adding justifyContent:'flex-end' to the tab styling (which you can do [at the theme level in components.tsx](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/styles/theme/components.tsx#L69-L71))

I poked around and found that @mui/lab has some cool helper components for composing content blocks controlled by tabs. See [the example here](https://mui.com/material-ui/react-tabs/#experimental-api) with `TabContext`, `TabList`, and `TabPanel`. It's pretty sweet. To give it a go, you first need to install @mui/lab (and for some reason when I went to test it out, it made me run `yarn add @mui/material @emotion/react @emotion/styled`, maybe to upgrade them 🤷🏻‍♀️)

smcclure17:
hi @chasulin --thanks for helping look into this issue 🙏 

I've been able to get the build to succeed by explicitly defining the types for styled `MuiLinks` (see eb26aed8b7b9447f3e9773cafdb27c573fef3776). This does exclude some extra types/attributes that TS was initially inferring (see screenshot) so I'm not sure if this is a great solution? But, it seems to be working...
image

packages/ui-components/src/components/MetricCompareTable/MetricCompareTable.style.ts
```diff
@@ -10,6 +10,8 @@ export const StyledTableCell = styled(TableCell)`
   padding: 0;
 `;
 
+// TODO: Issue with upgraded MUI/emotion requires an explicit type annotation
+// here. See https://github.com/covid-projections/act-now-packages/issues/486
```
chasulin: i dont think we need the whole URL here. it can just be-

> // TODO: Issue with upgraded MUI/emotion requires an explicit type annotation. See #486

packages/ui-components/src/components/MetricChartBlock/MetricChartBlock.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { MetricLineThresholdChart } from "../MetricLineThresholdChart";
+import { MetricValue } from "../MetricValue";
+import { MetricTab } from "./MetricChartBlock.style";
+
+export interface MetricChartBlockProps {
+  region: Region;
+  metrics: Array<Metric | string>;
+  width?: number;
+  height?: number;
+}
```
chasulin: lets add documentation to these props 🙏🏼. see other metric aware components for examples ([heres one](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/components/MetricSparklines/MetricSparklines.tsx))

packages/ui-components/src/components/MetricChartBlock/MetricChartBlock.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+  region: nations.findByRegionIdStrict("USA"),
+  metrics: metricList,
+};
+
+export const SmallExample = Template.bind({});
+SmallExample.args = {
+  region: nations.findByRegionIdStrict("USA"),
+  metrics: [...metricList, MetricId.MOCK_CASES_DELAY_1S],
+  width: 320,
+};
```
chasulin: instead of a hardcoded small screen size, it could be more useful to have a single example that's responsive (you can implement `AutoWidth` with a wrapping `Box` in the story template above)

chasulin (approved):
Left a few small comments, but otherwise looks good!

packages/ui-components/src/components/MetricChartBlock/MetricChartBlock.tsx
smcclure17: ah, right! sorry!

PR #480, opened 2022-12-06
Implement footer

faihegberg:
Implement a simple footer container. Closes https://github.com/covid-projections/act-now-template/issues/59

### DIFF
--- a/packages/ui-components/src/components/Footer/Footer.stories.tsx
+++ b/packages/ui-components/src/components/Footer/Footer.stories.tsx
@@ -1,7 +1,14 @@
 import React from "react";
 import { ComponentStory, ComponentMeta } from "@storybook/react";
 import { Footer } from ".";
-import { Box, Typography, Stack, Link, IconButton } from "@mui/material";
+import {
+  Box,
+  Typography,
+  Stack,
+  Link,
+  IconButton,
+  useTheme,
+} from "@mui/material";
 import FacebookIcon from "@mui/icons-material/Facebook";
 import TwitterIcon from "@mui/icons-material/Twitter";
 import InstagramIcon from "@mui/icons-material/Instagram";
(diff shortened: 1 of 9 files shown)


chasulin (commented):
i think this was supposed to be implemented in the template repo. i left some comments though! (for when you move the code over to the template)

packages/ui-components/src/components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+    <Footer>
+      <Box maxWidth={100}>
+        <a href="/">
+          <img
+            src="https://upload.wikimedia.org/wikipedia/commons/thumb/f/f7/Bananas.svg/2560px-Bananas.svg.png"
+            width="100%"
+          />
+        </a>
+      </Box>
+      <Typography variant="footer">
```
chasulin, line 33: this is `paragraphLarge` in the mocks, so we dont need a custom typography variant

Screen Shot 2022-12-07 at 10 44 20 AM

packages/ui-components/src/components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+            Contact
+          </Link>
+          &nbsp;·&nbsp;
+          <Link href="/" color="inherit">
+            Terms
+          </Link>
+        </Typography>
+      </Stack>
+      <Stack direction="row">
+        <IconButton edge="start" sx={{ color: theme.palette.footer.content }}>
```
chasulin, line 53: these icons are white, so we dont need a custom palette variant. they can pull from the existing theme

packages/ui-components/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { Stack, Box, useTheme } from "@mui/material";
+
+export interface FooterProps {
+  children: React.ReactNode;
+}
+
+export const Footer = ({ children }: FooterProps) => {
+  const theme = useTheme();
+  return (
+    <Box bgcolor={theme.palette.footer.background} paddingX={3} paddingY={8}>
```
chasulin, line 11: the outermost container should be a [footer element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/footer), so that it shows up as such in the dom. you can do `styled("footer")...` and add any needed styling there

packages/ui-components/src/index.ts
```diff
@@ -73,6 +76,10 @@ declare module "@mui/material/styles" {
       main: string;
       axis: string;
     };
+    footer: {
+      background: string;
+      content: string;
+    };
```
chasulin, line 82: see above comments ([one](https://github.com/covid-projections/act-now-packages/pull/480/files#r1042373410), [two](https://github.com/covid-projections/act-now-packages/pull/480/files#r1042374757)). these custom theme variants probably arent needed

packages/ui-components/src/components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+            width="100%"
+          />
+        </a>
+      </Box>
+      <Typography variant="footer">
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam sit amet
+        imperdiet lectus.
+      </Typography>
+      <Stack direction="row">
+        <Typography variant="footer">
```
chasulin, line 38: we shouldnt necessarily use `Typography` as a wrapper element to apply styling. wrapping these links in `Typography` wraps them in a `span` in the dom. 

an alternate option would be to replace the wrapping `Stack` with `Box sx={{color: theme.palette.common.white}}` (or any color), and that color property will be applied to the dot separators. For the color of the links, we can make them into a reusable emotion component with styling, so we dont need to repeat the styles inline

packages/ui-components/src/components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+          <Link href="/" color="inherit">
+            Contact
+          </Link>
+          &nbsp;·&nbsp;
+          <Link href="/" color="inherit">
+            Terms
+          </Link>
+        </Typography>
+      </Stack>
+      <Stack direction="row">
```
chasulin, line 52: nit: this can be a `ButtonGroup`. theres not such a big difference between having it as a `Stack` vs `Box` vs `ButtonGroup` in this situation (because we're not using any of the styling/props that come with ButtonGroup), but i think `ButtonGroup` comes with some accessibility attributes

faihegberg:
Closing since this has been implemented in the template repo.

PR #484, opened 2022-12-08
Adjust minValue and maxValue logic in charts

### LINKED ISSUE #386: Improve handing of minValue, maxValue in charts
We have a few places where we want to set minValue, maxValue in the charts

- The metric might have a defined minValue and maxValue
- The timeseries will have minValue, maxValue
- In some cases, the type of data could have a specific natural range (vaccination, 0% - 100%)
- A specific visual representation might require start from 0 (Bar Chart)
- The minValue and maxValue could be contextual (we want to start at 0 in some cases, but at the min value in other cases)

We need to investigate a way to make the behaviors more consistent across charts. For example, we could have an order of precedence on how the minValue, maxValue are taken from (component props, timeseries, metric definition)

faihegberg:
This PR adjusts `minValue` and `maxValue` calculation logic for our charts to be more consistent with each other. Based on discussion with the team ([doc](https://www.dropbox.com/scl/fi/zmykem7z5r1plryn79r6a/minValue-and-maxValue-in-Charts.paper?dl=0&rlkey=wzsfi96zd9y5ffctkhjuc1zcj) for context), in general `minValue` and `maxValue` are now calculated as follow:

_If the metric definition has a specified `minValue` or `maxValue`, use that value.
If not, and the chart is not a threshold chart, then use the data’s minimum and maximum value. (Threshold chart will use the lower and upper bounds first before resorting to the data’s minimum and maximum value)._

Closes #386

### DIFF
--- a/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
+++ b/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
@@ -62,7 +62,7 @@ export const MetricLineChart = ({
   });
 
   const yScale = scaleLinear({
-    domain: [metric.minValue ?? minValue, maxValue],
+    domain: [metric.minValue ?? minValue, metric.maxValue ?? maxValue],
     range: [chartHeight, 0],
   });
 
--- a/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
+++ b/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
@@ -62,7 +62,7 @@ export const MetricLineChart = ({
   const chartWidth = width - marginLeft - marginRight;
 
   const { minDate, maxDate } = timeseries;
-  const { minValue, maxValue } = getChartRange(metric, timeseries);
+  const [minValue, maxValue] = getChartRange(metric, timeseries);
 
   const xScale = scaleUtc({
     domain: [minDate, maxDate],
--- a/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
+++ b/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
@@ -62,6 +62,8 @@ export const MetricLineChart = ({
   });
 
   const yScale = scaleLinear({
+    // If available, use minValue and maxValue from the metric definition.
+    // Otherwise, use min and max value of the data.
     domain: [metric.minValue ?? minValue, metric.maxValue ?? maxValue],
     range: [chartHeight, 0],
   });
--- a/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
+++ b/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
@@ -54,7 +54,7 @@ export const MetricLineChart = ({
   const chartHeight = height - marginTop - marginBottom;
   const chartWidth = width - marginLeft - marginRight;
 
-  const { minDate, maxDate, maxValue } = timeseries;
+  const { minDate, maxDate, minValue, maxValue } = timeseries;
 
   const xScale = scaleUtc({
     domain: [minDate, maxDate],
@@ -62,7 +62,7 @@ export const MetricLineChart = ({
   });
 
   const yScale = scaleLinear({
-    domain: [0, maxValue],
+    domain: [metric.minValue ?? minValue, maxValue],
     range: [chartHeight, 0],
   });
 
--- a/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
+++ b/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
@@ -70,8 +70,7 @@ export const MetricLineChart = ({
   });
 
   const yScale = scaleLinear({
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    domain: [minValue!, maxValue!],
+    domain: [minValue, maxValue],
     range: [chartHeight, 0],
   });
 
--- a/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
+++ b/packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
@@ -16,6 +16,7 @@ import { LineChart } from "../LineChart";
 import { useMetricCatalog } from "../MetricCatalogContext";
 import { MetricTooltip } from "../MetricTooltip";
 import { PointMarker } from "../PointMarker";
+import { getChartRange } from "../../common/utils/charts";
 
 export interface MetricLineChartProps extends BaseChartProps {
   metric: Metric | string;
@@ -54,7 +55,8 @@ export const MetricLineChart = ({
   const chartHeight = height - marginTop - marginBottom;
   const chartWidth = width - marginLeft - marginRight;
 
-  const { minDate, maxDate, minValue, maxValue } = timeseries;
+  const { minDate, maxDate } = timeseries;
+  const { minValue, maxValue } = getChartRange(metric, timeseries);
 
   const xScale = scaleUtc({
     domain: [minDate, maxDate],
@@ -62,9 +64,8 @@ export const MetricLineChart = ({
   });
 
   const yScale = scaleLinear({
-    // If available, use minValue and maxValue from the metric definition.
-    // Otherwise, use min and max value of the data.
-    domain: [metric.minValue ?? minValue, metric.maxValue ?? maxValue],
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    domain: [minValue!, maxValue!],
     range: [chartHeight, 0],
   });
 
--- a/packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
+++ b/packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
@@ -3,8 +3,6 @@ import React from "react";
 import { Skeleton } from "@mui/material";
 import { Group } from "@visx/group";
 import { scaleLinear, scaleUtc } from "@visx/scale";
-import max from "lodash/max";
-import min from "lodash/min";
 
 import { assert } from "@actnowcoalition/assert";
 import { Metric } from "@actnowcoalition/metrics";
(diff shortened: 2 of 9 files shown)


packages/ui-components/src/components/MetricSparklines/MetricSparklines.tsx
```diff
@@ -64,10 +64,30 @@ export const MetricSparklines = ({
...
+        : undefined;
+
     return (
       <SparkLine
         timeseriesLineChart={timeseriesLineChart}
         timeseriesBarChart={timeseriesBarChart}
+        minValue={minValue}
+        maxValue={maxValue}
         {...optionalProps}
       />
```
faihegberg, line 109: Since `Sparkline` isn't metric-aware but the min and max value of the y-axis is calculated inside it, it must receive min and max value from `MetricSparklines` which is metric-aware.

packages/ui-components/src/components/SparkLine/SparkLine.tsx
```diff
@@ -48,7 +56,7 @@ export const SparkLine = ({
   });
 
   const yScaleBar = scaleLinear({
-    domain: [0, timeseriesBarChart.maxValue],
+    domain: [minValue ?? 0, maxValue ?? timeseriesBarChart.maxValue],
```
faihegberg: We use `minValue` prop here instead of `timeseriesBarChart.minValue` as we're not necessarily using the minimum value of the data as the lowest y-axis value. In the case of `MetricSparklines` for example, we calculate the `minValue` first prior to passing it down to `Sparkline` as a `minValue` prop.

packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
```diff
@@ -54,15 +54,17 @@ export const MetricLineChart = ({
...
   const xScale = scaleUtc({
     domain: [minDate, maxDate],
     range: [0, chartWidth],
   });
 
   const yScale = scaleLinear({
-    domain: [0, maxValue],
+    // If available, use minValue and maxValue from the metric definition.
+    // Otherwise, use min and max value of the data.
+    domain: [metric.minValue ?? minValue, metric.maxValue ?? maxValue],
```
pnavarrc: Consider naming things to make the logic easier to follow:

```tsx
const { 
  minDate, 
  maxDate, 
  minValue: minValueTimeseries, 
  maxValue: maxValueTimeseries 
} = timeseries;

const { 
  minValue: minValueMetric, 
  maxValue: maxValueMetric 
} = metric;

const minValue = minValueMetric ?? minValueTimeseries;
const maxValue = maxValueMetric ?? maxValueTimeseries;

// ... use minValue and maxValue from this point on...
```

Consider writing a utility function to do this since we need to apply this logic in a few places 

```tsx
const { minValue, maxValue } = getChartRange(metric, timeseries)
````

File: packages/ui-components/src/components/SparkLine/SparkLine.tsx (hunk 2 above)
Comment (pnavarrc) [reply packages/ui-components/src/components/SparkLine/SparkLine.tsx]: Maybe let's apply a similar logic as the one used for the metric charts? We can always make `propMinValue = 0` by default while preserving a similar logic as for the metric charts case

```tsx
const minValue = propMinValue ?? timeseriesMinValue;
const maxValue = propMaxValue ?? timeseriesMaxValue;
```

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
```diff
@@ -88,8 +88,17 @@ export const MetricLineThresholdChart = ({
...
+  // Otherwise, use the lower bound of the lowest threshold.
+  // If neither are available, use min value of the data.
+  const minChartValue =
+    metric.minValue ?? min(intervals.map(({ lower }) => lower)) ?? minValue;
+
+  // If available, use maxValue from the metric definition.
+  // Otherwise, use the higher bound of the highest threshold.
+  // If neither are available, use max value of the data.
+  const maxChartValue =
+    metric.maxValue ?? max(intervals.map(({ upper }) => upper)) ?? maxValue;
```
pnavarrc: Again, consider creating some variables with good names to make this more readable. Also note that in this case, the intervals already account for the min and max values of the time series, so we only need to do:

```tsx
const minValueIntervals = min(intervals.map(({ lower }) => lower));
const minValue = minValueMetric ?? minValueIntervals;
// ...
```

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
```diff
@@ -105,12 +102,13 @@ export const MetricSeriesChart = ({
...
 
-  // Note: We use minValue if provided. If not, we default to start from zero,
-  // which is best in most cases, unless minDataValue is negative, in which
-  // case we use that value instead.
-  const minYValue = isNumber(minValue) ? minValue : Math.min(0, minDataValue);
+  // Get the lowest minValue from the metric definitions of included metrics.
+  const metricDefinitionMin = min(metrics.map(({ minValue }) => minValue));
+
+  // Get the highest maxValue from the metric definitions of included metrics.
+  const metricDefinitionMax = max(metrics.map(({ maxValue }) => maxValue));
```
pnavarrc, line 125: Consider writing a function to calculate `minValueMetric` and `maxValueMetric`, and     creating some variables to make the logic easier to follow

pnavarrc (commented):
Looks mostly good, added a few recommendations to make the code easier to understand (and to avoid accounting for the timeseries min and max value twice in the interval chart)

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
faihegberg: Ah right! Thank you.

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
faihegberg: I think this is already a pretty simplified way of getting the lowest minimum value of the metric definitions. I will revise the comment above it though.

packages/ui-components/src/components/SparkLine/SparkLine.tsx
faihegberg: Assigning a default value to `propMinValue` means `propMinValue` will always be used. So I renamed the variables according to your suggestions to increase clarity.

packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
```diff
@@ -54,17 +55,17 @@ export const MetricLineChart = ({
...
     domain: [minDate, maxDate],
     range: [0, chartWidth],
   });
 
   const yScale = scaleLinear({
-    // If available, use minValue and maxValue from the metric definition.
-    // Otherwise, use min and max value of the data.
-    domain: [metric.minValue ?? minValue, metric.maxValue ?? maxValue],
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    domain: [minValue!, maxValue!],
```
faihegberg: I'm not a fan of doing this. If you think there's a straightforward way for `getChartRange` to guarantee a return value that isn't undefined, let me know! Currently `getChartRange` is still potentially returning an `undefined` value to accommodate charts that use `Timeseries` instead of `NonEmptyTimeseries`.

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
```diff
@@ -90,18 +90,19 @@ export const MetricLineThresholdChart = ({
...
-  const maxChartValue =
-    metric.maxValue ?? max(intervals.map(({ upper }) => upper)) ?? maxValue;
+  const maxValueMetric = metric.maxValue;
+  const maxValueIntervals = max(intervals.map(({ upper }) => upper));
+  const maxChartValue = maxValueMetric ?? maxValueIntervals;
 
   const yScale = scaleLinear({
-    domain: [minChartValue, maxChartValue],
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    domain: [minChartValue!, maxChartValue!],
```
faihegberg: Same here regarding guaranteeing a defined value.

packages/ui-components/src/components/MetricLineChart/MetricLineChart.tsx
pnavarrc: I think it would be better to have `getChartRange` to take a non-empty timeseries instead, you need to assert that the timeseries are not empty anyways before getting `minValue` and `maxValue`

pnavarrc: In this case, we know that the timeseries are not empty because we are checking for that here https://github.com/covid-projections/act-now-packages/pull/484/files#diff-3cbc6826ad9491ed4cdacdc4c3344e7f2e757fb6d9238d2ed4364804f0b28e75R57

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
pnavarrc: we are already checking that the timeseries have data, so you should be able to use non-empty timeseries https://github.com/covid-projections/act-now-packages/pull/484/files#diff-f09764f6748515abce5678bcbc8e5d190ec6480c4655a0ab06e32c52816b2179R70

packages/ui-components/src/components/MetricSparklines/MetricSparklines.tsx
```diff
@@ -76,10 +77,27 @@ export const MetricSparklines = ({
...
 
+    const { minValue: minValueLineChart, maxValue: maxValueLineChart } =
+      getChartRange(metricLineChart, timeseriesLineChart);
+
+    const { minValue: minValueBarChart, maxValue: maxValueBarChart } =
+      getChartRange(metricBarChart, timeseriesBarChart);
+
+    const minValue =
+      minValueLineChart && minValueBarChart
+        ? Math.min(minValueLineChart, minValueBarChart)
```
pnavarrc: This is not what we want. Both `minValueLineChart` and `minValueBarChart` are numbers, and either of them could be zero. If you run your code, you will have

```tsx
const minValue = 
  minValueLineChart && minValueBarChart // 0 && -1 → false
    ? Math.min(minValueLineChart, minValueBarChart) // → -1
    : undefined

// → minValue = undefined 
```

we just need to use the minimum between both.

packages/ui-components/src/components/SparkLine/SparkLine.tsx
```diff
@@ -58,7 +66,7 @@ export const SparkLine = ({
   });
 
   const yScaleBar = scaleLinear({
-    domain: [0, timeseriesBarChart.maxValue],
+    domain: [propMinValue ?? 0, propMaxValue ?? timeseriesBarChart.maxValue],
```
pnavarrc: It might be better to create a `minValue` and `maxValue` variables with this logic instead as a convention, in some charts we need to use them, so it would be best to have the logic in a more visible location

```tsx
const minValue = propMinValue ?? 0;

// ...
const yScaleBar = scaleLinear({
  domain: [minValue, maxValue],
  // ...
})

// do something else with minValue without duplicating the logic
```

packages/ui-components/src/components/MetricLineThresholdChart/MetricLineThresholdChart.tsx
faihegberg: Ahh the possible `null` came from the use of `min` and `max` from lodash, not from the intervals 😅

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
```diff
@@ -137,7 +139,10 @@ export const MetricSeriesChart = ({
   });
 
   const yScale = scaleLinear({
-    domain: [minYValue, maxValue],
+    domain: [
+      metricDefinitionMin ?? minDataValue,
+      metricDefinitionMax ?? maxDataValue,
+    ],
```
pnavarrc: It's better to create variables that include this calculation

pnavarrc (approved):
Looks good, minus my comment below - also, please add documentation to the getChartRange following our new conventions

PR #490, opened 2022-12-13
Fix keyboard navigation on RegionSearch

### LINKED ISSUE #363: RegionSearch - Keyboard navigation doesn't work

pnavarrc:
Close https://github.com/covid-projections/act-now-packages/issues/363 - Fix keyboard navigation on `RegionSearch`

The function passed to `renderOption` needs to return a list item, and inject a couple of accessibility props and event handlers that make all work. We were using links which were working when the user clicks on the items, but would interfere with the focus status of each list item, so I updated `onChange` and remove the styled link.

image

I updated the styles a bit to match the current style and removed the hover styling on the `SearchItem` since the styles injected on the list item already have some padding and hove styling.

### DIFF
--- a/packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
+++ b/packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
@@ -3,6 +3,7 @@ import React, { HTMLAttributes } from "react";
 import SearchIcon from "@mui/icons-material/Search";
 import {
   Autocomplete,
+  AutocompleteChangeReason,
   AutocompleteProps,
   TextField,
   createFilterOptions,
@@ -59,7 +60,11 @@ export const RegionSearch = ({
   return (
     <Autocomplete
       options={options}
-      onChange={(e, region: Region | null, reason: string) => {
+      onChange={(
+        e,
+        region: Region | null,
+        reason: AutocompleteChangeReason
+      ) => {
         if (region && reason === "selectOption") {
           // Typescript doesn't allow to assign a string to window.location
           // https://github.com/microsoft/TypeScript/issues/48949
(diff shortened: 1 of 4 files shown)


packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
```diff
@@ -65,17 +59,22 @@ export const RegionSearch = ({
   return (
     <Autocomplete
       options={options}
-      onChange={(e, item: Region | null) => onChange(item)}
+      onChange={(e, region: Region | null, reason: string) => {
+        if (region && reason === "selectOption") {
+          // Typescript doesn't allow to assign a string to window.location
+          // https://github.com/microsoft/TypeScript/issues/48949
+          (window as any).location = regionDB.getRegionUrl(region);
```
mikelehen, line 71: Can we do `window as Window` instead? https://github.com/microsoft/TypeScript/issues/48949#issuecomment-1203967132

In general I'm guessing there's a next.js way to navigate the browser which we would ideally hook into somehow, but this is probably fine for now.

packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
```diff
@@ -65,17 +59,22 @@ export const RegionSearch = ({
   return (
     <Autocomplete
       options={options}
-      onChange={(e, item: Region | null) => onChange(item)}
+      onChange={(e, region: Region | null, reason: string) => {
```
mikelehen: Can we type `reason` as `AutocompleteChangeReason` so

mikelehen (approved):
Don't forget to add a changeset!

packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
pnavarrc: I tried `window as Window`, but still didn't work 🤷 (I'm not sure why, maybe our version of TS?). 

We could use `useRouter` from Next, but we will need to figure out a way to pass any type of router to this component if we want to make it generic enough (or we could pass the `onSelect` as a prop so the user can call the navigation) - going to punt on that for now.

packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
pnavarrc: good call!

PR #491, opened 2022-12-13
Documentation cleanup

chasulin:
This PR makes the documentation in `ui-components` ✨ consistent and comprehensive ✨. The audit encompasses all files within the `/components` folder, as well as `BaseChartProps` in `common/utils/charts.ts` and `BaseUSMapProps` in `common/utils/maps.ts`

I generally followed the conventions and formatting used by MUI to document their props, as well as some other misc. conventions I went by, including:
- Document **all** component props
- Note a default value when there is one, using `@default defaultValue`. No added empty line between the prop's documentation and the `default` value.
- Note the params of functional props when there are any, using `@param {paramType} paramName`. Add one empty line between the prop's documentation and the `param` notation.
- Generally don't need to add documentation to props in storybook files, as they're usually just adding types to examples of generics.
- In *most* cases, don't put example code (ie. an coded out example of how to implement the component) in the component file's documentation. Examples should live in the storybook file.
- We have a lot of props that are used throughout multiple interfaces (ie. `xScale`, `yScale`, `getTooltip`). Let's keep language consistent and use what is decided on in this PR. If you are documenting a new prop that is one of these often reused props, reference/reuse the existing documentation.
- When writing documentation, generally try to use the present tense, and avoid using 'would' or passive language.
- Comments that are documentation are formatted with /** * */, ie:
```
/**
* This is an example of documentation of a prop.
* Width of the map container.
*/
```
```
/**
* This is an example of documentation explaining what a particular component is.
* AutoWidth is a wrapper that takes the width of the
* parent component and passes it down to the children.
* Under the hood, it uses `ParentSize` from @visx/responsive.
*/
```
- Comments that are explanatory in a note-ish way are formatted with `//` , ie:
```
// This is an example of a comment writing a note about code that's on the following line.
// Return the child if it already has a numeric `width` prop.
```
```
// This is an example of a comment writing a note about code that's on the following line. 
// For `minYValue`, we use `minValue` if provided.
// If no `minValue` is provided, we default to 0,
// unless `minDataValue` is negative, in which case we use that value.
```

There are also some small edits to props, which I've added comments throughout the PR to point out.

All in all, this is meant to be used as a good clean starting point and a guide. **Nothing too strict!** 😎


mikelehen:
Thanks @chasulin!  You can go to sleep now. 😛

packages/ui-components/src/components/MetricLegendCategorical/MetricLegendCategorical.tsx
```diff
@@ -23,7 +25,7 @@ const getItemLabel = (item: Category) => item.name ?? item.id;
 
 export const MetricLegendCategorical = ({
   metric,
-  orientation,
+  orientation = "horizontal",
```
chasulin, line 28: added this default value of "horizontal" to match the default value in the non-metric-aware version (`LegendCategorical`)

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
```diff
@@ -4,40 +4,83 @@ import { LegendThresholdHorizontal } from "./LegendThresholdHorizontal";
...
 /**
- * `LegendThreshold` represents a scale with thresholds that separate
- * a set of categories. By default, the labels between each category are shown.
+ * LegendThreshold represents a scale with thresholds that separate a set
+ * of categories, each represented by a color and labels.
  */
 
-export const LegendThreshold = <T,>(props: LegendThresholdProps<T>) => {
+export const LegendThreshold = <T,>({
+  showLabels = true,
```
chasulin, line 77: moved this default `true` value up a level. previously we were declaring it twice, in both `LegendThresholdVertical` ([here](https://github.com/covid-projections/act-now-packages/pull/491/files#diff-b659c80e8fef40b6eb507457663650a9915e9512e82918f3a8c869724cb29b12L16)) and `LegendThresholdHorizontal` ([here](https://github.com/covid-projections/act-now-packages/pull/491/files#diff-3921c12dc909ffb0a5014f0c67d810a34c28c8a0f0b82b6115e64a91a503f408L25)). by moving it up, we declare it once and keep things in sync for both variations.

packages/ui-components/src/components/MetricLegendThreshold/MetricLegendThreshold.tsx
```diff
@@ -33,6 +49,7 @@ export const MetricLegendThreshold = ({
   startLabel,
   endLabel,
   showOverview = true,
+  showLabels = true,
```
chasulin, line 52: set this default `showLabels` value to `true`, to match the default value in the non-metric-aware version `LegendThreshold`.

packages/ui-components/src/components/BarChart/BarChart.tsx
```diff
@@ -7,51 +7,37 @@ import { ScaleLinear, ScaleTime } from "d3-scale";
...
- *   />
- * </svg>
- * ```
- *
- * @param timeseries Timeseries to represent as a bar chart.
- * @param xScale d3-scale to transform dates into pixel positions on the x-axis
- * @param yScale d3-scale to transform values into pixel heights of each bar.
- * @param barWidth width (in pixels) of each bar.
- *
- * @returns An SVG Group element
```
chasulin, line 53: removed this example code, as we already have examples [in the storybook file](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/components/BarChart/BarChart.stories.tsx). see note in PR description about trying to keep all example code in storybook files rather than in component files.

pnavarrc:
We discussed this on a call, but going to leave this comment here in case others are also looking at this PR:

I would recommend following the syntax and conventions of [tsdoc](https://tsdoc.org/), which are slightly different. For example, the type of parameters is not included in the annotations since it's already available in the TS code.

There a handy plugin https://tsdoc.org/pages/packages/eslint-plugin-tsdoc/ that could help us check that the docstrings follow the conventions (I ran the plugin on this [branch](https://github.com/covid-projections/act-now-packages/compare/pablo/eslint-plugin-tsdoc?expand=1), it throws ~ 400 warnings, most of them related to the type annotations).

I think the best course of action is to ignore the type annotations in this PR and focus on the descriptions, and then write a follow up PR (happy to help with that) to update the types and install the plugin.

packages/ui-components/src/common/utils/maps.ts
```diff
@@ -1,9 +1,28 @@
...
+  /**
+   * Function that returns the `regionUrl` for the region corresponding to a given regionId.
+   * @default undefined
+   *
+   * @param {string} regionId RegionId of the region for which to get the regionUrl.
+   */
   getRegionUrl?: (regionId: string) => string | undefined;
+  /**
+   * Width of the SVG containing the map.
+   */
```
pnavarrc, line 25: Note (for later): It doesn't seem that this interface is US-specific, maybe we could convert it to `BaseMapProps` with some minor tweaks (adding optional height and/or width to height ratio)

packages/ui-components/src/components/AutoWidth/AutoWidth.tsx
```diff
@@ -7,23 +7,27 @@ import isNumber from "lodash/isNumber";
...
+ * Under the hood, it uses `ParentSize` from @visx/responsive.
  *
- * If the child element already has a `width` property, the passed property
- * will be preserved. The child component will only be rendered once the
- * parent has been fully measured (width > 0) to prevent from rendering
- * the component too early.
+ * If the child element already has a `width` property, it will
+ * be overwritten by the passed-down width. The child element
+ * will only be rendered once the parent has a width greater than 0,
+ * in order to prevent from rendering the component too early.
```
pnavarrc: Maybe we should rephrase this (the original string is also confusing). The behavior is:

- If the children component instance is being passed a width prop, that one is used
- If the children component instance is not being passed a width prop, then AutoWidth injects the measured width

```tsx
// we shouldn't be using AutoWidth in this case anyways
// the component will use the prop value (300)
<div style={{ width: 500 }}>
  <AutoWidth>
    <Chart width={300} /> 
  </AutoWidth>
</div>

// the component will use the width of the parent (500)
<div style={{ width: 500 }}>
  <AutoWidth>
    <Chart /> 
  </AutoWidth>
</div>
```

packages/ui-components/src/components/LegendCategorical/LegendCategorical.tsx
```diff
@@ -6,22 +6,36 @@ import Typography from "@mui/material/Typography";
...
+   * @param {number} itemIndex Index of the legend item.
+   */
   getItemColor: (item: T, itemIndex: number) => string;
-  /** Function that returns the label of each legend item */
+  /**
+   * Function that returns the legend item's label.
+   *
+   * @param {T} item The legend item.
+   * @param {number} itemIndex Index of the legend item.
+   */
```
pnavarrc, line 25: Note for later: We can use https://tsdoc.org/pages/tags/typeparam/ for generics to annotate `T`

packages/ui-components/src/components/LegendThreshold/LegendThreshold.tsx
```diff
@@ -4,40 +4,83 @@ import { LegendThresholdHorizontal } from "./LegendThresholdHorizontal";
...
+   * Width of the thermometer.
+   * @default 12 (vertical orientation)
+   * @default 300 (horizontal orientation)
+   */
   width?: number;
-  /** Border radius of the thermometer bar */
+  /**
+   * Border radius of the thermometer.
+   * @default 6 (vertical orientation)
+   * @default 10 (horizontal orientation)
```
pnavarrc, line 26: Note for later: it feels weird to have 2 defaults, maybe there is a way to document the props and their corresponding values in the specific types for each type of legend? (we can figure this out later)

packages/ui-components/src/components/LineChart/LineChart.tsx
```diff
@@ -9,13 +9,17 @@ import { ScaleLinear, ScaleTime } from "d3-scale";
...
-  /** d3-scale to transform point dates to pixel positions on the x-axis */
+  /**
+   * d3-scale to convert between date points and pixel positions.
+   */
   xScale: ScaleTime<number, number>;
-
-  /** d3-scale to transform point values to pixel positions on the y-axis */
+  /**
+   * d3-scale to convert between numerical points and pixel positions.
+   */
```
pnavarrc, line 22: Note for later: Maybe there is a way to reuse the comments so we don't have to copy them over every time?

packages/ui-components/src/components/SparkLine/SparkLine.tsx
```diff
@@ -35,7 +46,6 @@ export const SparkLine = ({
   height = 50,
   barWidth = 2,
 }: SparkLineProps) => {
-  const theme = useTheme();
```
pnavarrc, line 38: 👀  - how did the linter miss that before? 🤔

packages/ui-components/src/components/TimeseriesLineChart/TimeseriesLineChart.tsx
```diff
@@ -10,6 +10,9 @@ import { AxesTimeseries } from "../AxesTimeseries";
 import { LineChart } from "../LineChart";
 
 export interface TimeseriesLineChartProps extends BaseChartProps {
+  /**
+   * Non-empty timeseries used to draw the line chart.
+   */
```
pnavarrc, line 15: Note for later: Maybe we should make all the charts that take time series take non-empty time series?

pnavarrc (approved):
This looks great, thanks @chasulin ! I added a few notes that we can address later at some point, but this is a great improvement!

Let's sync tomorrow with the team on using tsdoc syntax and linting rules and update that on a follow up PR, depending on how the discussion goes.

packages/ui-components/src/components/AutoWidth/AutoWidth.tsx
chasulin: cool, i updated the phrasing a bit more to make it clearer. also, [do you want to revisit this todo you put in the file?](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/components/AutoWidth/AutoWidth.tsx#L6-L8) I'm not sure what the intended meaning there is. Should it be turned into a gh issue? Deleted? etc

packages/ui-components/src/components/LineChart/LineChart.tsx
chasulin: hm maybe. its easy enough to copy them over, so i wouldnt dwell on this one too much

packages/ui-components/src/components/TimeseriesLineChart/TimeseriesLineChart.tsx
chasulin: want to add this as a gh issue? im not sure it falls into the category of documentation audit stuff

packages/ui-components/src/components/AutoWidth/AutoWidth.tsx
pnavarrc: Yeah, that should be a GH issue - the problem is, the AutoWidth component should only children that can receive an optional `width`, not a required `width`, I'm going to create an issue.

packages/ui-components/src/components/TimeseriesLineChart/TimeseriesLineChart.tsx
pnavarrc: yeah I will create an issue, nothing to do with the documentation! 👍

PR #492, opened 2022-12-14
Dialog component

chasulin:
[Storybook examples](https://act-now-packages--pr492-casulin-dialog-qs6v50sr.web.app/storybook/index.html?path=/story/components-dialog--with-title)


packages/ui-components/src/common/hooks/useDialog.ts
```diff
@@ -0,0 +1,18 @@
+import { useState } from "react";
+
+/**
+ * React hook used to generate three props for the Dialog component.
```
mikelehen: It's not very clear what this is for, when / why / how I would use it, etc.  How about:

> React hook used to manage the the opening / closing of the Dialog component.  
> 
> See {@link Dialog} for example usage.

And then add an example there (see my other comment).

packages/ui-components/src/components/Dialog/Dialog.style.ts
```diff
@@ -0,0 +1,3 @@
+import { styled } from "../../styles";
+
+export const Container = styled("div")``;
```
mikelehen: I think plop auto-generated this but it's unused?

packages/ui-components/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,54 @@
...
+   * Content of the dialog title.
+   */
+  title?: React.ReactNode;
+  /**
+   * Main content of the dialog.
+   */
+  children: React.ReactNode;
+}
+
+export const Dialog = ({ open, onClose, title, children }: DialogProps) => {
```
mikelehen, line 37: Should we have a doc comment?  Maybe reference `useDialog()` add an `@example` (and perhaps copy/paste in one of the Storybook examples)?

packages/ui-components/src/common/hooks/useDialog.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @param initialIsOpen An initial open state. If undefined, `open` defaults to an initial value of false.
+ */
+
+export function useDialog(
```
mikelehen: I'd be tempted to rename to `useDialogState()` but don't feel strongly.

mikelehen (approved):
Minor suggestions, but looks good!

packages/ui-components/src/components/Dialog/Dialog.style.ts
chasulin: ah yes whoops, good catch

packages/ui-components/src/common/hooks/useDialog.ts
chasulin: ah yes, that wordage is clearer. thanks!

packages/ui-components/src/components/Dialog/Dialog.tsx
chasulin: good call, i added a doc comment. i added an `@example` in the hook file, but idk about copying/pasting a storybook example into the component file. in [the first doc cleanup pr](https://github.com/covid-projections/act-now-packages/pull/491), i noted the following in the description:
> In most cases, don't put example code (ie. an coded out example of how to implement the component) in the component file's documentation. Examples should live in the storybook file.

let me know if you have objections though, its all flexible

mikelehen: Ah, cool!  Let's stick with that convention for now then!  We can discuss the best way to use docs / storybook to help devs succeed in using our components later.  Maybe there's a way to link directly to storybook examples or something.

chasulin: sounds good

Issue #452, opened 2022-11-28
Dark Theme


mikelehen:
We have a dark theme (borrowed from Pablo's hackathon project), but haven't verified it works well for all components. We should go through Storybook and look for big issues and fix them.

PR #493, opened 2022-12-15
Dark theme improvements

faihegberg:
Create a story for dark theme palette.

Minor tweaks to the following components for dark theme. The goal is just to make the component more visible for now, but specific color choices can probably be determined with Josh's guidance.
- Contained button hover state
- Chip
- Indicator color on tab component
- Indicator color on toggle button
- Color for h4, h5, h6, paragraphLarge, and overline typography

Relates to https://github.com/covid-projections/act-now-packages/issues/452

### DIFF
--- a/packages/ui-components/src/stories/Palette.stories.tsx
+++ b/packages/ui-components/src/stories/Palette.stories.tsx
@@ -51,6 +51,8 @@ const ColorBox = ({ color }: { color: string }) => (
   />
 );
 
+const NON_COLOR_PROPERTIES = ["mode", "getContrastText", "augmentColor"];
+
 // Filter out the palette attributes that don't correspond to colors,
 // such as functions or numeric constants. The `mode` attribute is
 // a string to name themes (dark, light, highContrast), but not a color,
@@ -58,7 +60,7 @@ const ColorBox = ({ color }: { color: string }) => (
 function getPaletteGroups(theme: Theme) {
   const paletteGroups = (Object.keys(theme.palette) as (keyof Palette)[])
     .filter((name) => isObject(theme.palette[name]))
-    .filter((name) => name !== "mode")
+    .filter((name) => !NON_COLOR_PROPERTIES.includes(name))
     .map((name) => {
       const colorGroup = theme.palette[name] as Record<string, string>;
       const colorGroupKeys = Object.keys(colorGroup);
--- a/packages/ui-components/src/stories/Palette.stories.tsx
+++ b/packages/ui-components/src/stories/Palette.stories.tsx
@@ -1,49 +1,18 @@
 import React from "react";
-
 import { Box, Grid, Palette, Stack, Typography } from "@mui/material";
 import isObject from "lodash/isObject";
-
-import theme from "../styles/theme";
+import darkTheme from "../styles/theme/dark-theme";
+import lightTheme from "../styles/theme";
+import { Theme } from "@mui/material";
 
 export default {
   title: "Design System/Palette",
 };
 
-// Filter out the palette attributes that don't correspond to colors,
-// such as functions or numeric constants. The `mode` attribute is
-// a string to name themes (dark, light, highContrast), but not a color,
-// so we filter that out as well.
-const paletteGroups = (Object.keys(theme.palette) as (keyof Palette)[])
-  .filter((name) => isObject(theme.palette[name]))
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/stories/Palette.stories.tsx
```diff
@@ -75,9 +44,58 @@ const ColorBox = ({ color }: { color: string }) => (
...
       backgroundColor: color,
     }}
   />
 );
+
+// Filter out the palette attributes that don't correspond to colors,
+// such as functions or numeric constants. The `mode` attribute is
+// a string to name themes (dark, light, highContrast), but not a color,
+// so we filter that out as well.
+function getPaletteGroups(theme: Theme) {
```
faihegberg, line 60: Turning this into a function so we can easily pass a theme to it and it will return the corresponding palette group. So we don't need to repeat code.

packages/ui-components/src/styles/theme/dark-theme.tsx
```diff
@@ -68,11 +77,79 @@ const darkThemeConfig = {
...
+    overline: {
+      color: "#fff",
+    },
   },
   components: {
+    MuiButton: {
+      styleOverrides: {
+        containedPrimary: ({ theme }: { theme: Theme }) => ({
+          "&: hover": {
+            backgroundColor: theme.palette.grey[500],
```
faihegberg, line 95: These color decisions are just to make the component more visible/presentable for now. We'd probably want to consult Josh for final approval.

packages/ui-components/src/styles/theme/dark-theme.tsx
```diff
@@ -68,11 +77,79 @@ const darkThemeConfig = {
...
+          ":first-of-type": {
+            borderTop: "none",
+          },
+          ":last-of-type": {
+            borderBottom: "none",
+          },
+          "&.Mui-selected": {
+            ...theme.typography.labelSmall,
+            borderRight: `solid 2px ${theme.palette.common.white}`,
+            backgroundColor: "transparent",
```
faihegberg, line 148: This is pretty repetitive since most of this is taken from the standard (i.e. light) theme. The only difference is the border color under `&.Mui-selected`. However, leaving any property out would revert the whole selector (in this case, `groupedHorizontal` and `groupedVertical` to revert to MUI's default. It seems if a CSS property isn't present in dark theme, it won't check whether it is present in light theme, but will just go straight back to MUI's default. If you can think of a way to not repeat this, let me know!

packages/ui-components/src/stories/Palette.stories.tsx
```diff
@@ -1,50 +1,19 @@
...
-    const colorGroup = theme.palette[name] as Record<string, string>;
-    const colorGroupKeys = Object.keys(colorGroup);
-    const colors = colorGroupKeys.map((key) => {
-      return {
-        name: `theme.palette.${name}.${key}`,
-        color: colorGroup[key],
-      };
-    });
-    return { name, colors };
-  });
```
faihegberg, line 29: This is simple converted to a function

packages/ui-components/src/styles/theme/dark-theme.tsx
pnavarrc: We can revisit this after the break, maybe there is a way to use "named" colors, similar to "primary") instead of accessing `white`, `black` or `grey[300]`, but that might need a bit of extra investigation

packages/ui-components/src/stories/Palette.stories.tsx
```diff
@@ -75,9 +44,58 @@ const ColorBox = ({ color }: { color: string }) => (
...
 );
+
+// Filter out the palette attributes that don't correspond to colors,
+// such as functions or numeric constants. The `mode` attribute is
+// a string to name themes (dark, light, highContrast), but not a color,
+// so we filter that out as well.
+function getPaletteGroups(theme: Theme) {
+  const paletteGroups = (Object.keys(theme.palette) as (keyof Palette)[])
+    .filter((name) => isObject(theme.palette[name]))
+    .filter((name) => name !== "mode")
```
pnavarrc: Consider filtering out `getContrastText` and `augmentColor` (I think those are functions). You can just exclude them by name as done here with `"mode"`

image

pnavarrc (approved):
Looks good! Added a small suggestion, but looks good to ship

PR #494, opened 2022-12-16
Documentation cleanup pt. 2

chasulin:
A continuation of documentation cleanup in #491. Changes made in this PR:
- Remove the redundant type declarations from after `@param` (replaces with ` - ` as used in [the example in TSDoc docs](https://tsdoc.org/pages/tags/returns/)
- Add `@returns` documentation for functional props

For now, I've nixed the plan of replacing `@default` with `@defaultValue` and `@param` with `@typeParam` (in the case of generics) because these changes were making the values appear as regular comments and not as highlighted values (see screenshot below). We can discuss this more after break, when I write up our conventions and put them into a readme of some sort.

Screen Shot 2022-12-15 at 7 55 36 PM

Screen Shot 2022-12-15 at 7 55 46 PM

### DIFF
--- a/packages/ui-components/src/components/Axis/Axis.style.tsx
+++ b/packages/ui-components/src/components/Axis/Axis.style.tsx
@@ -8,14 +8,14 @@ import typography from "../../styles/theme/typography";
 
 export type AxisLeftProps = React.ComponentProps<typeof VxAxisLeft> & {
   /**
-   * The className applied to the outermost axis group element.
+   * ClassName applied to the outermost axis group element.
    */
   className?: string;
 };
 
 export type AxisBottomProps = React.ComponentProps<typeof VxAxisBottom> & {
   /**
-   * The className applied to the outermost axis group element.
+   * ClassName applied to the outermost axis group element.
    */
   className?: string;
   /**
--- a/packages/ui-components/src/components/Axis/Axis.style.tsx
+++ b/packages/ui-components/src/components/Axis/Axis.style.tsx
@@ -8,14 +8,14 @@ import typography from "../../styles/theme/typography";
 
 export type AxisLeftProps = React.ComponentProps<typeof VxAxisLeft> & {
   /**
-   * ClassName applied to the outermost axis group element.
+   * Class name applied to the outermost axis group element.
    */
   className?: string;
 };
 
 export type AxisBottomProps = React.ComponentProps<typeof VxAxisBottom> & {
   /**
-   * ClassName applied to the outermost axis group element.
+   * Class name applied to the outermost axis group element.
    */
   className?: string;
   /**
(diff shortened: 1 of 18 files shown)


packages/ui-components/src/components/Axis/Axis.style.tsx
```diff
@@ -8,14 +8,14 @@ import typography from "../../styles/theme/typography";
 
 export type AxisLeftProps = React.ComponentProps<typeof VxAxisLeft> & {
   /**
-   * The className applied to the outermost axis group element.
+   * ClassName applied to the outermost axis group element.
```
mikelehen: Is "ClassName" a thing?  By default I'd expect "The class name" or something?

packages/ui-components/src/components/Axis/Axis.style.tsx
```diff
@@ -8,14 +8,14 @@ import typography from "../../styles/theme/typography";
...
    */
   className?: string;
 };
 
 export type AxisBottomProps = React.ComponentProps<typeof VxAxisBottom> & {
   /**
-   * The className applied to the outermost axis group element.
+   * ClassName applied to the outermost axis group element.
```
mikelehen: ditto

packages/ui-components/src/components/MetricCatalogContext/MetricCatalogContext.tsx
```diff
@@ -45,8 +45,8 @@ const MetricCatalogContext = createContext<MetricCatalog>(defaultMetricCatalog);
  *   // ...render component
  * ```
  *
- * @param metricCatalog The MetricCatalog instance that we want to make available.
- * @param children The component tree that we want to have access to the metric catalog.
+ * @param metricCatalog - The MetricCatalog instance that we want to make available.
+ * @param children - The component tree that we want to have access to the metric catalog.
  * @returns React.ContextProvider with the given metricCatalog
```
mikelehen, line 61: Unrelated to this PR, but should this comment be above `MetricCatalogProvider` instead of `MetricCatalogPRoviderProps` ?  I feel like something's out-of-whack here.

mikelehen (approved):
Looks good with some minor questions!

packages/ui-components/src/components/Axis/Axis.style.tsx
chasulin: oh good call, updating

packages/ui-components/src/components/MetricCatalogContext/MetricCatalogContext.tsx
chasulin: ooh yes you're right. i just rearranged so the ordering is now props, component documentation, component

PR #496, opened 2023-01-02
Add ability to have footer content under metrics in MetricChartBlock.

mikelehen:
Needed for https://github.com/covid-projections/act-now-template/issues/61

### DIFF
--- a/packages/ui-components/src/components/MetricChartBlock/MetricChartBlock.tsx
+++ b/packages/ui-components/src/components/MetricChartBlock/MetricChartBlock.tsx
@@ -36,7 +36,7 @@ export interface MetricChartBlockProps {
    * Function that returns footer content (e.g. explanatory text, share buttons,
    * etc.) to be displayed under the chart for a given metric.
    *
-   * @param metric Metric for which to render footer content.
+   * @param metric - Metric for which to render footer content.
    * @returns Footer content to render.
    */
   renderChartFooter?: (metric: Metric) => React.ReactNode;
--- a/packages/ui-components/src/components/MetricChartBlock/MetricChartBlock.tsx
+++ b/packages/ui-components/src/components/MetricChartBlock/MetricChartBlock.tsx
@@ -2,7 +2,7 @@ import { useState } from "react";
 import React from "react";
 
 import { TabContext, TabList, TabPanel } from "@mui/lab";
-import { Stack, Typography } from "@mui/material";
+import { Box, Stack, Typography } from "@mui/material";
 
 import { assert } from "@actnowcoalition/assert";
 import { Metric } from "@actnowcoalition/metrics";
@@ -31,6 +31,15 @@ export interface MetricChartBlockProps {
    * Height of the chart block.
    */
   height?: number;
+
+  /**
+   * Function that returns footer content (e.g. explanatory text, share buttons,
+   * etc.) to be displayed under the chart for a given metric.
+   *
+   * @param metric Metric for which to render footer content.
+   * @returns Footer content to render.
+   */
+  renderChartFooter?: (metric: Metric) => React.ReactNode;
 }
 
 const TabContent = ({ metric, region }: { metric: Metric; region: Region }) => {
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -785,7 +785,7 @@
   dependencies:
     "@babel/helper-plugin-utils" "^7.10.4"
 
-"@babel/plugin-syntax-jsx@^7.12.13", "@babel/plugin-syntax-jsx@^7.18.6":
+"@babel/plugin-syntax-jsx@^7.17.12", "@babel/plugin-syntax-jsx@^7.18.6":
```
mikelehen, line 788: These changes are from https://github.com/covid-projections/act-now-packages/pull/479/files#diff-ed54e778b1263cc888fe4a18544602094a59e7a7b57e749729088f9fc2d8ef6f I think.  They're unrelated to my changes.

packages/ui-components/src/components/MetricChartBlock/MetricChartBlock.tsx
```diff
@@ -31,6 +31,15 @@ export interface MetricChartBlockProps {
    * Height of the chart block.
    */
   height?: number;
+
+  /**
+   * Function that returns footer content (e.g. explanatory text, share buttons,
+   * etc.) to be displayed under the chart for a given metric.
+   *
+   * @param metric Metric for which to render footer content.
```
chasulin: tiny nit - [in this PR](https://github.com/covid-projections/act-now-packages/pull/494), i updated our documentation to use syntax similar to TSDoc, which includes a ` - ` ([example](https://github.com/covid-projections/act-now-packages/pull/494/files#diff-8095341789e8a7c2a55c1336b1113c530eef54dda9d885c53dffa689b9ceeaa2R44)). see this line of the PR description:

> Remove the redundant type declarations from after @param (replaces with - as used in [the example in TSDoc docs](https://tsdoc.org/pages/tags/returns/)

im writing up these conventions into a readme of sorts today

mikelehen: Oops, yes! Thanks for the reminder, fixed.

packages/ui-components/src/components/MetricChartBlock/MetricChartBlock.tsx
```diff
@@ -99,6 +109,9 @@ export const MetricChartBlock = ({
                 width={width}
                 height={height}
               />
+              <Box width={width} mt={3}>
+                {renderChartFooter(metric)}
+              </Box>
```
chasulin: suggestion: instead of adding a Box, we can wrap the content of TabPanel in a Stack with `spacing={3}`, so if more items are added into the panel at any point, the spacing will apply without having to add additional wrappers.

theres also something funky happening with spacing/alignment of the whole chart block (which doesn't totally have to do with your updates, so feel free to merge your pr and someone can fix spacing in a follow-up).

see here with added red border for clarity:

Screen Shot 2023-01-02 at 10 25 15 AM

a few things: 

theres [this paddingLeft on the tabs list](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/components/MetricChartBlock/MetricChartBlock.tsx#L80) that we can nix, since the tabs should probably sit flush to the side of the chart block:

Screen Shot 2023-01-02 at 10 28 17 AM

nixing the Box and wrapping the panel content in a Stack actually seems to get rid of the issue of the chart footer spilling outside of the chart block container, but the chart itself still has that issue. it looks like this is because the TabPanel has 24px of padding built into it, which isn't taken into account [when we pass width](https://github.com/covid-projections/act-now-packages/pull/496/files#diff-00d76e407ee189ac748109ce9b458bcadf444d90eed26586e4ab7d05bc858042R109) into `MetricChart` a few lines up. to fix as is, we can change the MetricChart's `width` prop to be `width={width - 48}` (prob with an explanatory comment that `48` is `padding*2` of the TabPanel)

chasulin: actually! scratch that. if we're using CAN as a guide, it looks like all the tab content (tab list + tab panel content) sits flush to the edges of the chart block container (screenshot below), and we leave padding considerations to the parent container/page spacing. so instead, here, we can keep the width prop of MetricChart as is (`width={width}`) and just add `sx={{p:0}}` to TabPabel.

Screen Shot 2023-01-02 at 10 44 41 AM

chasulin (approved):

PR #499, opened 2023-01-03
Apply missed CR Feedback from #496: Remove padding and increase width.

mikelehen:
See https://github.com/covid-projections/act-now-packages/pull/496#discussion_r1060103404

@chasulin Sorry, even though you mentioned the 2-part feedback in Slack I somehow missed the second part and merged without it.  This fixes it.


chasulin (approved):
🚀

PR #500, opened 2023-01-03
Use AutoWidth rather than requiring a width passed to MetricChartBlock.

mikelehen:


packages/ui-components/src/components/MetricChartBlock/MetricChartBlock.tsx
```diff
@@ -103,12 +99,14 @@ export const MetricChartBlock = ({
...
-                  metric={metric}
-                  region={region}
-                  width={width}
-                  height={height}
-                />
+                <AutoWidth>
+                  <MetricChart
+                    metric={metric}
+                    region={region}
+                    width={/*let auto-width set it*/ 0}
```
mikelehen, line 106: This feels a little hacky, but I'm guessing it's the only way to make it work.  The other option is to use `ParentSize` directly.

chasulin (commented):
this seems to work! (which is slightly confusing to me..[this line of documentation in AutoWidth](https://github.com/covid-projections/act-now-packages/blob/develop/packages/ui-components/src/components/AutoWidth/AutoWidth.tsx#L21-L22) needs to be adjusted, as i thought that the child having a hardcoded with would mean that AutoWidth's width wouldnt be applied). 

an idea- maybe we move the implementation of AutoWidth right into the chart files, so theyre always responsive. we did this for CAN charts ([example](https://github.com/covid-projections/covid-projections/blob/develop/src/components/Charts/ChartCaseDensity.tsx#L180-L199)) and it seemed to work out well-- it makes the charts responsive by default, and if we ever want a fixed-width/non-responsive chart, we could just wrap the implementation in a Box with a declared width.

cc @pnavarrc - thoughts?

mikelehen:
@chasulin Oops, I should have updated that comment.  My code changes it so `AutoWidth` will still apply its width if width is set to zero in the child component.

chasulin:
> @chasulin Oops, I should have updated that comment. My code changes it so `AutoWidth` will still apply its width if width is set to zero in the child component.

ooooh i see, got it. thanks for clarifying 🙏🏼

pnavarrc:
@chasulin I think we need to explore a few alternatives, because `AutoWidth` doesn't work well with containers that don't have an intrinsic width (flex boxes), and having children imposing restrictions on parent (the parent needs to have a width) components is kind of an anti-pattern.

I'm going to outline a few options so we can discuss during standup.

pnavarrc (approved):
Looks good! (thanks @chasulin for reviewing too)

chasulin:
@pnavarrc yeah, i meant wrapping charts in AutoWidth (which are composed svg's -- not all components, and not components that use flexbox/dont have intrinsic width). so, similar to that CAN example that I linked above ([here](https://github.com/covid-projections/covid-projections/blob/develop/src/components/Charts/ChartCaseDensity.tsx#L180-L199))

PR #512, opened 2023-01-04
Restore d3-scale-chromatic

pnavarrc:
So I wanted to remove `d3-scale-chromatic` because I thought it was causing the template repo to break, but it wasn't it (we have been using `d3-scale-chromatic` for a while), so I'm restoring it and will investigate what was breaking the template repo (it might have been the `fail` function that @mikelehen fixed)


mikelehen (approved):
LGTM.

PR #513, opened 2023-01-04
Compare table style fix

### LINKED ISSUE #394: Compare Table: There is a background overlap when scrolling the cells horizontally
https://user-images.githubusercontent.com/114084/201764556-6d81455c-90bf-4695-a39c-cf13b31a3660.webm

chasulin:
Closes #394 

Instead of controlling hover styling via MUI TableRow's `hover` prop (which applies a background color that is a transparent grey and causes the overlap bug seen in the gh issue), we now set the hover background color ourselves in the style file (same grey, just not transparent)

### DIFF
--- a/.changeset/hot-pots-try.md
+++ b/.changeset/hot-pots-try.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/ui-components": patch
+---
+
+fix compare table style bug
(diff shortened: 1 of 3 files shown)


mikelehen (commented):
Nicely done! Don't forget a changeset. 😁

pnavarrc (approved):

PR #529, opened 2023-01-10
Update ui-component dependencies to make it compatible with the template

pnavarrc:
Part of https://github.com/act-now-coalition/act-now-template/issues/77  - Fixing incompatible version of @mui/material between the packages and template repo.

## Changes

- https://github.com/act-now-coalition/act-now-packages/pull/529/commits/c4b9f5f82a7569e23b3b3d747964a3f14357535c Make `d3-scale-chromatic` a dev dependency. This library is ES only, which is causing problems when trying to build the template repository. Since we only use an array from it in components, I'm hardcoding the array. We still use it in Storybook, so I left it as a dev dependency.
- https://github.com/act-now-coalition/act-now-packages/pull/529/commits/8530740cc3a461e53fbe3c0b2169accbf7dfeb52 Add `encoding` as a dependency of `ui-components`. The package `node-fetch` doesn't declare its dependency on `encoding`, but not having the package breaks the build on the template repo.
- https://github.com/act-now-coalition/act-now-packages/pull/529/commits/f1cbbde4212077255bb6c5ae305ce7728dcf2bbb Upgrade TypeScript to 4.9 - This is to solve an out of memory error we were getting on the CI workflows
- https://github.com/act-now-coalition/act-now-packages/pull/529/commits/bdb19610046bbc7eb7abfc8706e95932397ae60c Upgrade @mui/material to 5.11.4 and fix some TypeScript errors (updating the types of the dark theme, adding `as typeof Box` in a few places, etc). 

## Testing

This can be tested locally using `yarn link` until we merge and publish the packages. Here is how I tested locally:

- Pull this branch locally, remove `node_modules`, run `yarn install` and `yarn build:all`
- Run `yarn link` on `packages/ui-components`
- Go to the template repo locally, pull the branch linked to https://github.com/act-now-coalition/act-now-template/pull/82 (`pablo/0110-fix-dependencies`)
- Remove `node_modules`, run `yarn` and then `yarn link @actnowcoalition/ui-components`
- Run `yarn build`

The template repo should build without issues.


packages/assert/package.json
```diff
@@ -24,7 +24,7 @@
     "build": "yarn build:esm && yarn build:cjs"
   },
   "devDependencies": {
-    "typescript": "^4.6.4"
+    "typescript": "^4.9.4"
```
mikelehen, line 27: So I'm not sure if / how much it matters but I'd kind of expect the `yarn.lock` file in each packages/<package> directory to be updated to reflect that we upgraded typescript.  i.e. you might need to run `yarn` in each package.

I'm still kind of confused / unsure if we're supposed to have yarn.lock files in each package directory, but it seems like `yarn` always creates them, so I guess so?  Not 100% sure.

packages/ui-components/package.json
```diff
@@ -41,13 +41,15 @@
     "@storybook/react": "^6.5.14",
     "@storybook/testing-library": "^0.0.13",
     "@testing-library/react": "^13.4.0",
+    "@types/d3-scale-chromatic": "^3.0.0",
     "@types/react-copy-to-clipboard": "^5.0.4",
     "@visx/mock-data": "^2.1.2",
     "babel-loader": "^8.2.5",
+    "d3-scale-chromatic": "^3.0.0",
```
mikelehen, line 48: FWIW- The primary change between 2.0.0 and 3.0.0 was switching to an ES module, so I kind of suspect we could just use 2.0.0 and things would work fine, but let's just get things working first.

mikelehen (approved):
Looks okay to me!

packages/ui-components/package.json
pnavarrc: Could be an option, or just getting rid of it entirely

packages/assert/package.json
pnavarrc: Hm, I'm going to remove all of them and re-generate them just in case. I'm also confused about that.

PR #530, opened 2023-01-10
Fix #365: Preserve 'extra' fields from metric definitions on Metric class.

### LINKED ISSUE #365: Putting extra fields into the MetricDefinition they are not properly stored in the corresponding Metric class, so they are not accessible.

mikelehen:


smcclure17 (approved):
lgtm!

PR #533, opened 2023-01-11
Make metric legend threshold more responsive

### LINKED ISSUE #334: Improve responsiveness of MetricLegendThreshold (horizontal)
The width of the bars should be automatically calculated (maybe using AutoWidth), making sure that the labels have enough space. The labels might look better if we don't allow line breaks on the side labels.

Image

faihegberg:
Currently the thermometer inside our horizontal metric legend threshold is not fully responsive. It will scale as one expands the window, but not as one shrinks the window. This is due to the automatic minimum size of a flex item (more info [here](https://github.com/airbnb/visx/issues/1014#issuecomment-1180677878)). Setting a minimum value to the parent container resolves the issue.

Closes https://github.com/act-now-coalition/act-now-packages/issues/334

Before:
image

After:
image
image

### DIFF
--- a/.changeset/empty-pumpkins-jump.md
+++ b/.changeset/empty-pumpkins-jump.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/ui-components": patch
+---
+
+Make metric legend threshold thermometer more responsive
--- a/packages/ui-components/src/components/LegendThreshold/LegendThresholdHorizontal.tsx
+++ b/packages/ui-components/src/components/LegendThreshold/LegendThresholdHorizontal.tsx
@@ -44,7 +44,7 @@ export const LegendThresholdHorizontalInner = <T,>({
   const totalHeight = height + labelsHeight + indicatorHeight;
 
   return (
-    <svg width="100%" height={totalHeight}>
+    <svg width={width} height={totalHeight}>
       {/* Indicator */}
       {getItemShowIndicator &&
         items.map((item: T, itemIndex) => {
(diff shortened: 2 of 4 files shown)


chasulin (commented):
the example with long labels currently doesnt work well on mobile (screenshot below). im also a little wary of using "100%" on svg//im not totally sure how its working. hmm. im thinking this needs a little more digging (if anything, just to figure out a solution that works for mobile)

Screen Shot 2023-01-11 at 5 20 07 PM

faihegberg:
> the example with long labels currently doesnt work well on mobile (screenshot below). im also a little wary of using "100%" on svg//im not totally sure how its working. hmm. im thinking this needs a little more digging (if anything, just to figure out a solution that works for mobile)
> 
> Screen Shot 2023-01-11 at 5 20 07 PM

What do you envision it to look like on mobile?

I can think of two options.
1. Convert to vertical legend threshold.
2. Enable the labels to wrap (this might be overcomplicating it a little bit. Currently the `startLabel` and `endLabel` props accept and `ReactNode` which I think keeps things flexible).

If we're not sure, perhaps we can consult Josh and revisit this as a follow-up.

pnavarrc:
Also if the computed width equivalent to 100% and the width passed as props don't match, then the legend is cut off. This is from [Storybook](https://act-now-packages--pr533-faihegberg-make-metr-7zbcetd3.web.app/storybook/index.html?path=/story/components-legendthreshold--horizontal-with-indicator&args=width:400), using 400px as prop

image

pnavarrc (approved):
Looks good!

PR #543, opened 2023-01-12
Add sideEffects: false to package.json

pnavarrc:
Part of https://github.com/act-now-coalition/act-now-packages/issues/537 - Enabling tree shaking for `@actnowcoalition/ui-components`

The package is free of side effects (as far as I know), not sure why this is the only package where we don't have this flag. I'm going to also add `"type": "module"` on a follow up PR (just want to test the changes one by one).


mikelehen (approved):
LGTM!

PR #544, opened 2023-01-12
Add the no-duplicate-imports rule to the ESLint config

pnavarrc:
Enabling the rule to prevent us from having unmerged imports. This is mostly a stylistic choice (I noticed a few duplicated imports while writing the tutorial), but makes our code look a bit better. (if someone doesn't like this rule we can discuss it during standup too!)

No changeset since we are not altering API/functionality


mikelehen (approved):
Nice! Definitely a fan. 👍

PR #547, opened 2023-01-17
Add a loading state to the `MetricCompareTable`

### LINKED ISSUE #473: Compare table: Better loading state
We don't have a loading state for the compare table.  We should, but I'm not sure how to get the width / height.  It looks like right now (for our storybook examples at least) we size the table through `sx` property.  Maybe we need to pass this through to `Skeleton` or something.

pnavarrc:
Close https://github.com/act-now-coalition/act-now-packages/issues/473 - Add loading state to the `MetricCompareTable` component.

See in [Storybook](https://act-now-packages--pr547-pablo-473-compare-lo-wkgipb7z.web.app/storybook/index.html?path=/story/components-metriccomparetable--loading-delay)

### DIFF
--- a/.changeset/little-walls-travel.md
+++ b/.changeset/little-walls-travel.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/ui-components": patch
+---
+
+Add a loading state to MetricCompareTable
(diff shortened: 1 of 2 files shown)


mikelehen (approved):
Looks good!

PR #548, opened 2023-01-17
Enable Storybook composition

pnavarrc:
Enable Storybook composition. With this flag, we can "include" the ui-components Storybook into the template Storybook, so users can see which components are available as they use them.

https://storybook.js.org/docs/react/sharing/storybook-composition#with-feature-flags

I also added `storybook-static` to `.gitignore`, it was missing. No changeset since it doesn't change our API

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,3 +108,4 @@ dist
 # Generated docs / storybook to be published.
 public/docs/
 public/storybook/
+storybook-static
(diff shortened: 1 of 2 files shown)


chasulin (approved):
I'm not sure how to test this, but 🚢🚢

pnavarrc:
I should have added testing instructions, my bad 😅 

```sh
yarn build-storybook
```

generates the folder `storybook-static` which should have a `stories.json` file, which is used for composition.

PR #549, opened 2023-01-17
Handle no data in metric chart

### LINKED ISSUE #541: `MultiMetricMultiRegion` chart breaks if the user selects a time period with no data
Selecting a number of days that makes the chart empty breaks the chart. Instead, it should fail gracefully. For example, in https://demo-world-happiness-report.vercel.app/ - if you select 180 days, the app crashes, because the data doesn't cover the last 180 days.

The chart should fail more gracefully

faihegberg:
Currently when our `MetricSeriesChart` encounters a timeseries with no data in it for the selected time period, it will fail ungracefully. Specifically, it will throw an internal assertion failure that's visible on the console, but not immediately apparent to the user. This PR handles this failure by ensuring the chart informs the user there's no data if it encounters a timeseries with no data for the selected time period.

Closes https://github.com/act-now-coalition/act-now-packages/issues/541

Before
image

After
image

### DIFF
--- a/packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
+++ b/packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
@@ -105,6 +105,17 @@ export const MetricSeriesChart = ({
     return <Skeleton variant="rectangular" width={width} height={height} />;
   }
 
+  try {
+    getDateRange(timeseriesList);
+    getValueRange(timeseriesList);
+  } catch (err) {
+    return (
+      <ErrorBox width={width} height={height}>
+        No data in the provided time range.
+      </ErrorBox>
+    );
+  }
+
   const seriesList = series
     .filter(({ metric, region }) => data.hasMetricData(region, metric))
     .map((seriesItem, seriesIndex) => ({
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
```diff
@@ -105,6 +105,17 @@ export const MetricSeriesChart = ({
...
+  try {
+    getDateRange(timeseriesList);
+    getValueRange(timeseriesList);
+  } catch (err) {
+    return (
+      <ErrorBox width={width} height={height}>
+        No data in the provided time range.
+      </ErrorBox>
+    );
+  }
```
faihegberg: Using a `try-catch` to keep the assertion check inside `getDateRange` and `getValueRange`. So these functions methods can still gracefully return a collection of Date and number objects, respectively.

packages/ui-components/src/components/MetricSeriesChart/MetricSeriesChart.tsx
```diff
@@ -105,6 +105,17 @@ export const MetricSeriesChart = ({
     return <Skeleton variant="rectangular" width={width} height={height} />;
   }
 
+  try {
```
pnavarrc: I think it might be better to show the error message if `timeseriesList` is empty

```tsx

if (error) {
  <Error />
} else if (timeseriesList.length === 0) {
  // Error: No data in the range
} else if (!data) {
  // Loading...
}

```

pnavarrc (commented):
I think there is a better alternative, added a note

pnavarrc:
The loading state now shows the "no data in the time range" message for a bit https://act-now-packages--pr549-faihegberg-handle-no-2xez2ang.web.app/storybook/index.html?path=/story/components-metricserieschart--loading-delay, I think the check is still not quite what we need

pnavarrc (approved):
Looks good!

PR #550, opened 2023-01-18
Fix #536: Fix alignment of icon on RegionSearch

### LINKED ISSUE #536: Fix alignment of icon on `RegionSearch`
The icon is too far to the left

Image

mikelehen:
Before:
image

After:
image

### DIFF
--- a/packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
+++ b/packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
@@ -80,6 +80,7 @@ export const RegionSearch = ({
         }
       }}
       clearIcon={<></>}
+      forcePopupIcon={false}
       renderInput={customRenderInput ?? defaultRenderInput}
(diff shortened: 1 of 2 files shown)


packages/ui-components/src/components/RegionSearch/RegionSearch.tsx
```diff
@@ -80,6 +80,7 @@ export const RegionSearch = ({
         }
       }}
       clearIcon={<></>}
+      forcePopupIcon={false}
```
mikelehen, line 83: On line 63 we replace the endAdornment with our own icon, but Material-UI normally has its own icons that it adds padding for.  By setting `forcePopupIcon` to false, we tell it not to display those icons and so it doesn't add the padding.  Based on https://stackoverflow.com/a/61258146

chasulin (approved):
ahhh bless! this was driving me nuts but i hadn't gotten to fixing it myself. 🙏🏼🙏🏼🙏🏼🙏🏼

PR #552, opened 2023-01-19
Better support for long periods for `MultiRegionMultiMetricChart`

### LINKED ISSUE #540: `MultiMetricMultiRegion` chart is not designed for yearly/monthly data
The `MultiMetricMultiRegion` chart is not designed for yearly or monthly data. The `timePeriod` prop does support any time periods, but it's kind of complicated for the user to set them up (they need to call `subtractDate`, etc). The label of the time periods selector also says "Past number of days".

One option could be to change the label to "Time period" and export a function to make it easier for the user to create their own options. For example

Maybe we should also have more options to be used with yearly/monthly data. For example, we could have something like:

```tsx
const timePeriods = [
  generatePeriod(14, TimeUnit.DAYS),
  generatePeriod(3, TimeUnit.MONTHS),
  generatePeriod(6, TimeUnit.MONTHS),
]
/**
 * Time period
 * ------------
 *   14 days
 *   3 months
 *   6 months
 *   All time
 */


<MultiRegionMultiMetricChart
  metrics={ALL_METRICS}
  regions={regions.all}
  timePeriods={timePeriods}
  initialMetric={MetricId.LIFE_LADDER}
  initialRegions={[regionA, regionB]}
  height={600}
  width={400}
/>
```

pnavarrc:
Close https://github.com/act-now-coalition/act-now-packages/issues/540 - Improve support for longer time periods for `MultiRegionMultiMetriChart`

image

Currently, the time selector on the `MultiRegionMultiMetriChart` component has "Number of days" as selector in the label, which doesn't really work for yearly data. We do allow users to pass a custom `timePeriods` list to populate the menu, but generating a list of time periods involves subtracting time and adding labels, so I added the function `createTimePeriodOption` to make it a bit easier to generate the options for the menu. This way, a user can generate a list more easily:

```tsx
const timePeriods = [
  createTimePeriodOption(1, TimeUnit.MONTH), // 1 month
  createTimePeriodOption(6, TimeUnit.MONTH), // 6 months
  createTimePeriodOption(1, TimeUnit.YEAR),  // 1 year
  createTimePeriodOption(3, TimeUnit.YEAR),  // 3 years
]
```

Also, I updated the component to ensure that the "All time" is always added to this list.

## Changes

**`time-utils`**
- Add `YEARS` to the `TimeUnit` enum so we can select intervals covering multiple years.
- Implement `getTimeUnitLabel` function to get the name of a given `TimeUnit` (For example, `getTimeUnitLabel(3, TimeUnit.MONTHS)` →  `"months"`

**`MultiRegionMultiMetricChart`**
- Implement the `createTimePeriodOption` function to allow users to easily create time periods more easily
- Update the label on the time selector from _Number of days_ → _Time period_
- Always add the "All time" option, so users don't have to

### DIFF
--- a/.changeset/chatty-ears-hang.md
+++ b/.changeset/chatty-ears-hang.md
@@ -1,0 +1,6 @@
+---
+"@actnowcoalition/ui-components": patch
+"@actnowcoalition/time-utils": patch
+---
+
+Better support for long time periods for MultiRegionMultiMetricChart
--- a/.changeset/chatty-ears-hang.md
+++ b/.changeset/chatty-ears-hang.md
@@ -1,6 +1,5 @@
 ---
 "@actnowcoalition/ui-components": patch
-"@actnowcoalition/time-utils": patch
 ---
 
-Better support for long time periods for MultiRegionMultiMetricChart
+Update `MultiRegionMultiMetricChart` to better support time intervals not based in a number of days.
--- a/packages/time-utils/src/index.ts
+++ b/packages/time-utils/src/index.ts
@@ -1,6 +1,6 @@
 import { DateTime, Duration } from "luxon";
 
-import { assert } from "@actnowcoalition/assert";
+import { assert, fail } from "@actnowcoalition/assert";
 
 export * from "./PureDate";
 
@@ -249,9 +249,8 @@ export function getTimeUnitLabel(amount: number, unit: TimeUnit): string {
     case TimeUnit.YEARS:
       return pluralize(amount, "year", "years");
     default:
-      return `${amount} ${unit}`;
+      fail(`Unsupported unit: ${unit}`);
   }
-  return `${amount} ${unit}`;
 }
 
 /**
--- a/packages/time-utils/src/index.ts
+++ b/packages/time-utils/src/index.ts
@@ -27,6 +27,7 @@ export enum TimeUnit {
   DAYS = "day",
   WEEKS = "week",
   MONTHS = "month",
+  YEARS = "year",
 }
 
 /**
@@ -160,6 +161,8 @@ function getTimeUnitOption(amount: number, unit: TimeUnit): Duration {
       return Duration.fromObject({ weeks: amount });
     case TimeUnit.MONTHS:
       return Duration.fromObject({ months: amount });
+    case TimeUnit.YEARS:
+      return Duration.fromObject({ years: amount });
   }
 }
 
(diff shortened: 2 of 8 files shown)


packages/ui-components/src/components/MultiRegionMultiMetricChart/utils.ts
```diff
@@ -27,16 +31,26 @@ export interface TimePeriod {
...
 }
+
+/**
+ * Creates a custom time period option for the MultiRegionMultiMetricChart
+ * component.
+ *
+ * @param amount - Number of units of time.
+ * @param unit - Unit of time.
+ */
+export function createTimePeriodOption(
```
mikelehen: I feel like in JavaScript it might be idiomatic to just call this `timePeriod()` but I don't feel strongly at all.

packages/time-utils/src/index.ts
```diff
@@ -223,3 +226,49 @@ export function assertDateOnly(date: Date): void {
...
+    case TimeUnit.DAYS:
+      return pluralize(amount, "day", "days");
+    case TimeUnit.WEEKS:
+      return pluralize(amount, "week", "weeks");
+    case TimeUnit.MONTHS:
+      return pluralize(amount, "month", "months");
+    case TimeUnit.YEARS:
+      return pluralize(amount, "year", "years");
+    default:
+      return `${amount} ${unit}`;
```
mikelehen: I think this should just be `return unit;`, right?  Alternatively you could `fail('Unsupported unit: ' + unit);`

packages/time-utils/src/index.ts
```diff
@@ -223,3 +226,49 @@ export function assertDateOnly(date: Date): void {
...
+  }
+  return `${amount} ${unit}`;
```
mikelehen: Do we need this?  It should get caught by the `default` switch case, no?

packages/time-utils/src/index.ts
mikelehen: As an aside, I was tempted to suggest we refactor this to be `getTimePeriodLabel()` and have it return the full "<amount> <unit(s)>" string since it seems like *usually* that will be what you want...  by could imagine maybe needing to style the amount and the units separately or something, so keeping it as `getTimeUnitLabel()` seems reasonable too.  We could also expose both. No strong feelings. 🤷

.changeset/chatty-ears-hang.md
```diff
@@ -0,0 +1,6 @@
+---
+"@actnowcoalition/ui-components": patch
+"@actnowcoalition/time-utils": patch
+---
+
+Better support for long time periods for MultiRegionMultiMetricChart
```
mikelehen: Super nit, but I think in theory we'd be putting more effort into our changeset descriptions rather than into the PR descriptions. 😁 

image

And this would maybe ideally be two changesets (one for ui-components and one for time-utils).  

I don't think it's worth bothering with for now though.  We can get more disciplined once we're launched and (hopefully) have people watching our changelog release notes.

mikelehen (approved):
A few small things, but looks good!

.changeset/chatty-ears-hang.md
pnavarrc: That's actually very reasonable, I can at least add a separated item for the changes in time-utils 👍

packages/ui-components/src/components/MultiRegionMultiMetricChart/utils.ts
pnavarrc: I'm going to call it `timePeriodOption` to avoid using the same name as the `timePeriod` prop

packages/time-utils/src/index.ts
pnavarrc: right!

packages/time-utils/src/index.ts
pnavarrc: I'm going to leave it for now - I was on the fence too

PR #554, opened 2023-01-19
Fix #538: Add a default tooltip for metric maps.

### LINKED ISSUE #538: Implement a default tooltip for maps
The MetricWorldMap (and other maps) require a `getTooltip` function that takes the regionId and returns the content of the tooltip. This prop should be optional. We could provide a good default tooltip that includes

- region name
- metric name and current value (for metric-based maps)

We should still support the `getTooltip` so users can customize the tooltip if they need to.

mikelehen:
I don't think there's a good default to add for non-metric-aware maps (displaying the region ID isn't very user-friendly) and so this only applies to metric-aware maps.

image

### DIFF
--- a/packages/ui-components/src/components/MetricUSNationalMap/MetricUSNationalMap.tsx
+++ b/packages/ui-components/src/components/MetricUSNationalMap/MetricUSNationalMap.tsx
@@ -3,14 +3,15 @@ import React from "react";
 import { useTheme } from "@mui/material";
 
 import { Metric } from "@actnowcoalition/metrics";
-import { RegionDB } from "@actnowcoalition/regions";
+import { Region, RegionDB } from "@actnowcoalition/regions";
 
 import { useDataForRegionsAndMetrics } from "../../common/hooks";
 import { ErrorBox } from "../ErrorBox";
+import { MetricMapTooltipContent } from "../MetricMapTooltipContent";
 import { USNationalMap, USNationalMapProps } from "../USNationalMap";
 
 export interface MetricUSNationalMapProps
-  extends Omit<USNationalMapProps, "getRegionUrl"> {
+  extends Omit<USNationalMapProps, "getRegionUrl" | "getTooltip"> {
   /**
    * Metric represented by the map's coloring.
    */
@@ -20,6 +21,13 @@ export interface MetricUSNationalMapProps
    * Used for generating region links, coloring the map, etc.
    */
   regionDB: RegionDB;
+  /**
+   * Function that returns tooltip content for a given region.
+   *
+   * @param region - Region for which to get tooltip content.
+   * @returns Tooltip content for the region.
+   */
+  getTooltip?: (region: Region) => React.ReactNode;
 }
 
 /**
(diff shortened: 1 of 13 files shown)


packages/ui-components/src/components/MetricUSNationalMap/MetricUSNationalMap.tsx
```diff
@@ -20,6 +21,13 @@ export interface MetricUSNationalMapProps
    * Used for generating region links, coloring the map, etc.
    */
   regionDB: RegionDB;
+  /**
+   * Function that returns tooltip content for a given region.
+   *
+   * @param region - Region for which to get tooltip content.
+   * @returns Tooltip content for the region.
+   */
+  getTooltip?: (region: Region) => React.ReactNode;
```
mikelehen, line 30: I'm replacing the `getTooltip()` from `BaseMapProps` with one that is optional and takes a `region: Region` instead of `regionId: string`.

packages/ui-components/src/components/MetricValue/MetricValue.tsx
```diff
@@ -23,12 +23,18 @@ export interface MetricValueProps {
    * @default "dataEmphasizedLarge"
    */
   variant?: TypographyProps["variant"];
+  /**
+   * MUI Typography color applied to the metric value text.
+   * @default "inherit"
+   */
+  color?: TypographyProps["color"];
```
mikelehen, line 30: Since tooltips have a black background I needed to be able to pass `color="inherit"` in.

packages/ui-components/src/components/MultiMetricUSStateMap/MultiMetricUSStateMap.tsx
```diff
@@ -33,12 +33,12 @@ export interface MultiMetricUSStateMapProps {
...
   /**
-   * Function that returns tooltip content for the region corresponding to a given regionId.
+   * Function that returns tooltip content for a given region.
    *
-   * @param regionId - RegionId of the region for which to get tooltip content.
+   * @param region - Region for which to get tooltip content.
    * @returns Tooltip content for the region.
    */
-  getTooltip?: (regionId: string) => React.ReactNode;
+  getTooltip?: (region: Region) => React.ReactNode;
```
mikelehen, line 41: Updated this to match the new definition from the other metric maps.

packages/ui-components/src/components/MetricUSNationalMap/MetricUSNationalMap.tsx
smcclure17: just out of curiosity, is there a reason for switching from `regionId` --> `Region`? Or just preference

smcclure17 (approved):
looks great! one small question

packages/ui-components/src/components/MetricUSNationalMap/MetricUSNationalMap.tsx
mikelehen: Basically giving `getTooltip()` access to the `Region` saves the consumer a step of having to look up the regionId.  For our base maps, we don't have access to the regions DB, so all we can provide is the `regionId`.  But our metric-aware maps can be more helpful!

smcclure17: ahh got it--didn't realize the base maps didn't have access to the regions--makes sense, thanks!

PR #557, opened 2023-01-19
Fix #555: Remove relativeUrl from Region.

### LINKED ISSUE #555: Remove `relativeUrl` from `Region`
Now that `RegionDB` has the `getRegionRelativeUrl` method, we should remove `regionUrl` from the `Region` object to avoid confusion.

Using `getRegionRelativeUrl` makes it easier for users to customize the region URL based on the needs of their project.

mikelehen:


faihegberg (approved):
LGTM!

PR #558, opened 2023-01-23
Tabs improvements

chasulin:
Updates (4):

1. Makes tab styling a bit more compact (uses CAN as a reference). In implementing my demo/test site (first photo), I felt that the tab content was a bit too big and not compact enough. This PR updates spacing and font-size to be slightly smaller (second photo)

Screen Shot 2023-01-23 at 2 45 45 PM

Screen Shot 2023-01-23 at 2 46 56 PM

2. Updates the bottom border of the tabs list take up the entire width of the chart block (similar to CAN)
3. Fixes the alignment syntax (`"flex-start"` -> `flex-start`) so it takes properly
4. Updates `showMetricDot` (in `MetricValue`) to resolve to false if the metric doesn't have categories. It was previously resolving to true for all metrics if there is a non-null `currentValue`, which was creating unwanted spacing even when there was no `MetricDot` rendering (photo below)

**Before:**

Screen Shot 2023-01-23 at 2 51 09 PM

**After:** 

Screen Shot 2023-01-23 at 2 51 26 PM


packages/ui-components/src/components/MetricChartBlock/MetricChartBlock.style.ts
```diff
@@ -3,9 +3,11 @@ import { Tab } from "@mui/material";
 import { styled } from "../../styles";
 
 export const MetricTab = styled(Tab)`
-  align-items: "flex-start";
+  align-items: flex-start;
```
mikelehen, line 6: Heh, nice. Good catch.

mikelehen (approved):
LGTM except you're checking in packages/ui-components/yarn.lock which wasn't checked in before.  I think @pnavarrc recently removed these from git (we used to check them in).  TBH I'm not 100% sure if we are supposed to check them in in a monorepo or not.

If not, then we should probably add them to a .gitignore somewhere to prevent accidentally checking them in.

PR #569, opened 2023-01-26
Update maps to better handle regions without a URL

pnavarrc:
Some regions in the map don't have a corresponding region in the database, which was causing the app to crash since we were trying to create `Link` items to regions without a valid `href` attribute. This PR updates the maps to not create a link when we don't have the url for a given region.

### DIFF
--- a/.changeset/cold-candles-reflect.md
+++ b/.changeset/cold-candles-reflect.md
@@ -2,4 +2,4 @@
 "@actnowcoalition/ui-components": patch
 ---
 
-Improve WorldMap to better handle regions without URLs
+Improve maps to better handle missing region URLs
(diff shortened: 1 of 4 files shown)


mikelehen (approved):
This looks okay, though two thoughts:

1. I wonder if just doing `getRegionUrl(...) ?? ""` or something would be sufficient rather than omitting the `Link` entirely.
2. Really I think it's a more fundamental bug that our maps and our regions database aren't in agreement about valid region ids.  I don't know how hard it is to fix that though.

pnavarrc:
Good points!

1. The problem happens in the template, because we are overriding MUI Link to use the Next Link. I'm surprised by the fact that the app will crash just because a Link is missing the href (but it did)
2. It shouldn't be hard to fix. One way to do it would be to just filter out the shapes that don't match regions in the provided database. Other objects (disputed territories and the like) shouldn't cause conflicts.

PR #571, opened 2023-01-26
Add `ComponentLoaded` component and `useMutationObserver` hook

smcclure17:
`ComponentLoaded` contains a class that should be used to signal that the parent component is fully loaded. It is appended to Metric-aware components when the data is finished fetching. This way, our share image/screenshot system knows when the page is ready to capture. 

The `useMutationObserver` hook is used to "watch" specific components in order to tell when the `component-loaded` class has been appended to it. 

In concert with: https://github.com/act-now-coalition/act-now-template/pull/98


packages/ui-components/src/common/hooks/useMutationObserver.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Hook to observe for mutations on a DOM element.
+ *
+ * Interpolated from https://tobbelindstrom.com/blog/useMutationObserver/
+ *
+ * @param ref Reference object to observe for mutations.
+ * @param callback Callback function to execute when a mutation occurs.
+ * @param options Options to pass to the MutationObserver.
+ */
+export function useMutationObserver(
```
smcclure17, line 19: This hook is used by https://github.com/act-now-coalition/act-now-template/pull/98 to target class name changes for only a specific component. This way, we can wait until this specific component has finished loading before taking a screenshot

packages/ui-components/src/common/hooks/useMutationObserver.ts
```diff
@@ -0,0 +1,38 @@
...
+    if (observer && target) {
+      observer.observe(target, options);
+      // Disconnect/shutdown observer on hook cleanup.
+      return () => observer.disconnect();
+    }
+  }, [ref, observer, options]);
+}
+
+export const isSSR = !(
+  typeof window !== "undefined" && window.document?.createElement
```
mikelehen: Does this need to be exported?  I think not.  If we are exporting it on purpose, it should have a doc comment.

It'd be kind of nice to have a comment regardless (I'm not 100% sure what this is doing / why), but I guess this came from the example you cited, so maybe we don't know exactly what it's for.

packages/ui-components/src/common/hooks/useMutationObserver.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @param callback Callback function to execute when a mutation occurs.
+ * @param options Options to pass to the MutationObserver.
+ */
+export function useMutationObserver(
+  ref: RefObject<Element> | null,
+  callback: MutationCallback,
+  options?: MutationObserverInit
+): void {
+  const observer = useMemo(
+    () => (isSSR ? null : new MutationObserver(callback)),
```
mikelehen, line 26: Can you move isSSR to the top of this file so it's defined before it's used?  I think that's a little safer / cleaner.

packages/ui-components/src/common/hooks/useMutationObserver.ts
```diff
@@ -0,0 +1,38 @@
...
+    [callback]
+  );
+
+  if (!options) {
+    options = {};
+  }
```
mikelehen: nit: could do `options = options ?? {}` ... or just replace the usage of `options` below with `options ?? {}`

packages/ui-components/src/components/ComponentLoaded/ComponentLoaded.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { Box } from "@mui/material";
+
+/**
+ * Empty component to be appended to another component to indicate that it has fully loaded.
+ *
+ * Used for share image generation to ensure we wait until the component has loaded
+ * before capturing the image.
+ */
+export const ComponentLoaded = () => {
+  return <Box className="component-loaded" style={{ display: "none" }}></Box>;
```
mikelehen: Maybe prefix with "act-now-" or something since this is a global class name and we want to avoid conflicting with any class names the consumer is using.

packages/ui-components/src/components/MetricSparklines/MetricSparklines.tsx
```diff
@@ -100,13 +101,18 @@ export const MetricSparklines = ({
...
-      />
+      <>
+        <SparkLine
+          timeseriesLineChart={timeseriesLineChart}
+          timeseriesBarChart={timeseriesBarChart}
+          minValue={minValue}
+          maxValue={maxValue}
+          {...optionalProps}
+        />
+        {timeseriesBarChart.hasData() && timeseriesLineChart.hasData() && (
```
mikelehen: Why isn't this `data && ...` like the others?

I'm not sure, but it seems like this could cause the screenshot to hang / never render if the metric doesn't have any data.

mikelehen (approved):
LGTM with a couple nits and one potential flag.

packages/ui-components/src/common/hooks/useMutationObserver.ts
smcclure17: Oop, yeah this doesn't need to be exported. And yeah, I'm not 100% sure, but my assumption is that with serverside rendering the `window` object isn't included in the initial render b/c that's a client thing, and it gets added once the client side content loads, but again, I'm not sure. I'll look to get clarity on this and will add a comment 🙏

packages/ui-components/src/components/MetricSparklines/MetricSparklines.tsx
smcclure17: I think I was following the asserts/checks further up in the file, but I don't remember why... Checking if `data` exists should be enough, so I'll change this to match the others

PR #573, opened 2023-01-30
Merge packages into @actnowcoalition/actnow.js

mikelehen:
- Package name: `@actnowcoalition/actnow.js` 
    - I couldn’t get subpath exports to work (so you have to do `import { Region } from` `"``@actnowcoalition/actnow.js``"` not `import { Region } from "@actnowcoalition/actnow.js/regions"`  I think it’s not possible to have subpath exports if you’re exposing CJS and ESM modules.
    - I had to change all internal imports to relative paths (`"``../../../metrics``"` instead of `"``@actnowcoalition/metrics``"`)
    - Removed `plop create-package` stuff.

You can probably review this commit by commit with one caveat: Originally I tried to keep "absolute" imports working locally so that we could do `import { Metric } from "src/metrics"` but this didn't work out and so I later switched to all relative imports.  You'll see some funky changes to storybook / jest done and then undone if you look at the commits in order.


.storybook/main.ts
```diff
@@ -0,0 +1,24 @@
...
+  },
+  babel: async (options: any) => {
+    // NOTE(michael 2023-01-30): By default Storybook adds a bunch of babel
+    // plugins that end up causing syntax errors with some TypeScript features
+    // (in particular constructor properties like `constructor(private foo:
+    // number) { }`). I eventually just tried removing them so we get babel
+    // defaults, and everything worked. Maybe because we're using a recent
+    // version of babel we don't need all of those plugins.
+    options["plugins"] = [];
+    return options;
```
mikelehen, line 22: Worth noting that this was a tweak I had to make to the storybook config once storybook started trying to compile all of our code (metrics, regions, etc.) instead of relying on the pre-built packages.  

It was a preexisting issue with our storybook config though.  You can repro the issue in `develop` by adding a dummy class like this to any file in ui-components:

```
class Foo {
  constructor(private bar: number) { }
}
```

tsconfig.build.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./tsconfig",
```
mikelehen, line 2: The root tsconfig.json is used by storybook and jest and it includes tests, stories, etc.  This one is used for the actual build and creates d.ts files and excludes stories / tests.

pnavarrc (approved):
Looks good!

PR #577, opened 2023-01-31
Add ACEEE cities Region database

smcclure17:
On the fence about whether we should indicate that these are the cities specifically used by the ACEEE framework (e.g. by naming the RegionDb `aceeeCitiesDb`) since there are only 100 of them, or if we can just add cities to this on an as-needed basis.

### DIFF
--- a/src/regions/datasets/us/cities.json
+++ b/src/regions/datasets/us/cities.json
@@ -1,0 +1,902 @@
+[
+  {
+    "regionId": "1600000US3901000",
+    "shortName": "Akron",
+    "state": "OH",
+    "population": 189347,
+    "coordinates": [41.08, -81.52],
+    "fullName": "Akron, OH",
+    "abbreviation": "Akron"
+  },
+  {
+    "regionId": "1600000US3502000",
+    "shortName": "Albuquerque",
+    "state": "NM",
+    "population": 562599,
+    "coordinates": [35.1, -106.64],
+    "fullName": "Albuquerque, NM",
+    "abbreviation": "Albuquerque"
+  },
+  {
+    "regionId": "1600000US4202000",
+    "shortName": "Allentown",
+    "state": "PA",
+    "population": 125944,
+    "coordinates": [40.59, -75.47],
+    "fullName": "Allentown, PA",
+    "abbreviation": "Allentown"
+  },
+  {
+    "regionId": "1600000US1304000",
+    "shortName": "Atlanta",
+    "state": "GA",
+    "population": 496461,
+    "coordinates": [33.76, -84.42],
+    "fullName": "Atlanta, GA",
+    "abbreviation": "Atlanta"
+  },
+  {
+    "regionId": "1600000US1304204",
+    "shortName": "Augusta",
+    "state": "GA",
+    "population": 201196,
+    "coordinates": [33.36, -82.07],
+    "fullName": "Augusta, GA",
+    "abbreviation": "Augusta"
+  },
+  {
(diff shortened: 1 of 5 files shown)


src/regions/datasets/us/cities_db.ts
```diff
@@ -0,0 +1,19 @@
...
+import RegionDB from "../../RegionDB";
+import citiesJson from "./cities.json";
+import statesDB from "./states_db";
+
+const metros = citiesJson.map((city) => {
+  return new Region(
+    city.regionId.split("US")[1],
+    city.fullName,
+    city.shortName,
+    city.abbreviation,
```
smcclure17, line 11: I looked into using airport codes for abbreviations but:
- some of the abbreviations wouldn't make sense (e.g. New York would either be LGA for LaGuardia or JFK)
- not all cities have an airport code.

so it would have been a fair amount of manual work to find and use the best abbreviations for each city. I'm happy to revisit this if we'd like, but I think we can probably punt on it until we have a real need for abbreviations

src/regions/datasets/us/cities.json
```diff
@@ -0,0 +1,902 @@
+[
```
smcclure17, line 1: definitions combined from https://github.com/act-now-coalition/who-demo/blob/climate-act-now/src/common/data/cities.json and https://www.dropbox.com/s/3we8dyoysy6e0hl/CS_Metric_Master_ActNow.xlsx

src/regions/datasets/us/cities_db.ts
```diff
@@ -0,0 +1,19 @@
+import { Region } from "../../Region";
+import RegionDB from "../../RegionDB";
+import citiesJson from "./cities.json";
+import statesDB from "./states_db";
+
+const metros = citiesJson.map((city) => {
+  return new Region(
+    city.regionId.split("US")[1],
```
smcclure17, line 8: the geoIds from the [metric data spreadsheet](https://www.dropbox.com/s/3we8dyoysy6e0hl/CS_Metric_Master_ActNow.xlsx) have long prefixes (e.g. `1600000US3901000`) so I'm stripping that and using just what seems to be the following 7-digit unique code (e.g. `3901000`)

src/regions/datasets/us/cities_db.ts
mikelehen: Yeah, sounds good. I'm not sure there's any abbreviation we could use that people would actually know / recognize, so no abbreviation sounds good.

src/regions/datasets/us/cities_db.ts
```diff
@@ -0,0 +1,19 @@
...
+    city.fullName,
+    city.shortName,
+    city.abbreviation,
+    Region.toSlug(city.fullName),
+    statesDB.all.find((state) => state.abbreviation === city.state) ?? null, // should never be null
```
mikelehen, line 13: Can we factor this out into a variable and then assert that it's not null?

mikelehen (approved):
Looks good, but maybe we can chat during standup about whether this is okay in act-now-packages if it's somewhat specific to ACEEE.

I think ideally we'd expand this list to some meaningful set (top 500 cities in the US or something) and then in the ACEEE repo we'd subset down to the ones we care about.  But I'm not 100% sure.  If these are the top 100 cities in the US, I feel more okay with putting them here. 🤷

smcclure17:
Per standup, will address comments and move to [`climate-act-now`](https://github.com/act-now-coalition/climate-act-now)

PR #579, opened 2023-01-31
Add support CSVs in long format

smcclure17:
E.g. to ingest data for [Climate Act Now](https://www.dropbox.com/s/3we8dyoysy6e0hl/CS_Metric_Master_ActNow.xlsx):

```ts
// Untested, but this should be the general workflow/setup
const provider = new CsvDataProvider("climate-long-csv, {
    url: '/path/to/file.csv',
    regionDb: regions,
    regionColumn: "GeoID",
    longFormatCsvOptions: {
        variableColumn: "Metric",
        valueColumn: "Score"
    }
  })


const climateMetric = new Metric({
  id: MetricId.SOME_METRIC,
  dataReference: { 
    providerId: "climate-long-csv", 
    variable: "Affordability ..." } // Where 'variable' is an entry from the 'Metric' column in the xlsx
})

...
```

### DIFF
--- a/src/metrics/data_providers/CsvDataProvider.test.ts
+++ b/src/metrics/data_providers/CsvDataProvider.test.ts
@@ -122,8 +122,17 @@ describe("CsvDataProvider", () => {
       longFormatCsvProviderOptions
     );
 
-    expect(wideMetricData).toStrictEqual(longMetricData);
-    expect(wideMetricDataTs).toStrictEqual(longMetricDataTs);
+    // Can't use toStrictEqual() because long-format CSVs use 'variable' instead of 'column' in dataReference.
+    expect(wideMetricData.currentValue).toBe(longMetricData.currentValue);
+    expect(wideMetricData.hasTimeseries).toStrictEqual(
+      longMetricData.hasTimeseries
+    );
+    expect(wideMetricDataTs.currentValue).toStrictEqual(
+      longMetricDataTs.currentValue
+    );
+    expect(wideMetricDataTs.timeseries).toStrictEqual(
+      longMetricDataTs.timeseries
+    );
   });
 });
 
@@ -138,13 +147,13 @@ describe("CsvDataProvider with wide-format (default) csv", () => {
       /*includeTimeseries=*/ true,
       /*dateColumn=*/ "date"
     );
-    expect(metricDataNoTs.currentValue).toBe(150);
+    expect(metricDataNoTs.currentValue).toBe(120);
     expect(metricDataNoTs.hasTimeseries()).toBe(false);
 
     expect(metricDataTs.hasTimeseries()).toBe(true);
     expect(metricDataTs.timeseries.length).toBe(1);
-    expect(metricDataTs.timeseries.lastValue).toBe(150);
-    expect(metricDataTs.currentValue).toBe(150);
+    expect(metricDataTs.timeseries.lastValue).toBe(120);
+    expect(metricDataTs.currentValue).toBe(120);
   });
 
   test("fetchData() returns non-timeseries data if timeseries data is not available.", async () => {
@@ -152,7 +161,7 @@ describe("CsvDataProvider with wide-format (default) csv", () => {
       wideCsv,
       /*includeTimeseries=*/ true
     );
-    expect(metricData.currentValue).toBe(150);
+    expect(metricData.currentValue).toBe(120);
     expect(metricData.hasTimeseries()).toBe(false);
   });
 
@@ -175,7 +184,7 @@ describe("CsvDataProvider with wide-format (default) csv", () => {
   });
 });
 
-describe.only("CsvDataProvider with long-format CSV", () => {
+describe("CsvDataProvider with long-format CSV", () => {
   test("fetchData() yields expected data", async () => {
     const metricDataNoTs = await testFetchingCsvData(
       longCsv,
@@ -195,7 +204,7 @@ describe.only("CsvDataProvider with long-format CSV", () => {
     expect(metricDataNoTs.hasTimeseries()).toBe(false);
 
     expect(metricDataTs.hasTimeseries()).toBe(true);
-    expect(metricDataTs.timeseries.length).toBe(2);
+    expect(metricDataTs.timeseries.length).toBe(1);
     expect(metricDataTs.timeseries.lastValue).toBe(120);
     expect(metricDataTs.currentValue).toBe(120);
   });
@@ -223,16 +232,4 @@ describe.only("CsvDataProvider with long-format CSV", () => {
       )
     ).rejects.toThrow();
   });
-
-  // test("fetchData() fails if no data exists for the region", async () => {
-  //   expect(async () =>
-  //     testFetchingCsvData(
-  //       longCsv.replaceAll("36", "25"), // replace NY locations with MA so there's no data for NY.
-  //       /*includeTimeseries=*/ true,
-  //       /*dateColumn=*/ undefined,
-  //       /*metric=*/ testLongMetric,
-  //       longFormatCsvProviderOptions
-  //     )
-  //   ).rejects.toThrow("No data found for region");
-  // });
 });
--- a/src/metrics/data_providers/CsvDataProvider.test.ts
+++ b/src/metrics/data_providers/CsvDataProvider.test.ts
@@ -17,8 +17,8 @@ const csvTimeseries = `region,date,cool_metric
 const longCsv = `region,metric,value,another_column
 12,cool_metric,100,a
 12,another_metric,110,b
-36,cool_metric,120,d
-36,another_metric,130,c`;
+36,another_metric,130,c
+36,cool_metric,120,d`;
 
 const longCsvTimeseries = `region,date,metric,value,another_column
 12,2022-08-02,cool_metric,70,a
@@ -184,6 +184,18 @@ describe("CsvDataProvider with long-format CSV", () => {
     expect(metricData.hasTimeseries()).toBe(false);
   });
 
+  test("fetchData() fails if dateColumn is provided for non-timeseries data.", async () => {
+    expect(async () =>
+      testFetchingCsvData(
+        longCsv,
+        /*includeTimeseries=*/ false,
+        /*dateColumn=*/ "date",
+        /*metric=*/ undefined,
+        longCsvProviderOptions
+      )
+    ).rejects.toThrow("Missing date field");
+  });
+
   test("fetchData() fails if no data exists for the region", async () => {
     expect(async () =>
       testFetchingCsvData(
@@ -196,15 +208,15 @@ describe("CsvDataProvider with long-format CSV", () => {
     ).rejects.toThrow("No data found for region");
   });
 
-  test("fetchData() fails if dateColumn is provided for non-timeseries data.", async () => {
-    expect(async () =>
-      testFetchingCsvData(
-        longCsv,
-        /*includeTimeseries=*/ false,
-        /*dateColumn=*/ "date",
-        /*metric=*/ undefined,
-        longCsvProviderOptions
-      )
-    ).rejects.toThrow("Missing date field");
+  test("fetchData() returns null metric if the metric data for region is not found.", async () => {
+    const metricData = await testFetchingCsvData(
+      longCsv.replace(/\n.*$/, ""), // Remove last line so there's no data for the expected metric for NY.
+      /*includeTimeseries=*/ true,
+      /*dateColumn=*/ "date",
+      /*metric=*/ undefined,
+      longCsvProviderOptions
+    );
+    expect(metricData.currentValue).toBe(null);
+    expect(metricData.hasTimeseries()).toBe(false);
   });
 });
--- a/src/metrics/data_providers/CsvDataProvider.test.ts
+++ b/src/metrics/data_providers/CsvDataProvider.test.ts
@@ -9,10 +9,12 @@ const mockCsv = `region,cool_metric
 36,150
 12,`;
 
-const longCsv = `region,metric,Score
-12,cool_metric,1.2
-36,cool_metric,1.6
-25,cool_metric,1.7`;
+const longCsv = `region,metric,value,another_factor
+12,cool_metric,1.2,a
+12,another,1.3,b
+36,another,1.4,c
+36,cool_metric,1.6,d
+06,cooler_metric,1.7,d`;
 
 const csvTimeseries = `region,date,cool_metric
 36,2022-08-02,150
@@ -126,23 +128,28 @@ describe("CsvDataProvider", () => {
   });
 
   test("fetch long csv", async () => {
-    const provider = new CsvDataProvider(PROVIDER_ID, {
+    const longCsvProvider = new CsvDataProvider("long-csv", {
       regionDb: states,
       regionColumn: "region",
       format: "long",
+      longDataMetricColumn: "metric",
+      longDataValueColumn: "value",
+      // dateColumn: "date",
       csvText: longCsv,
     });
 
     // TODO: need to separate metric lookup from relying on metric id
     const metric = new Metric({
-      id: "cool_metric",
+      id: "metric",
       dataReference: {
-        providerId: PROVIDER_ID,
-        valueColumn: "Score",
-        column: "metric",
+        providerId: "long-csv",
+        column: "cool_metric",
       },
     });
-    const data = await provider.fetchDataForRegionAndMetric(newYork, metric);
+    const data = await longCsvProvider.fetchDataForRegionAndMetric(
+      newYork,
+      metric
+    );
     console.log(data);
     expect(data).toBeDefined();
   });
--- a/src/metrics/data_providers/CsvDataProvider.test.ts
+++ b/src/metrics/data_providers/CsvDataProvider.test.ts
@@ -1,7 +1,7 @@
 import { states } from "../../regions";
 import { Metric } from "../Metric";
 import { MetricCatalog } from "../MetricCatalog";
-import { CsvDataProvider } from "./CsvDataProvider";
+import { CsvDataProvider, CsvDataProviderOptions } from "./CsvDataProvider";
 
 const PROVIDER_ID = "csv-provider";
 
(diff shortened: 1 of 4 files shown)


src/metrics/data_providers/CsvDataProvider.test.ts
```diff
@@ -50,6 +68,32 @@ const testFetchingCsvData = async (
 };
 
 describe("CsvDataProvider", () => {
+  test("Constructor fails if neither url or csv data is provided.", () => {
```
smcclure17, line 76: Separating these out into their own describe block because they are not specific to `wide` or `long` formats

src/metrics/data_providers/CsvDataProvider.ts
```diff
@@ -102,22 +138,40 @@ export class CsvDataProvider extends SimpleMetricDataProviderBase {
     region: Region,
     metric: Metric
   ): Promise<MetricData<unknown>> {
+    const metricName = metric.dataReference?.column;
```
smcclure17: We should probably rename this `column` variable as it no longer strictly refer to a metric column (although it still does for `wide` CSVs). I'd propose something like `metricName` and documenting what that means. 

I'd prefer to do that in a follow-up PR to limit the scope of this PR, but am willing to address/update it here if we'd prefer

src/metrics/data_providers/CsvDataProvider.ts
```diff
@@ -34,21 +70,17 @@ export interface CsvDataProviderOptions {
...
- * ```
+ * Data Provider for importing data from wide- or long-format CSV files.
  */
 export class CsvDataProvider extends SimpleMetricDataProviderBase {
   private readonly regionDb: RegionDB;
   private readonly regionColumn: string;
   private readonly dateColumn?: string;
+  private readonly format: "wide" | "long";
   private readonly url?: string;
+  private readonly longDataMetricColumn?: string;
```
smcclure17: I'm not a huge fan of these names, but I can't think of anything particularly succinct and descriptive 😕 Open to other options

src/metrics/data_providers/CsvDataProvider.ts
```diff
@@ -26,6 +27,41 @@ export interface CsvDataProviderOptions {
...
+   * |CA     |2022-02-01 |var2  |66   |
+   * ```
+   *
+   */
+  format?: "wide" | "long";
+  /**
+   * For long-format CSVs, the name of the column containing the metric names
+   * (meaning, the values for Metric.dataReference.column). This should not be
+   * set for wide-format CSVs.
+   * */
```
mikelehen: nit: `*/`

src/metrics/data_providers/CsvDataProvider.ts
```diff
@@ -26,6 +27,41 @@ export interface CsvDataProviderOptions {
...
+  longDataMetricColumn?: string;
+  /**
+   * For long-format CSVs, the name of the column containing the observation values.
+   * This should not be set for wide-format CSVs.
+   * */
```
mikelehen: nit: `*/`

src/metrics/data_providers/CsvDataProvider.ts
mikelehen: I hate to make it longer but I'd maybe do `longFormat` instead of `longData` to match the `format: "long"` terminology.

That said, I'm actually a little tempted to get rid of the `longData` prefix and the `format` option, so you'd just have `metricColumn` and `valueColumn` as options.  Then if you use them, you must use both, and we'll implicitly treat the data as long format.  If you don't use them, then we'd assume wide format.  WDYT? Is that too implicit?

One other option would be something like:
```
{
  ...
  regionColumn: 'location',
  longFormat: {
    metricColumn: 'metric',
    valueColumn: 'value'
  }
}
```

with `longFormatOptions` being optional. Not sure I love it though.

mikelehen: One other thought to throw in here is that I think it'd be nice in the future to support multiple metric columns, like we have in can scrapers or like Pablo (I think) ran into with the Ember Energy data. So you could do `metricColumns: ['provider', 'variable', 'measure', 'unit']` And then in your metric definitions you'd do `columns: ['usa-facts', 'cases', 'cumulative', 'people']` or whatever...  We can save that for a future work item, but it's worth keeping in mind.

src/metrics/data_providers/CsvDataProvider.test.ts
```diff
@@ -110,13 +139,84 @@ describe("CsvDataProvider", () => {
...
+    longDataMetricColumn: "metric",
+    longDataValueColumn: "value",
+  };
+
+  test("fetchData() yields expected data", async () => {
+    const metricDataNoTs = await testFetchingCsvData(
+      longCsv,
+      /*includeTimeseries=*/ false,
+      /*dateColumn=*/ undefined,
+      /*metric=*/ undefined,
```
mikelehen: It took me a minute to understand why we were setting metric to undefined and how this was working.  I think it'd be clearer to just pass `testMetric` explicitly.  And then you don't need the `/*metric=*/` comment.

src/metrics/data_providers/CsvDataProvider.test.ts
```diff
@@ -110,13 +139,84 @@ describe("CsvDataProvider", () => {
...
+      longCsvProviderOptions
+    );
+    const metricDataTs = await testFetchingCsvData(
+      longCsvTimeseries,
+      /*includeTimeseries=*/ true,
+      /*dateColumn=*/ "date",
+      /*metric=*/ undefined,
```
mikelehen: Ditto here and below...

src/metrics/data_providers/CsvDataProvider.test.ts
```diff
@@ -110,13 +139,84 @@ describe("CsvDataProvider", () => {
...
+        longCsv.replaceAll("36", "25"), // replace NY locations with MA so there's no data for NY.
+        /*includeTimeseries=*/ true,
+        /*dateColumn=*/ undefined,
+        /*metric=*/ undefined,
+        longCsvProviderOptions
+      )
+    ).rejects.toThrow("No data found for region");
+  });
+
+  test("fetchData() returns null metric if the metric data for region is not found.", async () => {
```
mikelehen: null metric => null value ?

src/metrics/data_providers/CsvDataProvider.ts
```diff
@@ -26,6 +27,41 @@ export interface CsvDataProviderOptions {
...
+   * Example wide-format CSV:
+   * ```
+   * |region |date       |var1 |var2 |
+   * |TX     |2022-02-01 |12   |45   |
+   * |CA     |2022-02-01 |31   |66   |
+   * ```
+   *
+   * Example long-format CSV:
+   * ```
+   * |region |date       |metric|value|
```
mikelehen: FWIW- I'm a bit worried about us conflating "metric", "column", and "variable" terminology, possibly leading to user confusion...
* In the case of wide format, users map columns => metrics (via the metric definition)
* In the case of long format, users map rows => variables (via the csv data provider options) and then variables => metrics (via the metric definition).  And we overload "column" to mean variable in the metric definition, which isn't ideal.

I sort of feel like there might be a better way to represent all this, like maybe in the metric definition you could do:

```
providerId: 'my-long-format-csv',
selectRowsByColumnValues: {
  // select rows where "metric" column is "var1", could optionally add more column/value criteria.
  "metric": "var1"
}
valueColumn: "value"
```

or

```
providerId: 'my-wide-format-csv',
valueColumn: "var1"
```

Then they work the same way (note I renamed our existing "column" to "valueColumn"), but to do long format CSVs you could selects an arbitrary subset of rows when defining each metric. But I'm not sure how jarring of a change this would be...  So I'm fine with leaving this as-is for now, but maybe we can rework it in the future.

In the meantime:
1. Maybe change "metric" to "variable" in this comment.
2. I'm a little tempted to change "longDataMetricColumn" to "longDataVariableColumn" (or "variableColumn" if we shorten it; see below) and have them specify a `dataReference.variable` instead of `dataReference.column` in their metric definition if they're using long format (I think "column" is still clearer for wide format).   But I'm not sure.  WDYT?

src/metrics/data_providers/CsvDataProvider.ts
mikelehen: Per my other comment, I think `metricName` would be confusing since the CsvDataProvider isn't really metric-aware.  They will associate the CSV data to metrics separately in their metric definition.  So using a new term like "variable" makes more sense to me.  But per my other comment, I think I might be inclined to still use "column" for wide-format CSVs, but I'm not sure.

src/metrics/data_providers/data_provider_utils.ts
```diff
@@ -290,3 +291,72 @@ export async function getMetricDataFromDataRows(
...
+) {
+  const regionRows = dataRowsByRegionId[region.regionId];
+  assert(
+    regionRows,
+    `No data found for region ${region.regionId}.` +
+      `Please check the region ID and that the expected data exists.`
+  );
+  const metricDataRows = regionRows.filter(
+    (row) => row[metricField] === metricName
+  );
```
mikelehen: I'm a little wary about having to do this filter every time a metric is requested.  I was originally going to suggest that we could do a long-to-wide pivot when we read the CSV and then we could just use `getMetricDataFromDataRows()` as normal, which would avoid the filtering...  but that would make it harder to implement my `selectRows` API.  And I'm not sure how big of a perf impact it will have anyway.  So I'm fine leaving it like this until we have a concrete perf problem to solve.

src/metrics/data_providers/data_provider_utils.ts
```diff
@@ -290,3 +291,72 @@ export async function getMetricDataFromDataRows(
...
+/**
+ * Creates a MetricData object from a set of long-format data rows grouped by region ID,
+ * where metricName is the name of the metric and metricField is the name of the field containing
+ * the metric names.
+ *
+ * @param dataRowsByRegionId Data to transform into MetricData.
+ * @param region Region to get MetricData for.
+ * @param metric Metric to get MetricData for.
+ * @param metricName Name of the metric to get data for.
+ * @param metricField Name of field containing metric names.
```
mikelehen: Per my other comments, I think we are overloading "metric" and this should be "variableField" and "variableName" or something.

mikelehen (approved):
This looks good, though I had some ideas for other ways to structure the options / metric definition stuff to try to make things clearer and/or add better symmetry between long / wide format handling.  

But I also realize that our main goal is to import the climate data.  So if we want to get this landed with minimal changes and just open a github issue to revisit unification / cleanup stuff separately, I'm fine with that!

src/metrics/data_providers/CsvDataProvider.ts
mikelehen: Note: I just realized my proposed format wouldn't work well with the CMS since the CMS doesn't have a way to specify arbitrary key/value pairs (like `"metric": "var1"), so we'd probably need to do something like this instead:

```
providerId: 'my-long-format-csv',
selectRowsByColumnValues: [
  {
    column: "metric",
    value: "var1"
  }
],
valueColumn: "value"
```

I don't like that as much, but it would work. 🤷

src/metrics/data_providers/CsvDataProvider.ts
smcclure17: I don't think that's too implicit as long as it's documented well (which I hope I've done). I've switched to this approach, which I think has streamlined things a bit!

----

I agree about supporting multiple metrics. Per your suggestion, I've switched to pivoting the data. We could probably support multiple columns by merging columns together before this pivot, e.g.:

```
region,provider, variable, measure, unit, value
12,usa-facts, cases,cumulative,people,100
36,nyt,cases,cumulative,people,150
```

becomes 

```
region,provider_variable_measure_unit, value
12,usa-facts_cases_cumulative_people,100
36,nyt_cases_cumulative_people,150
```

which can then be pivoted and treated as a standalone variable:

```
region, usa-facts_cases_cumulative_people,nyt_cases_cumulative_people
12,100,
36,,150
```

But there's probably more to think about here. I can open a ticket for this

src/metrics/data_providers/CsvDataProvider.ts
smcclure17: Yeah, good point/idea. I've updated this to be `metricVariable` and: 

I've changed the metric definitions around so that, per your other comment, wide-format CSV metrics still expect a "column" field, but long-format CSVs expect a "variable" field--and I've noted this in the class documentation.

src/metrics/data_providers/data_provider_utils.ts
smcclure17: Yeah, I think pivoting the data is probably the better approach, so I've gone and implemented that to replace the filtering. This may have worse up-front performance, but it should only happen once (as the pivot happens before the caching) so it should make fetching metrics faster.

src/metrics/data_providers/CsvDataProvider.ts
smcclure17: Yeah, the terminology got a little muddled, thanks for flagging that and the improvements. I've made the "in the meantime" suggested updates 🙏 

I do like the `selectRows` API concept, though, and I think that would be nice to explore.

smcclure17:
Thanks @mikelehen for the comments and suggestions! 🙏 There were definitely ways to clean this up.

I've made some fairly large changes so I think it would be worth taking a second pass over if you have the time, but no stress if not! 

Main changes:
- Pivot data to wide in caching function to avoid filtering
- Remove "wide" or "long" format option in favor of a single `longFormatCsvOptions` for long-format data
- (Hopefully) clarify metric, variable, and column names/usage. 


Thanks again 🙏

mikelehen:
Sweet, will take a look! 🙏

src/metrics/data_providers/CsvDataProvider.ts
```diff
@@ -89,34 +120,48 @@ export class CsvDataProvider extends SimpleMetricDataProviderBase {
...
-    );
     return getMetricDataFromDataRows(
       dataRowsByRegionId,
       region,
       metric,
-      metricColumn,
-      this.dateColumn
+      variableName,
+      this.dateColumn,
+      /**strict=*/ false
```
mikelehen: nit: Extra `*` in `/**`

src/metrics/data_providers/data_provider_utils.ts
```diff
@@ -290,3 +292,45 @@ export async function getMetricDataFromDataRows(
...
+) {
+  const wide: { [regionId: string]: DataRow[] } = {};
+  for (const regionId of Object.keys(long)) {
+    wide[regionId] = [];
+    const regionData = long[regionId];
+    if (dateField) {
+      const dates = new Set(regionData.map((row) => row[dateField]));
+      for (const date of dates) {
+        const dateData = regionData.filter((row) => row[dateField] === date);
+        const wideRow = { region: regionId, [dateField]: date } as DataRow;
```
mikelehen: Hrm.  Explicitly adding `region` to each row seems funky.  I think:
1) We probably don't actually need it.  Since we store a regionId=>DataRow[] map, we don't need the region ids in the data rows.
2) If we're going to store it in the DataRows (so they more closely match what the wide format would look like), we should use the specified `regionColumn`/`regionKey` as the name rather than "region".  As-is if they e.g. had a variable called "region" (probably unlikely, I know) we would clobber it.

src/metrics/data_providers/data_provider_utils.ts
```diff
@@ -290,3 +292,45 @@ export async function getMetricDataFromDataRows(
...
+  dateField?: string
+) {
+  const wide: { [regionId: string]: DataRow[] } = {};
+  for (const regionId of Object.keys(long)) {
+    wide[regionId] = [];
+    const regionData = long[regionId];
+    if (dateField) {
+      const dates = new Set(regionData.map((row) => row[dateField]));
+      for (const date of dates) {
+        const dateData = regionData.filter((row) => row[dateField] === date);
```
mikelehen: This may not be a huge deal in practice, but I think this code is going to do more computation than is necessary.  Basically for each date, we call `filter()` on all the rows which will internally have to iterate over each row in the dataset to find the rows that match.  So if you had 1M rows with 1000 dates in them, we'd "visit" each row 1000 times for a total of 1 billion total row visits.

A theoretical optimal algorithm would probably convert wide to long with only 1 visit to each row, so 1 million total row visits.

I took a stab at rewriting it to avoid the extra filtering step by using a groupBy. It's probably not 100% optimal (since e.g. we do a groupBy() and then iterate over the result, so that means each row would be visited twice) but I think it gets us to an O(rows) computation instead of a O(rows*dates) computation if you are familiar with Big O notation...

```
export function pivotLongToWide(
  long: { [regionId: string]: DataRow[] },
  variableField: string,
  valueField: string,
  dateField?: string
): { [regionId: string]: DataRow[] } {
    const regionToWideRows = mapValues(long, (longRows) => {
    if (dateField) {
      const dateToLongRows = groupBy(longRows, (row) => row[dateField]);
      const wideRows = Object.entries(dateToLongRows).map(
        ([date, longRowsForDate]) => {
          const wideRow: DataRow = { [dateField]: date };
          for (const row of longRowsForDate) {
            wideRow[row[variableField] as string] = row[valueField];
          }
          return wideRow;
        }
      );
      return wideRows;
    } else {
      const wideRow: DataRow = {};
      for (const row of longRows) {
        wideRow[row[variableField] as string] = row[valueField];
      }
      return [wideRow];
    }
  });
  return regionToWideRows;
}
```

For the Climate Data (yearly data) it probably won't make a noticeable difference, but if you had monthly or daily data, I think this might make a meaningful difference.

Note that per my other comment, I did *not* copy the `regionId` into the resulting wide rows.  We can add that if we need / want it.

mikelehen (approved):
LGTM with nit and theoretical performance improvement suggestion.

src/metrics/data_providers/data_provider_utils.ts
smcclure17: Ah, yeah this isn't needed. I added it initially to match the wide data to test that the results were equal. The wide data retains this field since we use `groupBy` to index it by regions, but it isn't needed

src/metrics/data_providers/data_provider_utils.ts
smcclure17: ooh, nice! Yeah that makes a lot of sense, thanks for reworking this 🙏

PR #584, opened 2023-02-07
Allow raw years (2020) as date strings in CSV / JSON files.

### LINKED ISSUE #560: Allow raw years as dates for metric data providers?
Right now you must specify a full date (e.g. "2022-01-01") in your CSV files.  If you provide e.g. just a year, you'll get:

> INTERNAL ASSERTION FAILED: Date column must be a string. number found.

It would be nice to allow raw dates ("2022") or maybe even months ("2022-01").

mikelehen:
Fixes #560

### DIFF
--- a/.changeset/forty-moons-guess.md
+++ b/.changeset/forty-moons-guess.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/actnow.js": patch
+---
+
+Allow raw years ("2020") as date strings in CSV / JSON files.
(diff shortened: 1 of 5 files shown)


pnavarrc (approved):
Looks good! I was a bit confused by the title (I thought we allowed `"2020"`, but I see that we now allow the number `2020` as well)

mikelehen:
Yeah, the problem is actually a bit subtle.  It's related to the fact that CSVs don't differentiate numbers from strings so we have to infer the type and if you had a raw year date (`2022`) in a CSV we would parse it to a number and then reject it.  I realized though that we can "fix" that and force the date column to be parsed as a string.  That's probably a bit cleaner.  So I went ahead and tweaked the logic to do that.

PR #585, opened 2023-02-07
Make ShareButton styled-able

### LINKED ISSUE #583: ShareButton is not theme-able enough
When we first implemented the `ShareButton`, we wanted to limit how customizable it is, but we found a case where we would like to customize the color of a specific instance, so maybe it would be worth allowing to customize the button color in the same way as the MUI `Button` and `Typography` component do

https://demo-world-happiness-report-git-pablo-341709-act-now-coalition.vercel.app/country/canada

```tsx
// will be white
<Typography variant="h1" color="white">Title</Typography>

// not a valid prop
<ShareButton {...props} color="white" />
```

an alternative would be to wrap the "above the fold" section with a theme that modifies the buttons using the theme system, which is cleaner and more flexible, but also more work for the user trying to customize a particular section.

```tsx
<ThemeProvider theme={aboveTheFoldTheme}>
  <Typography variant="h1">Title</Typography>
  <ShareButton {...props} />
</ThemeProvider>
```

Another example of where theme wrapping might be a good idea is the footer, since we would need to change buttons, links and typography colors (if using a dark background).

Image

pnavarrc:
Close #583 - ShareButton is not theme-able enough

The problem is that as it is, the button doesn't allow for one-off customizations. A simple solution is to add `className` to the button props, which allows us to use `styled` to create custom versions of it.

image

### DIFF
--- a/src/ui-components/components/ShareButton/ShareButton.tsx
+++ b/src/ui-components/components/ShareButton/ShareButton.tsx
@@ -57,7 +57,7 @@ export interface ShareButtonProps {
   /**
    * MUI Button className applied to the anchor button.
    */
-  className?: string;
+  className?: ButtonProps["className"];
 }
 
 export const ShareButton = ({
--- a/src/ui-components/components/ShareButton/ShareButton.tsx
+++ b/src/ui-components/components/ShareButton/ShareButton.tsx
@@ -54,6 +54,10 @@ export interface ShareButtonProps {
    * @default "large"
    */
   size?: ButtonProps["size"];
+  /**
+   * MUI Button className applied to the anchor button.
+   */
+  className?: string;
 }
 
 export const ShareButton = ({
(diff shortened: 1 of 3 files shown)


mikelehen (approved):
LGTM though I wonder if there are other components that should do this too?  Like... maybe all of our components?  Not really sure. 😁

pnavarrc:
@mikelehen probably, although most components can be more easily manipulated using the theme or are styled-able already, but that's a good point, I'm going to make a note to confirm.

src/ui-components/components/ShareButton/ShareButton.tsx
```diff
@@ -54,6 +54,10 @@ export interface ShareButtonProps {
    * @default "large"
    */
   size?: ButtonProps["size"];
+  /**
+   * MUI Button className applied to the anchor button.
+   */
+  className?: string;
```
chasulin: how about `className?: ButtonProps["className"];` , which has the built in type of `string | undefined`

chasulin: and in documentation maybe lets add `@default ""` for completion

chasulin:
ooh this def works. one reservation though is that it nixes the ability to pass an anchor element other than a button that says `Share` with the icon. i think its worth giving flexibility in terms of what the anchor is

PR #588, opened 2023-02-07
Allow ShareButton to receive a custom anchor button

pnavarrc:
Follow up to #585 - This PR allows further customization of the ShareButton by allowing the user to pass a `Button` instance. I considered allowing to pass any clickable element, but that restricts the type of the `onClick` handlers and makes things a bit more complicated

image

### DIFF
--- a/src/ui-components/components/ShareButton/ShareButton.tsx
+++ b/src/ui-components/components/ShareButton/ShareButton.tsx
@@ -56,12 +56,13 @@ export interface ShareButtonProps {
   size?: ButtonProps["size"];
   /**
    * MUI Button className applied to the anchor button.
+   * @default ""
    */
   className?: ButtonProps["className"];
   /**
-   *
+   * MUI Button instance to use as anchor button.
    */
-  anchor?: React.ReactElement<ButtonProps, typeof Button>;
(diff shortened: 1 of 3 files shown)


src/ui-components/components/ShareButton/ShareButton.tsx
```diff
@@ -71,30 +76,48 @@ export const ShareButton = ({
...
+  // can call the original onClick property and the handleClick function.
+  const modifiedAnchorButton = anchorButton ? (
+    React.cloneElement(anchorButton, {
+      ...anchorButton.props,
+      onClick: (event: React.MouseEvent<HTMLButtonElement>) => {
+        handleClick(event);
+        anchorButton?.props?.onClick && anchorButton.props.onClick(event);
+      },
+    })
+  ) : (
```
faihegberg, line 102: It feels a little strange that we need to clone this to get props of a very similar element, but currently I can't think of a better alternative either.

faihegberg (approved):

src/ui-components/components/ShareButton/ShareButton.tsx
pnavarrc: We need to clone the element to be able to overwrite the `onClick` function from the element passed in, I don't think there is another way to do this

PR #592, opened 2023-02-08
Use validate package in user-facing functions

### LINKED ISSUE #561: Audit all assertions and create better user-facing error messages.
We should reserve assert() for internal assertions that a consumer of ActNowJS shouldn't be able to trigger, even if they misconfigure things or pass a bad parameter to a component. They should not be able to trigger an "INTERNAL ASSERTION FAILED" error message by misusing our packages.

We probably need to:
* Audit all usages of assert() and fail().
* Introduce some new validation helpers for providing user-facing error messages (that doesn't say internal assertion failed)
* Audit our API surface area and make sure we have reasonable validation in place.

faihegberg:
Improve error messages throughout our codebase by using the newly implemented "validate" package.

Closes https://github.com/act-now-coalition/act-now-packages/issues/561

### DIFF
--- a/.changeset/sixty-dogs-attend.md
+++ b/.changeset/sixty-dogs-attend.md
@@ -1,0 +1,5 @@
+---
+"@actnowcoalition/actnow.js": patch
+---
+
+Improve error messages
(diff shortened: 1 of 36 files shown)


mikelehen (approved):
Looks good! 🙏

